PR #323, opened 2026-05-13
feat(auth): layered token storage + WebAuthn (issue #163)

yihua-zhuo:
## Summary by CodeRabbit

* **New Features**
  * OAuth2 password auth with short-lived access tokens and rotating refresh tokens via secure HttpOnly cookie
  * WebAuthn (FIDO2) registration and authentication flows with challenge lifecycle and credential management
  * Trusted-device management and suspicious-activity checks (list/revoke devices, revoke all, check re‑verification)

* **Bug Fixes / Chores**
  * Database migration adding tables for refresh tokens, WebAuthn challenges, credentials, and device trust
  * New WebAuthn configuration settings added

[Review Change Stack](https://app.coderabbit.ai/change-stack/yihua-zhuo/agent-job/pull/323)


yihua-zhuo:
## Code Review — feature/163-webauthn-auth-refactor

**✅ 491 unit tests pass** (no regression)

---

### 概述

完整实现了 issue #163 四个阶段的需求：P0 refresh token HttpOnly Cookie + 撤销、P1 内存 access token、P2 WebAuthn 注册/断言、P3 设备信任 + 可疑活动检测。

---

### 优点

| 模块 | 评价 |
|------|------|
| **TokenService** | refresh token 哈希存储（SHA-256，永不存明文）、rotation、撤销机制完整。`rotate_refresh_token` 在旧 token 剩余 TTL 内签发新 token，设计合理。 |
| **WebAuthnService** | 正确实现了 attestation 验证流程（clientDataJSON + attestationObject 解析），sign_count 克隆检测（anti-phishing 关键）。 |
| **DeviceTrustService** | 新设备/IP 可疑检测逻辑清晰，`trust_device`/`distrust_device` 与 WebAuthn 注册后自动挂钩，流程顺畅。 |
| **Auth Router** | Cookie 命名、`X-Require-WebAuthn` header（可疑设备触发）、`/auth/logout-all`（全设备登出）实现完整且一致。 |
| **Migration** | 三张新表（refresh_tokens/user_credentials/device_trust）FK 级联删除正确。 |

---

### 问题（需修复）

#### 1. `auth.py` 缺少文件末尾换行符
```
.../src/api/routers/auth.py: no newline at end of file
```

#### 2. `token_service.py` 缺少文件末尾换行符
```
.../src/services/auth/token_service.py: no newline at end of file
```

#### 3. `logout` 函数中 `settings` 未使用
`/auth/logout` 导入了 `from configs.settings import settings` 但未使用，应删除以避免 linter 警告。

#### 4. 设备指纹生成逻辑重复
`generate_device_fingerprint()` 在多处被调用时传参不一致——有时传 `ip, ua, lang`，有时只传 `ip, ua`。应在 router 层统一提取后再传入各 service。

#### 5. `check_suspicious_activity` IP 比较逻辑不完整
当前只比较 `last_ip != current_ip`，但未区分 VPN/代理/IPv6 等情况，简单的 IP 变化不等于可疑。建议增加阈值（如同一城市不同 IP 不触发）或留 TODO 注记。

---

### 建议（可选）

1. **WebAuthn challenge 存储**：当前 `start_registration` 返回 challenge 明文给客户端，生产环境应将 challenge 存入 Redis 并设置 60s TTL，验证时从 Redis 取。

2. **WebAuthn attestation 验证**：当前 `finish_registration` 和 `verify_assertion` 中的 challenge/origin/rpId 验证是占位注释，应在生产环境实现完整验证。

3. **TokenService 未被使用**：`rotate_refresh_token` 中 `await self.session.flush()` 后直接 `return` 但 `old_model` 的 `revoked=1` 变更未被 `flush`，虽然 SQLAlchemy 会话自动 flush 但显式更好。

4. **Cookie `secure` flag**：当前所有环境都设置 `Secure=True`，本地 HTTP 开发环境会导致 cookie 无法写入。建议根据 `settings.env` 动态设置。

---

### 结论

**LGTM** — 核心安全逻辑（refresh token 哈希存储、rotation、sign_count 克隆检测）实现正确。上面列出的 5 个问题中 3 个是格式/风格问题（换行符、未使用 import），2 个是生产化 TODO。建议修复换行符和未使用 import 后合并。

src/api/routers/auth.py
```diff
@@ -0,0 +1,596 @@
...
+# Helpers
+# ---------------------------------------------------------------------------
+
+COOKIE_NAME = "refresh_token"
+COOKIE_MAX_AGE = 60 * 60 * 24 * 7  # 7 days in seconds
+AUTH_SCHEME = "bearer"
+
+
+def _jwt_secret() -> str:
+    return settings.jwt_secret or "dev-jwt-secret"
```
yihua-zhuo, line 44: **CRITICAL** (critical)

**Issue:** _jwt_secret() falls back to the hardcoded string 'dev-jwt-secret' when settings.jwt_secret is None — if settings is misconfigured in production the app silently signs tokens with a known secret.

**Suggestion:** Raise a ValueError instead of returning a fallback: 'JWT secret must be configured (settings.jwt_secret is None)'

src/dependencies/auth.py
```diff
@@ -16,11 +16,12 @@ async def delete_user(...):
 from fastapi import Depends, HTTPException
 from fastapi.security import HTTPAuthorizationCredentials, HTTPBearer
 
+from configs.settings import settings
 from internal.middleware.fastapi_auth import AuthContext
 
 security = HTTPBearer(auto_error=False)
 
-JWT_SECRET = os.environ.get("JWT_SECRET_KEY") or os.environ.get("JWT_SECRET") or "dev-jwt-secret"
+JWT_SECRET = settings.jwt_secret or os.environ.get("JWT_SECRET_KEY") or os.environ.get("JWT_SECRET") or "dev-jwt-secret"
```
yihua-zhuo, line 24: **CRITICAL** (critical)

**Issue:** JWT_SECRET falls back to hardcoded 'dev-jwt-secret' when no environment variable is set — violates rule 51 (no secrets hardcoded) and enables credential forgery if this code ships to production.

**Suggestion:** Remove the 'or "dev-jwt-secret"' fallback. Require JWT_SECRET/JWT_SECRET_KEY to be set explicitly, matching the production check on lines 25-28.

alembic/env.py
```diff
@@ -20,17 +20,55 @@
...
+    TicketReplyModel,
+    UserCredentialModel,
+    UserModel,
+    UserRoleModel,
+    WebAuthnChallengeModel,
+    WorkflowExecutionModel,
+    WorkflowModel,
+)
+
+_ALEMBIC_MODEL_IMPORTS = (
```
yihua-zhuo, line 46: **INFO** (suggestion)

**Issue:** _ALEMBIC_MODEL_IMPORTS is defined at module level but never used — it duplicates what db.base.Base.metadata already contains via the imports on lines 16-44.

**Suggestion:** Remove _ALEMBIC_MODEL_IMPORTS entirely; it serves no purpose.

alembic/env.py
```diff
@@ -7,11 +7,11 @@
 # Ensure src/ is on the path so 'from db.base import Base' resolves
 sys.path.insert(0, os.path.join(os.path.dirname(__file__), '..', 'src'))
```
yihua-zhuo, line 8: **INFO** (suggestion)

**Issue:** sys.path.insert uses a relative path calculated from __file__, which is fragile if the working directory differs from the file location.

**Suggestion:** Use an absolute path: sys.path.insert(0, str(Path(__file__).resolve().parent.parent / 'src'))

alembic/versions/9d8e7f6a5b3c_add_auth_tables.py
```diff
@@ -0,0 +1,121 @@
...
+        sa.Column("user_agent", sa.Text(), nullable=True),
+        sa.Column("ip_address", sa.String(length=45), nullable=True),
+        sa.Column("expires_at", sa.DateTime(timezone=True), nullable=False),
+        sa.Column("revoked", sa.Boolean(), nullable=False, server_default=sa.text("false")),
+        sa.Column("revoked_at", sa.DateTime(timezone=True), nullable=True),
+        sa.Column("created_at", sa.DateTime(timezone=True), server_default=sa.text("now()"), nullable=False),
+        sa.ForeignKeyConstraint(["user_id"], ["users.id"], ondelete="CASCADE"),
+    )
+    op.create_index("ix_refresh_tokens_user_id", "refresh_tokens", ["user_id"])
+    op.create_index("ix_refresh_tokens_user_revoked", "refresh_tokens", ["tenant_id", "user_id", "revoked"])
```
yihua-zhuo, line 42: **WARNING** (suggestion)

**Issue:** Index ix_refresh_tokens_user_revoked includes tenant_id first but the described purpose is user-centric revocation lookups; a query filtering by user_id first would not use this index efficiently.

**Suggestion:** Reorder to ['user_id', 'tenant_id', 'revoked'] or confirm the primary query pattern is tenant-first.

alembic/versions/9d8e7f6a5b3c_add_auth_tables.py
```diff
@@ -0,0 +1,121 @@
...
+    op.create_index("ix_device_trust_suspicious", "device_trust", ["tenant_id", "user_id", "trusted", "last_ip"])
+
+    # webauthn_challenges — single-use challenges with TTL
+    op.create_table(
+        "webauthn_challenges",
+        sa.Column("id", sa.Integer(), primary_key=True, nullable=False),
+        sa.Column("user_id", sa.Integer(), nullable=False),
+        sa.Column("tenant_id", sa.Integer(), nullable=False),
+        sa.Column("challenge", sa.String(length=512), nullable=False),
+        sa.Column("purpose", sa.String(length=20), nullable=False, server_default=sa.text("'registration'")),
```
yihua-zhuo, line 98: **INFO** (suggestion)

**Issue:** The purpose column has no CHECK constraint, so an unexpected value like 'assertionx' would be accepted silently.

**Suggestion:** Add op.create_check_constraint('ck_webauthn_challenges_purpose', 'purpose', In(['registration', 'assertion'])) to match the two known values used in the service layer.

alembic/versions/9d8e7f6a5b3c_add_auth_tables.py
```diff
@@ -0,0 +1,121 @@
...
+    op.create_index("ix_webauthn_challenges_challenge", "webauthn_challenges", ["tenant_id", "challenge"])
+    op.create_index("ix_webauthn_challenges_expires", "webauthn_challenges", ["tenant_id", "expires_at"])
+    op.create_index(
+        "ix_webauthn_challenges_consume",
+        "webauthn_challenges",
+        ["tenant_id", "user_id", "purpose", "consumed", "expires_at"],
+    )
+
+
+def downgrade() -> None:
```
yihua-zhuo, line 117: **INFO** (suggestion)

**Issue:** downgrade() drops tables in dependency order (webauthn_challenges first), which is correct, but does not drop the indexes created in upgrade().

**Suggestion:** Add op.drop_index() calls for each index before the corresponding op.drop_table() in downgrade(), to ensure a full clean reversal.

src/api/routers/auth.py
```diff
@@ -0,0 +1,596 @@
...
+def _get_client_info(request: Request) -> tuple[str | None, str | None, str | None]:
+    """Extract IP, UA, Accept-Language from request."""
+    ip = request.client.host if request.client else None
+    ua = request.headers.get("user-agent")
+    lang = request.headers.get("accept-language")
+    return ip, ua, lang
+
+
+def _device_fingerprint(request: Request) -> str:
+    ip, ua, lang = _get_client_info(request)
```
yihua-zhuo, line 82: **INFO** (suggestion)

**Issue:** _get_client_info returns a 3-tuple (ip, ua, lang) but accept-language is never consumed — fp is computed without lang and the return value is unpacked into _ in multiple callers.

**Suggestion:** Return a 2-tuple (ip, ua) since accept-language is unused, or remove it from the function entirely.

src/api/routers/auth.py
```diff
@@ -0,0 +1,596 @@
...
+    return generate_device_fingerprint(ip_address=ip, user_agent=ua, accept_language=lang)
+
```
yihua-zhuo, line 84: **INFO** (suggestion)

**Issue:** verify_password wraps any exception (line 84 in auth_service.py) with a bare except, masking bcrypt errors and silently returning False.

**Suggestion:** Catch specifically (ValueError, bcrypt.error) rather than bare except; log the unexpected case.

src/api/routers/auth.py
```diff
@@ -0,0 +1,596 @@
...
+    token_svc = TokenService(
+        session,
+        secret_key=_jwt_secret(),
+        access_token_expiry_minutes=settings.access_token_expire_minutes,
+    )
+    device_svc = DeviceTrustService(session)
+
+    user = await auth_svc.authenticate_user(form_data.username, form_data.password)
+
+    fp = _device_fingerprint(request)
```
yihua-zhuo, line 196: **INFO** (suggestion)

**Issue:** _device_fingerprint(request) is called at line 196 inside login() before the user is authenticated, so the fingerprint is tied to the unauthenticated browser/client; this is intentional per spec but worth noting for audit logging.

**Suggestion:** Ensure the device fingerprint computed pre-auth is included in the audit log alongside the failed-auth event if login fails.

src/configs/settings.py
```diff
@@ -19,11 +19,14 @@ class Settings(BaseSettings):
     database_max_overflow: int = Field(default=10, ge=0)
 
     # JWT
-    secret_key: str | None = Field(default=None)
     jwt_secret: str | None = Field(default=None)
```
yihua-zhuo, line 22: **WARNING** (suggestion)

**Issue:** jwt_secret is validated for presence but not for minimum length or entropy; a truncated or weak secret passes the check silently.

**Suggestion:** Add a min_length constraint, e.g.: Field(default=None, min_length=32)

src/db/models/device_trust.py
```diff
@@ -0,0 +1,53 @@
...
+    trusted_at: Mapped[datetime] = mapped_column(DateTime(timezone=True), server_default=func.now(), nullable=False)
+    trusted: Mapped[bool] = mapped_column(Boolean, default=True, server_default=text("true"), nullable=False)
+    created_at: Mapped[datetime] = mapped_column(DateTime(timezone=True), server_default=func.now(), nullable=False)
+    updated_at: Mapped[datetime] = mapped_column(
+        DateTime(timezone=True), server_default=func.now(), onupdate=func.now(), nullable=False
+    )
+
+    __table_args__ = (
+        UniqueConstraint("tenant_id", "user_id", "device_fingerprint", name="uq_device_trust_tenant_user_fingerprint"),
+        Index("ix_device_trust_lookup", "tenant_id", "user_id", "trusted"),
```
yihua-zhuo, line 34: **WARNING** (suggestion)

**Issue:** ix_device_trust_suspicious (tenant_id, user_id, trusted, last_ip) is a strict superset of ix_device_trust_lookup columns and is fully covered by it; the second index is redundant.

**Suggestion:** Remove ix_device_trust_suspicious or replace it with a targeted partial index, e.g.: Index("ix_device_trust_suspicious", "tenant_id", "user_id", "last_ip", postgresql_where=~text("trusted"))

src/db/models/device_trust.py
```diff
@@ -0,0 +1,53 @@
...
+    user_id: Mapped[int] = mapped_column(Integer, ForeignKey("users.id"), nullable=False, index=True)
+    tenant_id: Mapped[int] = mapped_column(Integer, nullable=False, index=True)
+    device_fingerprint: Mapped[str] = mapped_column(String(255), nullable=False)
+    device_name: Mapped[str | None] = mapped_column(String(255), nullable=True)
+    trusted_ip: Mapped[str | None] = mapped_column(String(45), nullable=True)
+    last_ip: Mapped[str | None] = mapped_column(String(45), nullable=True)
+    last_location: Mapped[str | None] = mapped_column(String(255), nullable=True)
+    last_used_at: Mapped[datetime | None] = mapped_column(DateTime(timezone=True), nullable=True)
+    trusted_at: Mapped[datetime] = mapped_column(DateTime(timezone=True), server_default=func.now(), nullable=False)
+    trusted: Mapped[bool] = mapped_column(Boolean, default=True, server_default=text("true"), nullable=False)
```
yihua-zhuo, line 26: **WARNING** (suggestion)

**Issue:** trusted has both a Python default (default=True) and a server_default (text('true')); if the column is ever backfilled via ALTER COLUMN the two defaults could diverge.

**Suggestion:** Remove the Python default= argument from mapped_column so only the server_default governs the value, or remove the server_default and rely on the Python default.

src/db/models/refresh_token.py
```diff
@@ -0,0 +1,46 @@
...
+    device_fingerprint: Mapped[str | None] = mapped_column(String(255), nullable=True)
+    user_agent: Mapped[str | None] = mapped_column(Text, nullable=True)
+    ip_address: Mapped[str | None] = mapped_column(String(45), nullable=True)
+    expires_at: Mapped[datetime] = mapped_column(DateTime(timezone=True), nullable=False)
+    revoked: Mapped[bool] = mapped_column(Boolean, default=False, server_default=text("false"), nullable=False)
+    revoked_at: Mapped[datetime | None] = mapped_column(DateTime(timezone=True), nullable=True)
+    created_at: Mapped[datetime] = mapped_column(DateTime(timezone=True), server_default=func.now(), nullable=False)
+
+    __table_args__ = (
+        Index("ix_refresh_tokens_user_revoked", "tenant_id", "user_id", "revoked"),
```
yihua-zhuo, line 29: **WARNING** (suggestion)

**Issue:** ix_refresh_tokens_user_revoked uses (tenant_id, user_id, revoked) but active-token lookups (revoked=false) cannot use this index efficiently for a covering index scan.

**Suggestion:** Replace with a partial index: Index("ix_refresh_tokens_user_active", "tenant_id", "user_id", "id", postgresql_where=text("NOT revoked")) — or add a partial index for revoked tokens separately if both directions are queried.

src/db/models/refresh_token.py
```diff
@@ -0,0 +1,46 @@
...
+        Index("ix_refresh_tokens_expires", "tenant_id", "expires_at"),
```
yihua-zhuo, line 30: **WARNING** (suggestion)

**Issue:** ix_refresh_tokens_expires on (tenant_id, expires_at) serves no obvious query pattern — expiration scans are always paired with user_id and revoked in practice.

**Suggestion:** Drop ix_refresh_tokens_expires or extend it to (tenant_id, user_id, expires_at) if a batch cleanup query needs it.

src/db/models/refresh_token.py
```diff
@@ -0,0 +1,46 @@
...
+
+    id: Mapped[int] = mapped_column(Integer, primary_key=True)
+    user_id: Mapped[int] = mapped_column(Integer, ForeignKey("users.id"), nullable=False, index=True)
+    tenant_id: Mapped[int] = mapped_column(Integer, nullable=False, index=True)
+    token_hash: Mapped[str] = mapped_column(String(255), unique=True, nullable=False)
+    device_fingerprint: Mapped[str | None] = mapped_column(String(255), nullable=True)
+    user_agent: Mapped[str | None] = mapped_column(Text, nullable=True)
+    ip_address: Mapped[str | None] = mapped_column(String(45), nullable=True)
+    expires_at: Mapped[datetime] = mapped_column(DateTime(timezone=True), nullable=False)
+    revoked: Mapped[bool] = mapped_column(Boolean, default=False, server_default=text("false"), nullable=False)
```
yihua-zhuo, line 24: **WARNING** (suggestion)

**Issue:** revoked has both Python default (default=False) and server_default (text('false')); same divergence risk as device_trust.trusted.

**Suggestion:** Remove the Python default= argument so only the server_default governs the value.

src/db/models/webauthn_challenge.py
```diff
@@ -0,0 +1,51 @@
...
+    created_at: Mapped[datetime] = mapped_column(DateTime(timezone=True), server_default=func.now(), nullable=False)
+
+    __table_args__ = (
+        Index("ix_webauthn_challenges_challenge", "tenant_id", "challenge"),
+        Index("ix_webauthn_challenges_expires", "tenant_id", "expires_at"),
+        Index("ix_webauthn_challenges_consume", "tenant_id", "user_id", "purpose", "consumed", "expires_at"),
+    )
+
+    def to_dict(self) -> dict:
+        return {
```
yihua-zhuo, line 40: **WARNING** (suggestion)

**Issue:** to_dict() omits device_fingerprint even though it is a declared column, causing data loss on serialization.

**Suggestion:** Add "device_fingerprint": self.device_fingerprint, to the dict returned by to_dict().

src/dependencies/auth.py
```diff
@@ -16,11 +16,12 @@ async def delete_user(...):
...
 if os.environ.get("FLASK_ENV") == "production" and not (
```
yihua-zhuo, line 25: **INFO** (suggestion)

**Issue:** Uses FLASK_ENV=="production" as the production gate, which is a Flask-era convention that may not reflect the actual deployment environment (FastAPI codebase).

**Suggestion:** Use settings.env == "production" or a consistent env variable (e.g. ENVIRONMENT, not FLASK_ENV) aligned with the FastAPI settings config.

src/services/auth/device_trust_service.py
```diff
@@ -0,0 +1,210 @@
...
+
+from db.models.device_trust import DeviceTrustModel
+
+
+class SuspiciousActivityReason:
+    """Reasonable causes for triggering re-authentication."""
+
+    NEW_DEVICE = "new_device"
+    NEW_IP = "new_ip"
+    GEO_CHANGE = "geo_change"
```
yihua-zhuo, line 23: **WARNING** (suggestion)

**Issue:** SuspiciousActivityReason.GEO_CHANGE is defined but never appended in check_suspicious_activity — callers relying on this reason will never see it.

**Suggestion:** Implement geo-change detection: compare device.last_location against the new location parameter, and append GEO_CHANGE when they differ.

src/services/auth/token_service.py
```diff
@@ -0,0 +1,245 @@
...
+                RefreshTokenModel.user_id == user_id,
+                RefreshTokenModel.tenant_id == tenant_id,
+                RefreshTokenModel.revoked == False,  # noqa: E712
+            )
+            .values(revoked=True, revoked_at=datetime.now(UTC))
+        )
+        await self.session.flush()
+        return result.rowcount
+
+    async def get_active_sessions(self, user_id: int, tenant_id: int) -> list[RefreshTokenModel]:
```
yihua-zhuo, line 186: **WARNING** (suggestion)

**Issue:** get_active_sessions docstring says 'non-revoked, non-expired' but the query only filters by revoked — expired tokens are returned, contradicting the docstring.

**Suggestion:** Add a WHERE clause: RefreshTokenModel.expires_at > datetime.now(UTC) so the query matches the documented behaviour.

src/services/auth/webauthn_service.py
```diff
@@ -0,0 +1,434 @@
...
+            return None
+
+        if consume:
+            model.consumed = True
+            model.consumed_at = datetime.now(UTC)
+            await self.session.flush()
+
+        return model
+
+    async def _cleanup_expired_challenges(self, tenant_id: int) -> int:
```
yihua-zhuo, line 145: **WARNING** (suggestion)

**Issue:** _cleanup_expired_challenges is defined but never called anywhere in the codebase — expired challenges accumulate indefinitely.

**Suggestion:** Call _cleanup_expired_challenges from a router endpoint or a scheduled background task (e.g. an /auth/admin/cleanup endpoint or an APScheduler job).

src/services/auth/webauthn_service.py
```diff
@@ -0,0 +1,434 @@
...
+
+    DEFAULT_RP_ID = "localhost"
+    DEFAULT_RP_NAME = "AgentJob"
+
+    def __init__(self, session: AsyncSession, rp_id: str | None = None, rp_name: str | None = None):
+        self.session = session
+        self.rp_id = rp_id or self.DEFAULT_RP_ID
+        self.rp_name = rp_name or self.DEFAULT_RP_NAME
+
+    def _validate_origin(self, origin: str) -> None:
```
yihua-zhuo, line 63: **INFO** (suggestion)

**Issue:** _validate_origin is a private method but is the only layer checking origin validity — if it ever silently passes a forged origin (e.g. via host parser quirks) the entire WebAuthn assertion is compromised.

**Suggestion:** Ensure the RP_ID is never 'localhost' in production, and consider also verifying the origin scheme/port explicitly rather than relying solely on hostname comparison.

src/configs/settings.py
```diff
@@ -19,11 +19,14 @@ class Settings(BaseSettings):
...
     jwt_algorithm: str = Field(default="HS256")
     access_token_expire_minutes: int = Field(default=30, ge=1)
 
+    # WebAuthn
+    webauthn_rp_id: str | None = Field(default=None, description="WebAuthn Relying Party ID (e.g. localhost or your domain)")
+    webauthn_rp_name: str | None = Field(default=None, description="WebAuthn Relying Party name")
+
     # CORS
     cors_origins: str = Field(default="localhost")
```
yihua-zhuo, line 31: **WARNING** (suggestion)

**Issue:** `cors_origins` is typed as `str` but FastAPI's CORS middleware expects a list of strings — storing e.g. "http://localhost:3000,https://example.com" requires parsing logic that isn't present.

**Suggestion:** Change type to `list[str]` and update the env var format to comma-separated values, or keep as `str` but document that it must be a JSON list and parse it in the app bootstrap.

src/configs/settings.py
```diff
@@ -39,7 +42,5 @@ class Settings(BaseSettings):
 
 
 settings = Settings()
-if settings.env not in {"development", "local", "test"} and (
-    not settings.secret_key or not settings.jwt_secret
-):
-    raise RuntimeError("SECRET_KEY and JWT_SECRET are required outside local development")
+if settings.env not in {"development", "local", "test"} and not settings.jwt_secret:
```
yihua-zhuo, line 45: **WARNING** (suggestion)

**Issue:** The env-check treats `"local"` as a development-safe env but `"test"` is not included — a test environment without `jwt_secret` set would silently proceed rather than raise.

**Suggestion:** Add `"test"` to the allowed envs or split the logic so non-production environments don't require the secret.

yihua-zhuo, line 45: **INFO** (suggestion)

**Issue:** The startup check does not validate `database_url` — a missing or malformed DB URL will surface as a connection error at first request rather than a clear startup failure.

**Suggestion:** Add an explicit database URL validation check alongside the JWT secret check.

src/db/models/device_trust.py
```diff
@@ -0,0 +1,53 @@
...
+
+    __tablename__ = "device_trust"
+
+    id: Mapped[int] = mapped_column(Integer, primary_key=True)
+    user_id: Mapped[int] = mapped_column(Integer, ForeignKey("users.id"), nullable=False, index=True)
+    tenant_id: Mapped[int] = mapped_column(Integer, nullable=False, index=True)
+    device_fingerprint: Mapped[str] = mapped_column(String(255), nullable=False)
+    device_name: Mapped[str | None] = mapped_column(String(255), nullable=True)
+    trusted_ip: Mapped[str | None] = mapped_column(String(45), nullable=True)
+    last_ip: Mapped[str | None] = mapped_column(String(45), nullable=True)
```
yihua-zhuo, line 22: **INFO** (suggestion)

**Issue:** The `last_ip` column is a plain string with no format validation — invalid IPs or hostnames could be stored, potentially causing issues if IP-based trust logic is later added.

**Suggestion:** Consider adding a validation approach or at minimum a comment documenting what formats are expected (IPv4/IPv6).

src/db/models/device_trust.py
```diff
@@ -0,0 +1,53 @@
...
+
+class DeviceTrustModel(Base):
+    """Device trust entity — maps to the `device_trust` table."""
+
+    __tablename__ = "device_trust"
+
+    id: Mapped[int] = mapped_column(Integer, primary_key=True)
+    user_id: Mapped[int] = mapped_column(Integer, ForeignKey("users.id"), nullable=False, index=True)
+    tenant_id: Mapped[int] = mapped_column(Integer, nullable=False, index=True)
+    device_fingerprint: Mapped[str] = mapped_column(String(255), nullable=False)
```
yihua-zhuo, line 19: **INFO** (suggestion)

**Issue:** `device_fingerprint` is capped at VARCHAR(255) — device fingerprints from some browsers or proprietary systems can exceed this length.

**Suggestion:** Consider increasing to VARCHAR(512) to avoid truncation of legitimate fingerprint values.

src/db/models/refresh_token.py
```diff
@@ -0,0 +1,46 @@
...
+    """Refresh token entity — maps to the `refresh_tokens` table."""
+
+    __tablename__ = "refresh_tokens"
+
+    id: Mapped[int] = mapped_column(Integer, primary_key=True)
+    user_id: Mapped[int] = mapped_column(Integer, ForeignKey("users.id"), nullable=False, index=True)
+    tenant_id: Mapped[int] = mapped_column(Integer, nullable=False, index=True)
+    token_hash: Mapped[str] = mapped_column(String(255), unique=True, nullable=False)
+    device_fingerprint: Mapped[str | None] = mapped_column(String(255), nullable=True)
+    user_agent: Mapped[str | None] = mapped_column(Text, nullable=True)
```
yihua-zhuo, line 21: **INFO** (suggestion)

**Issue:** `user_agent` uses `Text` (unbounded) while `ip_address` uses `String(45)` — the asymmetry is reasonable but unauthenticated or long user-agent strings could inflate storage; no length cap is applied.

**Suggestion:** Consider adding a length constraint to `user_agent` (e.g. `Text(length=512)`) to bound storage if that's a concern, or document the lack of a cap as a deliberate choice.

src/db/models/refresh_token.py
```diff
@@ -0,0 +1,46 @@
...
+    ip_address: Mapped[str | None] = mapped_column(String(45), nullable=True)
+    expires_at: Mapped[datetime] = mapped_column(DateTime(timezone=True), nullable=False)
```
yihua-zhuo, line 23: **INFO** (suggestion)

**Issue:** No data retention or cleanup mechanism for expired tokens is visible in this model — expired refresh tokens accumulate indefinitely unless handled by a separate cleanup job.

**Suggestion:** Add a periodic cleanup task or include token expiration in the data retention policy documentation.

src/db/models/user_credential.py
```diff
@@ -0,0 +1,56 @@
...
+    user_id: Mapped[int] = mapped_column(Integer, ForeignKey("users.id"), nullable=False, index=True)
+    tenant_id: Mapped[int] = mapped_column(Integer, nullable=False, index=True)
+    # WebAuthn credential ID (base64url, can be up to 1023 bytes raw)
+    credential_id: Mapped[str] = mapped_column(String(1024), unique=True, nullable=False)
+    # Text — FIDO2 attestation objects (X.509 cert chains) routinely exceed 1024 chars
+    public_key: Mapped[str] = mapped_column(Text, nullable=False)
+    device_fingerprint: Mapped[str | None] = mapped_column(String(255), nullable=True)
+    device_name: Mapped[str | None] = mapped_column(String(255), nullable=True)
+    # WebAuthn sign counter to detect cloned credentials
+    sign_count: Mapped[int] = mapped_column(Integer, default=0, server_default=text("0"), nullable=False)
```
yihua-zhuo, line 26: **WARNING** (suggestion)

**Issue:** Column has both Python `default=0` and `server_default=text("0")` — one is redundant; SQLAlchemy uses server_default for INSERT when no value is passed, making the Python default unreachable.

**Suggestion:** Remove `default=0` and keep only `server_default=text("0")`, or vice versa — pick one side of the duplicate.

src/db/models/user_credential.py
```diff
@@ -0,0 +1,56 @@
...
+    authenticator_type: Mapped[str] = mapped_column(
+        String(50), default="fido2", server_default=text("'fido2'"), nullable=False
```
yihua-zhuo, line 28: **WARNING** (suggestion)

**Issue:** Column has both Python `default="fido2"` and `server_default=text("'fido2'")` — redundant pair.

**Suggestion:** Remove `default="fido2"` and keep only `server_default=text("'fido2'")`.

src/db/models/user_credential.py
```diff
@@ -0,0 +1,56 @@
...
+    )
+    transports: Mapped[str | None] = mapped_column(Text, nullable=True)
+    enabled: Mapped[bool] = mapped_column(Boolean, default=True, server_default=text("true"), nullable=False)
```
yihua-zhuo, line 31: **WARNING** (suggestion)

**Issue:** Column has both Python `default=True` and `server_default=text("true")` — redundant pair.

**Suggestion:** Remove `default=True` and keep only `server_default=text("true")`.

src/db/models/user_credential.py
```diff
@@ -0,0 +1,56 @@
...
+    created_at: Mapped[datetime] = mapped_column(DateTime(timezone=True), server_default=func.now(), nullable=False)
+    last_used_at: Mapped[datetime | None] = mapped_column(DateTime(timezone=True), nullable=True)
+    updated_at: Mapped[datetime] = mapped_column(
+        DateTime(timezone=True), server_default=func.now(), onupdate=func.now(), nullable=False
+    )
+
+    __table_args__ = (Index("ix_user_credentials_user_enabled", "tenant_id", "user_id", "enabled"),)
+
+    def to_dict(self) -> dict:
```
yihua-zhuo, line 40: **INFO** (suggestion)

**Issue:** to_dict() and updated_at / created_at isoformat guards are identical to webauthn_challenge.py — extract a shared _format_datetime() helper to avoid drift.

**Suggestion:** Add `src/db/models/_utils.py` with a helper, e.g. `format_dt = lambda dt: dt.isoformat() if dt else None`.

src/db/models/webauthn_challenge.py
```diff
@@ -0,0 +1,51 @@
...
+        String(20), nullable=False, default="registration", server_default=text("'registration'")
+    )
+    credential_id: Mapped[str | None] = mapped_column(String(1024), nullable=True)
+    device_fingerprint: Mapped[str | None] = mapped_column(String(255), nullable=True)
+    expires_at: Mapped[datetime] = mapped_column(DateTime(timezone=True), nullable=False)
+    consumed: Mapped[bool] = mapped_column(Boolean, default=False, server_default=text("false"), nullable=False)
+    consumed_at: Mapped[datetime | None] = mapped_column(DateTime(timezone=True), nullable=True)
+    created_at: Mapped[datetime] = mapped_column(DateTime(timezone=True), server_default=func.now(), nullable=False)
+
+    __table_args__ = (
```
yihua-zhuo, line 33: **INFO** (suggestion)

**Issue:** Composite index ix_webauthn_challenges_consume includes `purpose` column which has a fixed default of 'registration' and no other values appear in the codebase — the index may never be selective on purpose alone.

**Suggestion:** Confirm `purpose` values are not extensible before dropping it from the index: `Index("ix_webauthn_challenges_consume", "tenant_id", "user_id", "consumed", "expires_at")`.

src/db/models/webauthn_challenge.py
```diff
@@ -0,0 +1,51 @@
...
+        Index("ix_webauthn_challenges_challenge", "tenant_id", "challenge"),
+        Index("ix_webauthn_challenges_expires", "tenant_id", "expires_at"),
+        Index("ix_webauthn_challenges_consume", "tenant_id", "user_id", "purpose", "consumed", "expires_at"),
+    )
+
+    def to_dict(self) -> dict:
```
yihua-zhuo, line 39: **INFO** (suggestion)

**Issue:** to_dict() datetime formatting pattern is duplicated from user_credential.py — extract a shared helper.

**Suggestion:** Extract shared datetime formatting to `src/db/models/_utils.py`.

src/services/auth/__init__.py
```diff
@@ -0,0 +1,19 @@
+"""Auth services sub-package.
+
+Modules:
+- token_service: Refresh token lifecycle (P0)
+- webauthn_service: WebAuthn registration/assertion (P2)
+- device_trust_service: Device trust + suspicious activity detection (P3)
+"""
+
+from services.auth.device_trust_service import DeviceTrustService, SuspiciousActivityReason, generate_device_fingerprint
```
yihua-zhuo, line 9: **WARNING** (suggestion)

**Issue:** SuspiciousActivityReason is imported but not listed in __all__, making it inaccessible via the package public API.

**Suggestion:** Add "SuspiciousActivityReason" to __all__ or remove the import if not needed by consumers.

src/services/auth/device_trust_service.py
```diff
@@ -0,0 +1,210 @@
...
+        self.session = session
+
+    async def is_device_trusted(self, user_id: int, tenant_id: int, device_fingerprint: str) -> bool:
+        """Check if a device is currently trusted for the given user."""
+        result = await self.session.execute(
+            select(DeviceTrustModel).where(
+                DeviceTrustModel.user_id == user_id,
+                DeviceTrustModel.tenant_id == tenant_id,
+                DeviceTrustModel.device_fingerprint == device_fingerprint,
+                DeviceTrustModel.trusted == True,  # noqa: E712
```
yihua-zhuo, line 63: **WARNING** (suggestion)

**Issue:** Explicit `== True` boolean comparison (PEP8 E712), same pattern repeated on lines 133 and 206.

**Suggestion:** Replace `DeviceTrustModel.trusted == True` with `DeviceTrustModel.trusted`.

src/services/auth/device_trust_service.py
```diff
@@ -0,0 +1,210 @@
...
+        return result.rowcount > 0
+
+    async def distrust_all_devices(self, user_id: int, tenant_id: int) -> int:
+        """Revoke trust for all devices of a user (e.g. after password change)."""
+        result = await self.session.execute(
+            update(DeviceTrustModel)
+            .where(
+                DeviceTrustModel.user_id == user_id,
+                DeviceTrustModel.tenant_id == tenant_id,
+                DeviceTrustModel.trusted == True,  # noqa: E712
```
yihua-zhuo, line 133: **WARNING** (suggestion)

**Issue:** Explicit `== True` boolean comparison (PEP8 E712).

**Suggestion:** Replace `DeviceTrustModel.trusted == True` with `DeviceTrustModel.trusted`.

src/services/auth/device_trust_service.py
```diff
@@ -0,0 +1,210 @@
...
+        return bool(reasons), reasons
+
+    async def get_trusted_devices(self, user_id: int, tenant_id: int) -> list[DeviceTrustModel]:
+        """List all trusted devices for a user."""
+        result = await self.session.execute(
+            select(DeviceTrustModel)
+            .where(
+                DeviceTrustModel.user_id == user_id,
+                DeviceTrustModel.tenant_id == tenant_id,
+                DeviceTrustModel.trusted == True,  # noqa: E712
```
yihua-zhuo, line 206: **WARNING** (suggestion)

**Issue:** Explicit `== True` boolean comparison (PEP8 E712).

**Suggestion:** Replace `DeviceTrustModel.trusted == True` with `DeviceTrustModel.trusted`.

src/services/auth/token_service.py
```diff
@@ -0,0 +1,245 @@
...
+ACCESS_TOKEN_EXPIRY_MINUTES = 10
+REFRESH_TOKEN_EXPIRY_DAYS = 7
+
+
+def generate_secure_token(length: int = 64) -> str:
+    """Generate a cryptographically secure random token."""
+    return secrets.token_urlsafe(length)
+
+
+def hash_token(token: str) -> str:
```
yihua-zhuo, line 28: **WARNING** (suggestion)

**Issue:** Token hashing uses SHA-256 without a per-token secret or memory-hard KDF; an attacker with table read access could brute-force short/weak tokens.

**Suggestion:** Consider using HMAC-SHA256 or a memory-hard KDF (e.g. argon2-cffi) keyed with a secret, so raw token lookup alone is insufficient to verify a candidate.

src/services/auth/token_service.py
```diff
@@ -0,0 +1,245 @@
...
+    """SHA-256 hash a token for storage (never store raw tokens)."""
+    return hashlib.sha256(token.encode()).hexdigest()
+
+
+def _as_utc(value: datetime) -> datetime:
+    """Treat naive DB datetimes as UTC and preserve aware datetimes."""
+    return value.replace(tzinfo=UTC) if value.tzinfo is None else value.astimezone(UTC)
+
+
+def _tenant_from_raw_token(raw_token: str) -> int | None:
```
yihua-zhuo, line 38: **WARNING** (suggestion)

**Issue:** Custom token format `{tenant_id}.{random}` with a bare `int()` parse differs from JWT convention; a zero-padded or otherwise formatted tenant_id would silently mis-parse.

**Suggestion:** Add bounds validation: ensure `len(tenant) < 10` or use a fixed-width prefix to prevent silent truncation/mis-parse of edge-case tenant IDs.

src/services/auth/token_service.py
```diff
@@ -0,0 +1,245 @@
...
+        """
+        tenant_id = _tenant_from_raw_token(raw_token)
+        if tenant_id is None:
+            return None
+        token_hash = hash_token(raw_token)
+        result = await self.session.execute(
+            select(RefreshTokenModel).where(
+                RefreshTokenModel.tenant_id == tenant_id,
+                RefreshTokenModel.token_hash == token_hash,
+                RefreshTokenModel.revoked == False,  # noqa: E712
```
yihua-zhuo, line 144: **WARNING** (suggestion)

**Issue:** Explicit `== False` boolean comparison (PEP8 E712); same pattern repeated on lines 165, 179, 193.

**Suggestion:** Replace `RefreshTokenModel.revoked == False` with `~RefreshTokenModel.revoked` or negate the condition appropriately.

src/services/auth/token_service.py
```diff
@@ -0,0 +1,245 @@
...
+        tenant_id = _tenant_from_raw_token(raw_token)
+        if tenant_id is None:
+            return False
+        token_hash = hash_token(raw_token)
+        result = await self.session.execute(
+            update(RefreshTokenModel)
+            .where(
+                RefreshTokenModel.tenant_id == tenant_id,
+                RefreshTokenModel.token_hash == token_hash,
+                RefreshTokenModel.revoked == False,  # noqa: E712
```
yihua-zhuo, line 165: **WARNING** (suggestion)

**Issue:** Explicit `== False` boolean comparison (PEP8 E712).

**Suggestion:** Replace `RefreshTokenModel.revoked == False` with `~RefreshTokenModel.revoked` or negate the condition appropriately.

src/services/auth/token_service.py
```diff
@@ -0,0 +1,245 @@
...
+        return result.rowcount > 0
+
+    async def revoke_all_user_tokens(self, user_id: int, tenant_id: int) -> int:
+        """Revoke all refresh tokens for a user (account-wide logout)."""
+        result = await self.session.execute(
+            update(RefreshTokenModel)
+            .where(
+                RefreshTokenModel.user_id == user_id,
+                RefreshTokenModel.tenant_id == tenant_id,
+                RefreshTokenModel.revoked == False,  # noqa: E712
```
yihua-zhuo, line 179: **WARNING** (suggestion)

**Issue:** Explicit `== False` boolean comparison (PEP8 E712).

**Suggestion:** Replace `RefreshTokenModel.revoked == False` with `~RefreshTokenModel.revoked` or negate the condition appropriately.

src/services/auth/token_service.py
```diff
@@ -0,0 +1,245 @@
...
+        return result.rowcount
+
+    async def get_active_sessions(self, user_id: int, tenant_id: int) -> list[RefreshTokenModel]:
+        """List all active (non-revoked, non-expired) sessions for a user."""
+        result = await self.session.execute(
+            select(RefreshTokenModel)
+            .where(
+                RefreshTokenModel.user_id == user_id,
+                RefreshTokenModel.tenant_id == tenant_id,
+                RefreshTokenModel.revoked == False,  # noqa: E712
```
yihua-zhuo, line 193: **WARNING** (suggestion)

**Issue:** Explicit `== False` boolean comparison (PEP8 E712).

**Suggestion:** Replace `RefreshTokenModel.revoked == False` with `~RefreshTokenModel.revoked` or negate the condition appropriately.

src/services/auth/token_service.py
```diff
@@ -0,0 +1,245 @@
...
+            )
+        )
+        old_model = result.scalar_one_or_none()
+        if old_model is None:
+            return None
+        expires_at = _as_utc(old_model.expires_at)
+        if expires_at < datetime.now(UTC):
+            return None
+
+        # Compute remaining TTL precisely so rotation never extends the original session.
```
yihua-zhuo, line 229: **INFO** (suggestion)

**Issue:** rotate_refresh_token performs two sequential DB writes (revoke old, create new) without explicit transaction coordination; under a partial failure the old token could be revoked while the new insert fails silently.

**Suggestion:** Rely on the router-layer transaction boundary (rule 121) — ensure both operations occur within the same request-scoped unit of work; verify the calling router does not commit between the revoke and create calls.

src/services/auth/webauthn_service.py
```diff
@@ -0,0 +1,434 @@
...
+        """
+        challenge = await self._create_challenge(
+            user_id=user_id,
+            tenant_id=tenant_id,
+            purpose="registration",
+            device_fingerprint=device_fingerprint,
+            ttl_seconds=REGISTRATION_CHALLENGE_TTL_SECONDS,
+        )
+
+        user_id_b64 = base64.urlsafe_b64encode(str(user_id).encode()).rstrip(b"=").decode()
```
yihua-zhuo, line 181: **WARNING** (suggestion)

**Issue:** Parameter `username` is passed to start_registration but not used — it is stored in the response dict but never read from the parameter.

**Suggestion:** Either remove the `username` parameter from start_registration (fetch from user model upstream) or validate it matches the stored user identity.

src/services/auth/webauthn_service.py
```diff
@@ -0,0 +1,434 @@
...
+            "user_id": user_id,
+            "username": username,
+            "credential_nickname": credential_nickname,
+        }
+
+    async def finish_registration(
+        self,
+        user_id: int,
+        tenant_id: int,
+        username: str,
```
yihua-zhuo, line 219: **WARNING** (suggestion)

**Issue:** Parameter `username` is declared in finish_registration but never used inside the method body.

**Suggestion:** Remove the unused `username` parameter or use it for an integrity check (e.g. verify it matches the user record loaded from the user_id FK).

src/services/auth/webauthn_service.py
```diff
@@ -0,0 +1,434 @@
...
+    ValidationException,
+)
+
+
+def generate_challenge(length: int = 64) -> str:
+    """Generate a cryptographically random WebAuthn challenge."""
+    return base64.urlsafe_b64encode(secrets.token_bytes(length)).rstrip(b"=").decode()
+
+
+def decode_credential_id(credential_id: str) -> bytes:
```
yihua-zhuo, line 36: **WARNING** (suggestion)

**Issue:** decode_credential_id uses base64url decode logic (including manual padding) for both credential IDs and general base64 input, making the function name misleading.

**Suggestion:** Rename decode_credential_id to a general-purpose base64url decoder if it is used for non-credential data, or split into a dedicated credential_id decoder and a separate base64 utility.

src/services/auth/webauthn_service.py
```diff
@@ -0,0 +1,434 @@
...
+        Raises:
+            ValidationException: If challenge is expired/already-used/invalid.
+            ValidationException: For malformed attestation data.
+        """
+        client_data = registration_response.get("response", {}).get("clientDataJSON", "")
+        attestation_object = registration_response.get("response", {}).get("attestationObject", "")
+
+        # Parse client data JSON
+        try:
+            if len(client_data) < 256:
```
yihua-zhuo, line 235: **WARNING** (suggestion)

**Issue:** clientDataJSON can be either base64url (when short) or standard base64 (when longer); this length-gate heuristic is non-standard and could misdecode a legitimately short standard-base64 JSON.

**Suggestion:** Normalize clientDataJSON to standard base64 before decoding: attempt urlsafe decode first, fall back to standard b64decode; document the heuristic or always use one format.

src/services/auth/webauthn_service.py
```diff
@@ -0,0 +1,434 @@
...
+        db_challenge = await self._consume_challenge(user_id, tenant_id, presented_challenge, purpose="registration")
+        if db_challenge is None:
+            raise ValidationException("Challenge expired, already used, or invalid")
+
+        self._validate_origin(origin)
+
+        # Extract credential ID
+        credential_id_raw = registration_response.get("credential", {}).get("id", "")
+        if len(credential_id_raw) < 256:
+            credential_id = encode_credential_id(decode_credential_id(credential_id_raw))
```
yihua-zhuo, line 256: **INFO** (suggestion)

**Issue:** Same length-gate heuristic issue for credential_id decoding (short = base64url, long = base64).

**Suggestion:** Pick a single encoding scheme and apply it consistently, or add a flag to explicitly select the decoding strategy.

src/services/auth/webauthn_service.py
```diff
@@ -0,0 +1,434 @@
...
+            ConflictException: If clone detection fires (sign_count not incremented).
+        """
+        credential_id = assertion_response.get("credentialId", "")
+        client_data_json_b64 = assertion_response.get("response", {}).get("clientDataJSON", "")
+        authenticator_data_b64 = assertion_response.get("response", {}).get("authenticatorData", "")
+
+        # Parse client data
+        try:
+            if len(client_data_json_b64) < 256:
+                client_data_bytes = decode_credential_id(client_data_json_b64)
```
yihua-zhuo, line 343: **WARNING** (suggestion)

**Issue:** Same length-gate heuristic for clientDataJSON decoding in verify_assertion.

**Suggestion:** Use a single consistent decoding strategy for clientDataJSON across finish_registration and verify_assertion.

src/services/auth/webauthn_service.py
```diff
@@ -0,0 +1,434 @@
...
+        sign_count = 0
+        try:
+            if len(authenticator_data_b64) < 128:
+                auth_data_bytes = decode_credential_id(authenticator_data_b64)
+            else:
+                auth_data_bytes = base64.b64decode(authenticator_data_b64)
+            if len(auth_data_bytes) >= 41:
+                sign_count = int.from_bytes(auth_data_bytes[37:41], byteorder="big")
+        except (binascii.Error, ValueError):
+            sign_count = 0
```
yihua-zhuo, line 386: **WARNING** (suggestion)

**Issue:** Silently defaulting sign_count to 0 on parse error means a malformed authenticator data byte string is treated as a legitimate zero counter, potentially defeating clone detection.

**Suggestion:** Raise ValidationException instead of silently falling back to 0 when authenticator data cannot be parsed, so the caller can request re-authentication rather than proceeding unsafely.

src/services/auth/webauthn_service.py
```diff
@@ -0,0 +1,434 @@
...
+        Returns the challenge model if valid (exists, not expired, not consumed).
+        Returns None if challenge is missing, expired, or already consumed.
+        """
+        result = await self.session.execute(
+            select(WebAuthnChallengeModel).where(
+                WebAuthnChallengeModel.user_id == user_id,
+                WebAuthnChallengeModel.tenant_id == tenant_id,
+                WebAuthnChallengeModel.challenge == challenge,
+                WebAuthnChallengeModel.purpose == purpose,
+                WebAuthnChallengeModel.consumed == False,  # noqa: E712
```
yihua-zhuo, line 128: **WARNING** (suggestion)

**Issue:** Explicit `== False` boolean comparison (PEP8 E712); same pattern on lines 151, 369, 412.

**Suggestion:** Replace `WebAuthnChallengeModel.consumed == False` with `~WebAuthnChallengeModel.consumed` or equivalent.

src/services/auth/webauthn_service.py
```diff
@@ -0,0 +1,434 @@
...
+
+        return model
+
+    async def _cleanup_expired_challenges(self, tenant_id: int) -> int:
+        """Delete all consumed-and-expired challenges. Call periodically to keep table small."""
+        result = await self.session.execute(
+            delete(WebAuthnChallengeModel).where(
+                WebAuthnChallengeModel.expires_at < datetime.now(UTC),
+                WebAuthnChallengeModel.tenant_id == tenant_id,
+                WebAuthnChallengeModel.consumed == True,  # noqa: E712
```
yihua-zhuo, line 151: **WARNING** (suggestion)

**Issue:** Explicit `== True` boolean comparison (PEP8 E712).

**Suggestion:** Replace `WebAuthnChallengeModel.consumed == True` with `WebAuthnChallengeModel.consumed`.

src/services/auth/webauthn_service.py
```diff
@@ -0,0 +1,434 @@
...
+        if db_challenge.credential_id and db_challenge.credential_id != credential_id:
+            raise ValidationException("Credential ID mismatch")
+
+        # Find credential
+        result = await self.session.execute(
+            select(UserCredentialModel).where(
+                UserCredentialModel.credential_id == credential_id,
+                UserCredentialModel.user_id == user_id,
+                UserCredentialModel.tenant_id == tenant_id,
+                UserCredentialModel.enabled == True,  # noqa: E712
```
yihua-zhuo, line 369: **WARNING** (suggestion)

**Issue:** Explicit `== True` boolean comparison (PEP8 E712).

**Suggestion:** Replace `UserCredentialModel.enabled == True` with `UserCredentialModel.enabled`.

src/services/auth/webauthn_service.py
```diff
@@ -0,0 +1,434 @@
...
+    # -------------------------------------------------------------------------
+
+    async def get_user_credentials(self, user_id: int, tenant_id: int) -> list[UserCredentialModel]:
+        """List all WebAuthn credentials for a user."""
+        result = await self.session.execute(
+            select(UserCredentialModel)
+            .where(
+                UserCredentialModel.user_id == user_id,
+                UserCredentialModel.tenant_id == tenant_id,
+                UserCredentialModel.enabled == True,  # noqa: E712
```
yihua-zhuo, line 412: **WARNING** (suggestion)

**Issue:** Explicit `== True` boolean comparison (PEP8 E712).

**Suggestion:** Replace `UserCredentialModel.enabled == True` with `UserCredentialModel.enabled`.

PR #357, opened 2026-05-13
Auto: implement issue #57

### LINKED ISSUE #57: [UI] AI Chat Assistant — Conversational AI interface
## Problem

AI capabilities exist in backend (#43) but no chat UI.

## Dependencies

**Blocked by:** #41 (AI Agent Framework — must complete first)
**Blocked by:** #43 (AI Chat Backend — must complete first)

### Chat Interface (`/ai/chat`)
- [ ] ChatGPT-style interface
- [ ] Message history with timestamps
- [ ] Typing indicator during AI response
- [ ] Markdown rendered in responses
- [ ] Code blocks with syntax highlighting

### Message Features
- [ ] Send with Enter, newline with Shift+Enter
- [ ] Copy message / copy code
- [ ] Regenerate response
- [ ] Clear conversation button

### Context Integration
- [ ] Quick actions: "Show my top customers" / "Pipeline summary"
- [ ] Suggestion chips below input
- [ ] Link to related entities in responses

### Sessions
- [ ] Sidebar: past sessions list
- [ ] New conversation button
- [ ] Delete conversation
- [ ] Session title auto-generated

yihua-zhuo:
Closes #57

Plan: `.plans/issue-57.md`

1. **Plan** — Claude wrote a detailed implementation plan to `.plans/issue-57.md`.

Please review before merging.

## Summary by CodeRabbit

* **New Features**
  * Full AI chat experience with persistent sessions, sidebar session list, session restore/redirect, create/open/delete/clear conversations, regenerate replies, and typing indicator.
  * Rich message rendering: Markdown, GitHub-flavored extensions, code syntax highlighting, per-block copy, suggestion buttons, and Enter-to-send.

* **Bug Fixes / UX**
  * Floating AI panel and chat UI now use the real chat backend (no simulated responses).

* **Tests**
  * New unit and integration tests covering chat service and API flows.

* **Documentation**
  * Added implementation plan and acceptance criteria for the AI chat feature.

[Review Change Stack](https://app.coderabbit.ai/change-stack/yihua-zhuo/agent-job/pull/357)

### DIFF
--- a/tests/unit/test_customer_service.py
+++ b/tests/unit/test_customer_service.py
@@ -36,96 +36,69 @@ class TestCustomerCreateDTO:
     """Tests for CustomerCreateDTO Pydantic model."""
 
     def test_required_fields_only(self):
-        """CustomerCreateDTO with only required fields."""
         dto = CustomerCreateDTO(name="Alice", email="alice@example.com")
         assert dto.name == "Alice"
         assert dto.email == "alice@example.com"
-        assert dto.phone is None
-        assert dto.company is None
         assert dto.status == "lead"
         assert dto.owner_id == 0
-        assert dto.tags == []
 
     def test_all_fields(self):
-        """CustomerCreateDTO with all fields specified."""
         dto = CustomerCreateDTO(
             name="Bob",
             email="bob@example.com",
-            phone="13800138000",
-            company="Acme Corp",
-            status="customer",
-            owner_id=42,
-            tags=["vip", "enterprise"],
+            status="opportunity",
+            owner_id=5,
+            tags=["vip", "referral"],
         )
         assert dto.name == "Bob"
         assert dto.email == "bob@example.com"
-        assert dto.phone == "13800138000"
-        assert dto.company == "Acme Corp"
-        assert dto.status == "customer"
-        assert dto.owner_id == 42
-        assert dto.tags == ["vip", "enterprise"]
+        assert dto.status == "opportunity"
+        assert dto.owner_id == 5
+        assert dto.tags == ["vip", "referral"]
 
     def test_to_dict(self):
-        """to_dict produces expected flat dict for CustomerService."""
-        dto = CustomerCreateDTO(
-            name="Carol",
-            email="carol@example.com",
-            phone="13900139000",
-            company="Beta Ltd",
-            status="opportunity",
-            owner_id=7,
-            tags=["trial"],
-        )
-        d = dto.to_dict()
-        assert d == {
-            "name": "Carol",
-            "email": "carol@example.com",
-            "phone": "13900139000",
-            "company": "Beta Ltd",
-            "status": "opportunity",
-            "owner_id": 7,
-            "tags": ["trial"],
-        }
+        dto = CustomerCreateDTO(name="Charlie", email="charlie@example.com")
+        result = dto.to_dict()
+        assert result["name"] == "Charlie"
+        assert result["email"] == "charlie@example.com"
+        assert result["status"] == "lead"
+        assert result["owner_id"] == 0
+        assert result["tags"] == []
 
     def test_from_dict(self):
-        """from_dict round-trips through to_dict."""
-        original = CustomerCreateDTO(
-            name="Dave",
-            email="dave@example.com",
-            phone="13700137000",
-            company="Gamma",
-            status="lead",
-            owner_id=3,
-            tags=["prospect"],
-        )
-        reconstructed = CustomerCreateDTO.from_dict(original.to_dict())
-        assert reconstructed.name == original.name
-        assert reconstructed.email == original.email
-        assert reconstructed.phone == original.phone
-        assert reconstructed.company == original.company
-        assert reconstructed.status == original.status
-        assert reconstructed.owner_id == original.owner_id
-        assert reconstructed.tags == original.tags
+        data = {
+            "name": "Diana",
+            "email": "diana@example.com",
+            "status": "inactive",
+            "owner_id": 3,
+            "tags": ["newsletter"],
+        }
+        dto = CustomerCreateDTO.from_dict(data)
+        assert dto.name == "Diana"
+        assert dto.email == "diana@example.com"
+        assert dto.status == "inactive"
+        assert dto.owner_id == 3
+        assert dto.tags == ["newsletter"]
 
     def test_from_dict_missing_name_raises(self):
-        """from_dict raises ValueError when name is missing."""
-        with pytest.raises(ValueError, match="name is required"):
-            CustomerCreateDTO.from_dict({"email": "test@example.com"})
+        with pytest.raises(ValueError):
+            CustomerCreateDTO.from_dict({"email": "no-name@example.com"})
 
     def test_from_dict_missing_email_raises(self):
-        """from_dict raises ValueError when email is missing."""
-        with pytest.raises(ValueError, match="email is required"):
-            CustomerCreateDTO.from_dict({"name": "Test"})
+        with pytest.raises(ValueError):
+            CustomerCreateDTO.from_dict({"name": "No Email"})
 
     def test_default_status_is_lead(self):
-        """Default status when not specified."""
         dto = CustomerCreateDTO(name="Eve", email="eve@example.com")
         assert dto.status == "lead"
 
     def test_default_owner_id_is_zero(self):
-        """Default owner_id when not specified."""
         dto = CustomerCreateDTO(name="Frank", email="frank@example.com")
         assert dto.owner_id == 0
+
+    def test_to_dict_email_normalized(self):
+        dto = CustomerCreateDTO(name="Grace", email="  grace@example.com  ")
+        assert dto.email == "grace@example.com"
 
 
 class TestCreateCustomerService:
@@ -151,11 +124,10 @@ class TestCreateCustomerService:
         # Services flush; the router-bound get_db dependency owns commit (rule 121/122).
         mock_db_session.flush.assert_called_once()
         assert result.name == "Test"
-        assert result.email == "test@example.com"
 
     @pytest.mark.asyncio
     async def test_create_customer_accepts_dto(self, mock_db_session):
-        """create_customer accepts a CustomerCreateDTO instance."""
+        """create_customer works with a CustomerCreateDTO instance."""
         service = CustomerService(mock_db_session)
 
         async def fake_refresh(obj):
--- a/tests/unit/test_customer_service.py
+++ b/tests/unit/test_customer_service.py
@@ -36,69 +36,96 @@ class TestCustomerCreateDTO:
     """Tests for CustomerCreateDTO Pydantic model."""
 
     def test_required_fields_only(self):
+        """CustomerCreateDTO with only required fields."""
         dto = CustomerCreateDTO(name="Alice", email="alice@example.com")
         assert dto.name == "Alice"
         assert dto.email == "alice@example.com"
+        assert dto.phone is None
+        assert dto.company is None
         assert dto.status == "lead"
         assert dto.owner_id == 0
+        assert dto.tags == []
 
     def test_all_fields(self):
+        """CustomerCreateDTO with all fields specified."""
         dto = CustomerCreateDTO(
             name="Bob",
             email="bob@example.com",
-            status="opportunity",
-            owner_id=5,
-            tags=["vip", "referral"],
+            phone="13800138000",
+            company="Acme Corp",
+            status="customer",
+            owner_id=42,
+            tags=["vip", "enterprise"],
         )
         assert dto.name == "Bob"
         assert dto.email == "bob@example.com"
-        assert dto.status == "opportunity"
-        assert dto.owner_id == 5
-        assert dto.tags == ["vip", "referral"]
+        assert dto.phone == "13800138000"
+        assert dto.company == "Acme Corp"
+        assert dto.status == "customer"
+        assert dto.owner_id == 42
+        assert dto.tags == ["vip", "enterprise"]
 
     def test_to_dict(self):
-        dto = CustomerCreateDTO(name="Charlie", email="charlie@example.com")
-        result = dto.to_dict()
-        assert result["name"] == "Charlie"
-        assert result["email"] == "charlie@example.com"
-        assert result["status"] == "lead"
-        assert result["owner_id"] == 0
-        assert result["tags"] == []
+        """to_dict produces expected flat dict for CustomerService."""
+        dto = CustomerCreateDTO(
+            name="Carol",
+            email="carol@example.com",
+            phone="13900139000",
+            company="Beta Ltd",
+            status="opportunity",
+            owner_id=7,
+            tags=["trial"],
+        )
+        d = dto.to_dict()
+        assert d == {
+            "name": "Carol",
+            "email": "carol@example.com",
+            "phone": "13900139000",
+            "company": "Beta Ltd",
+            "status": "opportunity",
+            "owner_id": 7,
+            "tags": ["trial"],
+        }
 
     def test_from_dict(self):
-        data = {
-            "name": "Diana",
-            "email": "diana@example.com",
-            "status": "inactive",
-            "owner_id": 3,
-            "tags": ["newsletter"],
-        }
-        dto = CustomerCreateDTO.from_dict(data)
-        assert dto.name == "Diana"
-        assert dto.email == "diana@example.com"
-        assert dto.status == "inactive"
-        assert dto.owner_id == 3
-        assert dto.tags == ["newsletter"]
+        """from_dict round-trips through to_dict."""
+        original = CustomerCreateDTO(
+            name="Dave",
+            email="dave@example.com",
+            phone="13700137000",
+            company="Gamma",
+            status="lead",
+            owner_id=3,
+            tags=["prospect"],
+        )
+        reconstructed = CustomerCreateDTO.from_dict(original.to_dict())
+        assert reconstructed.name == original.name
+        assert reconstructed.email == original.email
+        assert reconstructed.phone == original.phone
+        assert reconstructed.company == original.company
+        assert reconstructed.status == original.status
+        assert reconstructed.owner_id == original.owner_id
+        assert reconstructed.tags == original.tags
 
     def test_from_dict_missing_name_raises(self):
-        with pytest.raises(ValueError):
-            CustomerCreateDTO.from_dict({"email": "no-name@example.com"})
+        """from_dict raises ValueError when name is missing."""
+        with pytest.raises(ValueError, match="name is required"):
+            CustomerCreateDTO.from_dict({"email": "test@example.com"})
 
     def test_from_dict_missing_email_raises(self):
-        with pytest.raises(ValueError):
-            CustomerCreateDTO.from_dict({"name": "No Email"})
+        """from_dict raises ValueError when email is missing."""
+        with pytest.raises(ValueError, match="email is required"):
+            CustomerCreateDTO.from_dict({"name": "Test"})
 
     def test_default_status_is_lead(self):
+        """Default status when not specified."""
         dto = CustomerCreateDTO(name="Eve", email="eve@example.com")
         assert dto.status == "lead"
 
     def test_default_owner_id_is_zero(self):
+        """Default owner_id when not specified."""
         dto = CustomerCreateDTO(name="Frank", email="frank@example.com")
         assert dto.owner_id == 0
-
-    def test_to_dict_email_normalized(self):
-        dto = CustomerCreateDTO(name="Grace", email="  grace@example.com  ")
-        assert dto.email == "grace@example.com"
 
 
 class TestCreateCustomerService:
@@ -124,10 +151,11 @@ class TestCreateCustomerService:
         # Services flush; the router-bound get_db dependency owns commit (rule 121/122).
         mock_db_session.flush.assert_called_once()
         assert result.name == "Test"
+        assert result.email == "test@example.com"
 
     @pytest.mark.asyncio
     async def test_create_customer_accepts_dto(self, mock_db_session):
-        """create_customer works with a CustomerCreateDTO instance."""
+        """create_customer accepts a CustomerCreateDTO instance."""
         service = CustomerService(mock_db_session)
 
         async def fake_refresh(obj):


frontend/package.json
```diff
@@ -22,7 +22,10 @@
     "react": "19.2.4",
     "react-dom": "19.2.4",
     "react-hook-form": "^7.75.0",
+    "react-markdown": "^10.1.0",
+    "react-syntax-highlighter": "^15.0.0",
```
yihua-zhuo: **CRITICAL** (critical)

**Issue:** package.json declares react-syntax-highlighter ^15.0.0 but package-lock.json has ^16.1.1 — the two are out of sync and one must be corrected.

**Suggestion:** Update the version to ^16.1.1 to match package-lock.json and re-run `npm install` to regenerate a consistent lock file.

frontend/src/app/(app)/ai/[sessionId]/page.tsx
```diff
@@ -0,0 +1,148 @@
...
+import { useChatMessages, useSendChatMessage } from "@/lib/api/queries";
+import { useChatStore, Message } from "@/lib/store/chat-store";
+import { ChatInput } from "@/lib/components/chat-input";
+import { ChatMessage } from "@/lib/components/chat-message";
+import { ChatSessionSidebar } from "@/lib/components/chat-session-sidebar";
+import { TypingIndicator } from "@/lib/components/typing-indicator";
+
+export default function AISessionPage() {
+  const { sessionId } = useParams<{ sessionId: string }>();
+  const sessionIdNum = Number.parseInt(sessionId, 10);
```
Comment (yihua-zhuo) [inline frontend/src/app/(app)/ai/[sessionId]/page.tsx]: **CRITICAL** (critical)

**Issue:** `Number.isInteger(sessionIdNum)` is always true because `Number.parseInt` always returns a number (including NaN), never an integer type. NaN values silently pass the guard and reach query/UI code.

**Suggestion:** Use `Number.isNaN(sessionIdNum) || sessionIdNum <= 0` instead of `!Number.isInteger(sessionIdNum) || sessionIdNum <= 0`.

frontend/src/lib/components/chat-message.tsx
```diff
@@ -0,0 +1,263 @@
...
+  timestamp?: Date;
+}
+
+interface ChatMessageProps {
+  message: Message;
+  onRegenerate?: () => void;
+  showRegenerate?: boolean;
+}
+
+export function ChatMessage({ message, onRegenerate, showRegenerate = false }: ChatMessageProps) {
```
yihua-zhuo: **CRITICAL** (critical)

**Issue:** ChatMessage is defined twice in the same file (lines 23 and 144). The first definition (lines 23–142) is dead code shadowed by the second export.

**Suggestion:** Delete the entire first ChatMessage function (lines 23–142) and keep only the second definition starting at line 144.

src/api/routers/ai.py
```diff
@@ -0,0 +1,194 @@
...
+        "id": msg.id,
+        "session_id": msg.session_id,
+        "tenant_id": msg.tenant_id,
+        "role": msg.role,
+        "content": msg.content,
+        "created_at": msg.created_at.isoformat() if msg.created_at else None,
+    }
+
+
+def _paginated_sessions(items, total, page, page_size) -> dict:
```
yihua-zhuo: **CRITICAL** (critical)

**Issue:** _paginated_sessions divides by page_size without guarding against zero or negative values despite page_size being constrained by Query(ge=1) at every call site — the function's own guard is absent and callers could bypass it.

**Suggestion:** Add guard in _paginated_sessions: total_pages = (total + page_size - 1) // page_size if page_size > 0 else 0

src/services/ai_service.py
```diff
@@ -0,0 +1,156 @@
...
+                and_(ChatSessionModel.id == session_id, ChatSessionModel.tenant_id == tenant_id)
+            )
+        )
+        db_session = session_result.scalar_one_or_none()
+        if db_session is not None:
+            if db_session.title is None and role == "user":
+                db_session.title = self.generate_title([{"role": "user", "content": content}])
+            db_session.updated_at = now
+
+        await self.session.flush()
```
yihua-zhuo: **CRITICAL** (critical)

**Issue:** save_message calls flush() then refresh() — same double+commit triple pattern violating rule 123.

**Suggestion:** Remove `await self.session.refresh(message)` — flush() populates the ID, router commit handles final visibility.

src/services/ai_service.py
```diff
@@ -0,0 +1,156 @@
...
+        return list(result.scalars().all())
+
+    async def update_session_title(
+        self, session_id: int, tenant_id: int, title: str
+    ) -> ChatSessionModel:
+        """Update the title of an existing session."""
+        session = await self.get_session(session_id, tenant_id)
+        session.title = title
+        session.updated_at = datetime.now(UTC)
+        await self.session.flush()
```
yihua-zhuo: **CRITICAL** (critical)

**Issue:** update_session_title calls flush() then refresh() — same double+commit triple pattern violating rule 123.

**Suggestion:** Remove `await self.session.refresh(session)` — flush() is sufficient per rule 122.

alembic/env.py
```diff
@@ -18,6 +18,8 @@
     ApiClientModel,
     CampaignEventModel,
     CampaignModel,
+    ChatMessageModel,
+    ChatSessionModel,
     CustomerModel,
     DashboardModel,
```
yihua-zhuo, line 22: **WARNING** (suggestion)

**Issue:** tenant_id is stored on ChatMessageModel but the session's message relationship (ChatSessionModel.messages) traverses it without a tenant guard — a service-layer bug could return another tenant's messages via the relationship.

**Suggestion:** Service-layer queries for messages should always filter by tenant_id explicitly; do not rely solely on the session's tenant_id to scope messages.

frontend/src/app/(app)/ai/[sessionId]/page.tsx
```diff
@@ -0,0 +1,148 @@
...
+  const prevMsgCountRef = useRef<number>(0);
+
+  const { isLoading: storeLoading, messages: messagesBySession, addMessage, setMessages, setLoading } =
+    useChatStore();
+
+  const messages = Number.isInteger(sessionIdNum) && sessionIdNum > 0
+    ? (messagesBySession[sessionIdNum] ?? [])
+    : [];
+
+  const { data: messagesData, isLoading: msgsLoading } = useChatMessages(sessionIdNum);
```
Comment (yihua-zhuo) [inline frontend/src/app/(app)/ai/[sessionId]/page.tsx]: **WARNING** (suggestion)

**Issue:** `useChatMessages` query has no `onError` handler — if the API call fails, the error is silently swallowed and the UI shows the empty (no-session) state.

**Suggestion:** Add `onError: (err) => console.error("[AISessionPage] loadMessages failed:", err)` to the `useChatMessages` call.

frontend/src/app/(app)/ai/[sessionId]/page.tsx
```diff
@@ -0,0 +1,148 @@
...
+  }, [messagesData, sessionIdNum, setMessages]);
+
+  useEffect(() => {
+    if (messages.length !== prevMsgCountRef.current) {
+      prevMsgCountRef.current = messages.length;
+      bottomRef.current?.scrollIntoView({ behavior: "smooth" });
+    }
+  }, [messages.length]);
+
+  const handleSend = useCallback(async (text: string, options?: { regenerate?: boolean }) => {
```
Comment (yihua-zhuo) [inline frontend/src/app/(app)/ai/[sessionId]/page.tsx]: **WARNING** (suggestion)

**Issue:** The `options.regenerate` parameter is passed through `handleSend` but never forwarded to `sendMessageMutation.mutateAsync` — the `regenerate` flag is silently dropped and has no effect.

**Suggestion:** Pass options to mutateAsync: `mutateAsync({ sessionId: sessionIdNum, content: text, options })`

frontend/src/app/(app)/ai/[sessionId]/page.tsx
```diff
@@ -0,0 +1,148 @@
...
+        id: String(m.id),
+        role: m.role as "user" | "assistant",
+        content: m.content,
+        timestamp: new Date(m.created_at),
+      }));
+      setMessages(sessionIdNum, loaded);
+    }
+  }, [messagesData, sessionIdNum, setMessages]);
+
+  useEffect(() => {
```
Comment (yihua-zhuo) [inline frontend/src/app/(app)/ai/[sessionId]/page.tsx]: **INFO** (suggestion)

**Issue:** Scroll-into-view on every message count change can conflict with user scroll position during active typing, causing jarring jumps.

**Suggestion:** Guard scroll to only trigger when the user is near the bottom, or debounce the scroll call.

frontend/src/app/(app)/ai/[sessionId]/page.tsx
```diff
@@ -0,0 +1,148 @@
...
+  const { sessionId } = useParams<{ sessionId: string }>();
+  const sessionIdNum = Number.parseInt(sessionId, 10);
+  const router = useRouter();
+  const bottomRef = useRef<HTMLDivElement>(null);
+  const prevMsgCountRef = useRef<number>(0);
+
+  const { isLoading: storeLoading, messages: messagesBySession, addMessage, setMessages, setLoading } =
+    useChatStore();
+
+  const messages = Number.isInteger(sessionIdNum) && sessionIdNum > 0
```
Comment (yihua-zhuo) [inline frontend/src/app/(app)/ai/[sessionId]/page.tsx]: **INFO** (suggestion)

**Issue:** Duplicate NaN validation — `Number.isInteger(sessionIdNum)` check is performed both at line 22 (guard before accessing `messagesBySession`) and again at line 30 in useEffect.

**Suggestion:** Define the guard once, e.g. `const isValidId = !Number.isNaN(sessionIdNum) && sessionIdNum > 0`, and reuse it.

frontend/src/app/(app)/ai/page.tsx
```diff
@@ -1,10 +1,42 @@
...
+  const { activeSessionId, setActiveSession } = useChatStore();
+  const createSession = useCreateChatSession();
+  const hasRunRef = useRef(false);
+
+  useEffect(() => {
+    if (activeSessionId != null && activeSessionId > 0) {
+      router.replace(`/ai/sessions/${activeSessionId}`);
+      return;
+    }
+    if (createSession.isPending || hasRunRef.current) return;
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** `hasRunRef.current` is set to true on entry and reset to false only in `onError`. If `onError` fires, subsequent renders set the ref back to true immediately, then attempt creation again — causing a retry loop or race condition.

**Suggestion:** Guard the entire effect body with `if (hasRunRef.current) return; hasRunRef.current = true;` and do not reset the ref in `onError`. Alternatively, handle retries explicitly with a state flag.

frontend/src/lib/api/ai.ts
```diff
@@ -0,0 +1,83 @@
+import { apiClient } from "../api-client";
+
+export interface ChatSession {
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** `ChatSession` and `ChatMessage` interfaces in `ai.ts` are duplicated in `queries.ts` via the `ChatMessage` usage. Duplication risks the two definitions drifting.

**Suggestion:** Move shared type definitions (ChatSession, ChatMessage) to a central types file and import from both locations.

frontend/src/lib/api/ai.ts
```diff
@@ -0,0 +1,83 @@
...
+  data: T;
+}
+
+interface SendMessageResponse {
+  user_message: ChatMessage;
+  assistant_message: ChatMessage;
+}
+
+export const aiApi = {
+  createSession(token?: string) {
```
yihua-zhuo: **INFO** (suggestion)

**Issue:** No timeout is set on any API call — long-running or hung requests have no client-side deadline.

**Suggestion:** Add an `AbortSignal.timeout(10000)` or equivalent to the underlying fetch/client calls to fail fast on unresponsive endpoints.

frontend/src/lib/components/ai-panel.tsx
```diff
@@ -1,89 +1,119 @@
...
+    clearSession,
+  } = useChatStore();
+
+  const messages = activeSessionId ? (messagesBySession[activeSessionId] ?? []) : [];
 
+  const createSessionMutation = useCreateChatSession();
+  const sendMessageMutation = useSendChatMessage();
+
+  // Create a session on first open if none exists
   useEffect(() => {
```
yihua-zhuo: **INFO** (suggestion)

**Issue:** useEffect dependency array includes setActiveSession but omits hasRunRef (a ref, correctly excluded) and activeSessionId changes while isPending is true could cause double-trigger race conditions.

**Suggestion:** Add a ref to track in-flight session creation and gate the effect more robustly, or use useCallback + useRef for the guard.

frontend/src/lib/components/ai-panel.tsx
```diff
@@ -1,89 +1,119 @@
...
+    if (!activeSessionId || isLoading || sendMessageMutation.isPending) return;
+    const id = String(Date.now());
+    addMessage(activeSessionId, { id, role: "user", content: text, timestamp: new Date() });
+    setLoading(true);
+    try {
+      await sendMessageMutation.mutateAsync({
+        sessionId: activeSessionId,
+        content: text,
+      });
+    } catch (err) {
```
yihua-zhuo: **INFO** (suggestion)

**Issue:** Error adding a fallback 'sorry' assistant message does not notify the user why it failed — they only see a generic error reply.

**Suggestion:** Consider surfacing a transient error banner or toast inside the panel instead of injecting a misleading assistant message.

frontend/src/lib/components/chat-message.tsx
```diff
@@ -0,0 +1,263 @@
...
+        </div>
+      )}
+      <div className="max-w-[80%] rounded-xl px-3 py-2 text-sm">
+        {message.role === "assistant" ? (
+          <div className="space-y-1">
+            <div className="space-y-2">
+              <ReactMarkdown
+                remarkPlugins={[remarkGfm]}
+                components={{
+                  code({ className, children, ...props }) {
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** The code component override does not forward ref, breaking ref-based features and causing React warnings in StrictMode.

**Suggestion:** Use `React.ForwardedRef` or a ref-forwarding wrapper: `const CodeBlock = React.forwardRef(({ className, children, ...props }, ref) => { ... });` and update the `components` map accordingly.

frontend/src/lib/components/chat-message.tsx
```diff
@@ -0,0 +1,263 @@
...
+                      );
+                    }
+                    return (
+                      <div className="relative group">
+                        <button
+                          type="button"
+                          onClick={async () => {
+                            try {
+                              await navigator.clipboard.writeText(code);
+                            } catch {
```
yihua-zhuo: **INFO** (suggestion)

**Issue:** Clipboard write failure is silently swallowed — user has no indication if copy failed.

**Suggestion:** Set a transient 'error' state and show a brief inline message or fallback icon instead of no-op.

frontend/src/lib/components/chat-session-sidebar.tsx
```diff
@@ -0,0 +1,151 @@
...
+          <div className="px-3 py-2 text-xs text-muted-foreground">Loading…</div>
+        )}
+        {sessions.length === 0 && !isLoading && (
+          <div className="px-3 py-4 text-center text-xs text-muted-foreground">
+            No conversations yet.
+            <br />
+            Start one above!
+          </div>
+        )}
+        {sessions.filter(Boolean).map((s) => (
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** sessions.filter(Boolean) silently filters null/undefined entries — if the API can return null sessions that's a data-quality issue that should be fixed upstream, not patched downstream.

**Suggestion:** Remove the filter and assert non-null: `sessions.map((s) =>` with a type assertion `s!` if TypeScript is strict, or fix the API to never return nulls.

frontend/src/lib/components/chat-session-sidebar.tsx
```diff
@@ -0,0 +1,151 @@
...
+                : "hover:bg-muted text-foreground"
+            )}
+          >
+            <div className="min-w-0 flex-1">
+              <p className="truncate text-sm font-medium">{s.title ?? "New conversation"}</p>
+              <p className="text-xs text-muted-foreground truncate">
+                {formatRelativeTime(s.updated_at)}
+              </p>
+            </div>
+            <span
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** Delete trigger is a <span> with role="button" instead of a native <button> element — fails accessibility for keyboard users and screen readers.

**Suggestion:** Replace the <span> with a <button type="button"> element and remove the redundant role attribute.

frontend/src/lib/store/chat-store.ts
```diff
@@ -0,0 +1,57 @@
...
+    set((state) => ({
+      messages: {
+        ...state.messages,
+        [sessionId]: msgs,
+      },
+    })),
+
+  setLoading: (loading) => set({ isLoading: loading }),
+
+  clearSession: (sessionId) =>
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** clearSession clears messages for a session but leaves activeSessionId dangling if that session was active, causing the UI to show a ghost active state.

**Suggestion:** In clearSession, also reset activeSessionId to null when clearing the currently active session: if (state.activeSessionId === sessionId) return { messages: { ...state.messages, [sessionId]: [] }, activeSessionId: null, isLoading: false }

src/api/routers/ai.py
```diff
@@ -0,0 +1,194 @@
...
+    items, total = await service.list_sessions(
+        tenant_id=ctx.tenant_id,
+        user_id=ctx.user_id,
+        page=page,
+        page_size=page_size,
+    )
+    return _paginated_sessions(items, total, page, page_size)
+
+
+@ai_router.delete("/sessions/{session_id}", status_code=204)
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** DELETE endpoint declares status_code=204 No Content but returns a JSON body {"success": true, ...}, violating HTTP semantics — the 204 response must have no body.

**Suggestion:** Remove the return statement body and let FastAPI emit 204 with no body, or change status_code to 200.

src/api/routers/ai.py
```diff
@@ -0,0 +1,194 @@
...
+):
+    """Delete an AI chat session and all its messages."""
+    if ctx.tenant_id is None:
+        raise HTTPException(status_code=401, detail="tenant_id is required in token")
+    service = AIService(session)
+    await service.delete_session(session_id, tenant_id=ctx.tenant_id)
+    return {"success": True, "message": "Session deleted"}
+
+
+@ai_router.get("/sessions/{session_id}/messages")
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** session_id path parameter is not validated before being passed to the service — a non-integer value will raise a Pydantic validation error (acceptable), but zero or negative integers will pass validation and hit the DB with no guard.

**Suggestion:** Add Path(..., ge=1) validation to the session_id parameter: async def get_messages(session_id: int = Path(..., ge=1), ...)

src/services/ai_service.py
```diff
@@ -0,0 +1,156 @@
...
+
+    async def save_message(
+        self,
+        session_id: int,
+        role: str,
+        content: str,
+        tenant_id: int,
+    ) -> ChatMessageModel:
+        """Save a single message and update the session's updated_at timestamp."""
+        if role not in ("user", "assistant"):
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** Role validation raises ValueError instead of ValidationException, violating the project convention that services raise AppException subclasses.

**Suggestion:** Change `raise ValueError(...)` to `raise ValidationException('role must be \'user\' or \'assistant\'')`.

tests/integration/test_ai_chat_integration.py
```diff
@@ -0,0 +1,238 @@
+"""Integration tests for AI Chat endpoints.
+
+Uses db_schema, tenant_id, async_session fixtures from conftest.py.
+Tests: session create → message send → message list → session delete.
+Rule: integration test services must not call commit() — get_db already commits.
+"""
+
```
yihua-zhuo: **INFO** (suggestion)

**Issue:** Module docstring states 'integration test services must not call commit()' but line 357 of conftest.py calls commit in the auth fixture setup — this is the correct pattern for fixture setup but the docstring should note that the bare async_session fixture requires explicit commit, unlike router-driven requests.

**Suggestion:** Clarify the module docstring: 'Services must not call commit() in test body. Fixture setup (auth_headers_web) may call commit to persist data needed by subsequent service calls within the same fixture.'

tests/integration/test_ai_chat_integration.py
```diff
@@ -0,0 +1,238 @@
...
+        with pytest.raises(NotFoundException):
+            await svc.get_messages(session_id=session1.id, tenant_id=tenant_id_2)
+
+    @pytest.mark.asyncio
+    async def test_generate_title(self, db_schema, tenant_id, async_session):
+        from services.ai_service import AIService
+
+        svc = AIService(async_session)
+
+        title = svc.generate_title([{"role": "user", "content": "Show my customers"}])
```
yihua-zhuo: **INFO** (suggestion)

**Issue:** test_generate_title is a pure unit test testing AIService.generate_title() — it does not use the database, so it belongs in tests/unit/test_ai_service.py rather than the integration test file.

**Suggestion:** Move test_generate_title to tests/unit/test_ai_service.py to keep integration tests focused on DB-bound behavior.

tests/integration/test_ai_chat_integration.py
```diff
@@ -0,0 +1,238 @@
...
+        assert "total" in data["data"]
+
+    @pytest.mark.asyncio
+    async def test_get_sessions_requires_auth(self, client):
+        """Unauthenticated GET /api/v1/ai/sessions returns 401/403."""
+        response = await client.get("/api/v1/ai/sessions")
+        assert response.status_code in (401, 403)
+
+    @pytest.mark.asyncio
+    async def test_send_message_and_get(self, api_client):
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** test_send_message_and_get verifies the placeholder assistant response content ('AI integration is coming soon') — this encodes implementation detail of the router and will break when the placeholder is replaced with a real LLM call (rule 70: regression tests for past bugs).

**Suggestion:** Only assert that assistant_message exists and has role='assistant'; remove the assertion on the specific placeholder text.

tests/unit/test_ai_router.py
```diff
@@ -0,0 +1,200 @@
...
+from api.routers.ai import ai_router
+from db.connection import get_db
+from internal.middleware.fastapi_auth import AuthContext
+
+
+def _make_auth_ctx(tenant_id: int = 1, user_id: int = 99) -> AuthContext:
+    return AuthContext(user_id=user_id, tenant_id=tenant_id, roles=[])
+
+
+class MockAIService:
```
yihua-zhuo: **INFO** (suggestion)

**Issue:** MockAIService class is defined but never imported or used by any test — dead code.

**Suggestion:** Remove the MockAIService class, or use it in a test that validates service-layer behavior is correctly wired through the router (e.g. a happy-path test with dependency override).

tests/unit/test_ai_router.py
```diff
@@ -0,0 +1,200 @@
...
+
+    async def save_message(self, session_id, role, content, tenant_id):
+        msg = MagicMock(
+            id=10, session_id=session_id, tenant_id=tenant_id,
+            role=role, content=content, created_at=None,
+        )
+        return msg
+
+
+class TestAIEndpoints:
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** All TestAIEndpoints tests only verify that unauthenticated requests return 401/403 — there are no happy-path tests that actually exercise the endpoint logic with valid auth. This leaves wiring bugs undetected.

**Suggestion:** Add at least one happy-path test per endpoint (e.g. POST /sessions returns 201 with valid auth) using app dependency_overrides to inject a mock AuthContext.

tests/unit/test_ai_router.py
```diff
@@ -0,0 +1,200 @@
...
+    def test_get_messages_requires_auth(self, client):
+        response = client.get("/api/v1/ai/sessions/1/messages")
+        assert response.status_code in (401, 403)
+
+    def test_send_message_requires_auth(self, client):
+        response = client.post("/api/v1/ai/sessions/1/messages", json={"content": "Hello"})
+        assert response.status_code in (401, 403)
+
+
+class TestCreateSession:
```
yihua-zhuo: **INFO** (suggestion)

**Issue:** TestCreateSession is a single-test class with only one test that checks auth rejection — equivalent to TestAIEndpoints.test_create_session_requires_auth, making TestCreateSession effectively redundant.

**Suggestion:** Remove TestCreateSession or repurpose it as the location for a happy-path create_session test with valid auth.

tests/unit/test_ai_router.py
```diff
@@ -0,0 +1,200 @@
+"""Unit tests for src/api/routers/ai.py — AI router endpoint tests."""
+
+from unittest.mock import MagicMock
```
yihua-zhuo: **INFO** (suggestion)

**Issue:** AsyncMock is imported but not used — dead import.

**Suggestion:** Remove `from unittest.mock import AsyncMock, MagicMock` and keep only `MagicMock`.

tests/unit/test_ai_service.py
```diff
@@ -0,0 +1,214 @@
...
+            obj.id = 10
+            obj.session_id = 1
+            obj.tenant_id = 1
+            obj.role = "user"
+            obj.content = "Hello"
+            obj.created_at = None
+
+        mock_session.refresh = fake_refresh
+
+        mock_session_obj = MagicMock(id=1, tenant_id=1, title=None, updated_at=None)
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** mix of `MagicMock` and `make_mock_session([])` for the mock session — make_mock_session provides a pre-wired mock but the test then patches execute and refresh on it, making the setup harder to follow. A consistent pattern would be easier to maintain.

**Suggestion:** Use `MagicMock(spec=AsyncSession)` or a dedicated test helper that wires `add`, `flush`, `refresh` as AsyncMock from the start, so patching is replaced with initialization.

frontend/package-lock.json
```diff
@@ -21,7 +21,10 @@
         "react": "19.2.4",
         "react-dom": "19.2.4",
         "react-hook-form": "^7.75.0",
+        "react-markdown": "^10.1.0",
+        "react-syntax-highlighter": "^16.1.1",
```
yihua-zhuo: **CRITICAL** (critical)

**Issue:** Lock file at ^16.1.1 conflicts with package.json's ^15.0.0 constraint — package.json is the source of truth and must be updated (rule 71).

**Suggestion:** Update package.json react-syntax-highlighter to ^16.1.1. This file requires no manual edit; it will self-correct after npm install.

frontend/src/lib/components/ai-panel.tsx
```diff
@@ -1,89 +1,123 @@
...
   return (
     <>
       {/* FAB trigger */}
       <button
         type="button"
-        onClick={() => setOpen(true)}
         className="fixed bottom-6 right-6 z-40 flex h-12 w-12 items-center justify-center rounded-full bg-primary text-primary-foreground shadow-lg hover:bg-primary/90 transition-colors"
         aria-label="Open AI assistant"
+        onClick={() => {
+          const panel = document.getElementById("ai-panel-root");
```
yihua-zhuo: **CRITICAL** (critical)

**Issue:** Direct DOM manipulation (classList.toggle) for panel visibility bypasses React state, making the open/close state invisible to React's reconciler and fragile under re-renders.

**Suggestion:** Lift open/close state into a useState boolean (e.g., `const [isOpen, setIsOpen] = useState(false)`) and render conditionally or apply the CSS class via template literals. Remove the FAB_CLASS constant and DOM query.

frontend/src/lib/components/chat-message.tsx
```diff
@@ -0,0 +1,142 @@
...
+                          className="text-xs !my-0 rounded-lg !bg-muted"
+                        >
+                          {code}
+                        </SyntaxHighlighter>
+                      </div>
+                    );
+                  },
+                  a({ href, children }) {
+                    return (
+                      <a href={href} target="_blank" rel="noopener noreferrer" className="text-primary underline">
```
yihua-zhuo: **CRITICAL** (critical)

**Issue:** href from AI-generated markdown content is rendered in an <a> tag without validating against javascript: or other dangerous URL schemes — XSS risk if the backend serves unescaped user content as markdown (Rule 52).

**Suggestion:** Add URL scheme validation in the a component: `const isSafe = href && /^https?:\/\//i.test(href); return isSafe ? <a href={href} ...> : <span>{children}</span>;`

frontend/src/lib/components/chat-session-sidebar.tsx
```diff
@@ -0,0 +1,150 @@
+"use client";
+import { useState } from "react";
+import { useRouter } from "next/navigation";
+import { Plus, Trash2, MessageSquare, X } from "lucide-react";
+import { cn } from "@/lib/utils";
+import { useChatSessions, useDeleteChatSession, useCreateChatSession } from "@/lib/api/queries";
+import { Button } from "@/components/ui/button";
+
+function formatRelativeTime(dateStr: string): string {
+  const date = new Date(dateStr);
```
yihua-zhuo: **CRITICAL** (critical)

**Issue:** new Date(dateStr) is called without validating the input. If dateStr is an empty string or unparseable, getTime() returns NaN, causing formatRelativeTime to return 'Invalid Date' or NaN strings displayed to users (Rule 50).

**Suggestion:** Guard at the top of formatRelativeTime: `const date = new Date(dateStr); if (Number.isNaN(date.getTime())) return 'Unknown';`

frontend/src/lib/store/chat-store.ts
```diff
@@ -0,0 +1,58 @@
+import { create } from "zustand";
+
+export interface Message {
+  id: string;
+  role: "user" | "assistant";
+  content: string;
+  timestamp: Date;
```
yihua-zhuo: **CRITICAL** (critical)

**Issue:** `Message.timestamp: Date` cannot round-trip through JSON. If Zustand's persist middleware (or any serialization) is ever added, it will silently break.

**Suggestion:** Store timestamps as ISO strings (`string`) or provide custom persist middleware serialization for `Date`.

src/api/__init__.py
```diff
@@ -1,6 +1,7 @@
 """Central API router — aggregates all feature routers."""
 
 from api.routers.activities import activities_router
+from api.routers.ai import ai_router
 from api.routers.automation import automation_router
```
yihua-zhuo, line 4: **CRITICAL** (critical)

**Issue:** `automation_router` is imported (line 5) but absent from `__all__` (line 15-27), making the public API list inaccurate andtooling that relies on `__all__` will miss it.

**Suggestion:** Add `"automation_router",` to the `__all__` list.

frontend/src/app/(app)/ai/[sessionId]/page.tsx
```diff
@@ -0,0 +1,150 @@
...
+    addMessage(sessionIdNum, { id: tempId, role: "user", content: text, timestamp: new Date() });
+    setLoading(true);
+    try {
+      await sendMessageMutation.mutateAsync({
+        sessionId: sessionIdNum,
+        content: text,
+        options,
+      });
+    } catch (err) {
+      console.error("[AISessionPage] sendMessage failed:", err);
```
Comment (yihua-zhuo) [inline frontend/src/app/(app)/ai/[sessionId]/page.tsx]: **INFO** (suggestion)

**Issue:** Error messages are logged to console but not surfaced to the user via UI feedback.

**Suggestion:** Add a toast notification or inline error display when sendMessage fails, not just a console.error.

frontend/src/app/(app)/ai/page.tsx
```diff
@@ -1,10 +1,43 @@
...
+    hasRunRef.current = true;
+    createSession.mutate({}, {
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** createSession.mutate is called with {} as first arg, but the mutationFn in queries.ts (line 487) takes zero arguments — passing {} is misleading and masks a type mismatch.

**Suggestion:** Pass no argument: createSession.mutate() — since mutationFn takes no parameters.

frontend/src/app/(app)/ai/page.tsx
```diff
@@ -1,10 +1,43 @@
...
+import { useRouter } from "next/navigation";
+import { useCreateChatSession } from "@/lib/api/queries";
+import { useChatStore } from "@/lib/store/chat-store";
+
 export default function AIPage() {
+  const router = useRouter();
+  const { activeSessionId, setActiveSession } = useChatStore();
+  const createSession = useCreateChatSession();
+  const hasRunRef = useRef(false);
+
```
yihua-zhuo: **INFO** (suggestion)

**Issue:** hasRunRef.current is guarded in the effect body but not listed in the useEffect deps array — ESLint exhaustive-deps would flag this.

**Suggestion:** Add hasRunRef to the deps array: [activeSessionId, createSession, router, setActiveSession, hasRunRef].

frontend/src/app/(app)/ai/page.tsx
```diff
@@ -1,10 +1,43 @@
...
+    }
+    if (createSession.isPending || hasRunRef.current) return;
+    hasRunRef.current = true;
+    createSession.mutate({}, {
+      onSuccess: (result) => {
+        try {
+          const newId = result.data.id;
+          setActiveSession(newId);
+          router.replace(`/ai/sessions/${newId}`);
+        } catch (err) {
```
yihua-zhuo: **INFO** (suggestion)

**Issue:** The catch block swallows err silently — the user sees no feedback if router.replace fails.

**Suggestion:** Log a console.error AND show a user-facing message (e.g., a toast) so the user knows navigation failed.

frontend/src/lib/api/ai.ts
```diff
@@ -0,0 +1,83 @@
...
+export interface ChatMessage {
+  id: number;
+  session_id: number;
+  tenant_id: number;
+  role: "user" | "assistant";
+  content: string;
+  created_at: string;
+}
+
+interface PaginatedSessions {
```
yihua-zhuo: **INFO** (suggestion)

**Issue:** PaginatedSessions and MessageList interfaces are defined locally and not reused — queries.ts also defines its own paginated envelope.

**Suggestion:** Share these interfaces from ai.ts or a central types file so they stay in sync.

frontend/src/lib/api/queries.ts
```diff
@@ -16,6 +17,8 @@ export const qk = {
   reminders: (upcomingOnly = false) => ["reminders", upcomingOnly] as const,
   activities: (page = 1, type = "") => ["activities", page, type] as const,
   slaBreaches: () => ["sla", "breaches"] as const,
+  aiSessions: (page = 1) => ["ai", "sessions", page] as const,
+  aiMessages: (sessionId: number) => ["ai", "messages", sessionId] as const,
 } as const;
```
yihua-zhuo, line 23: **INFO** (suggestion)

**Issue:** qk.aiSessions and qk.aiMessages exist here but the actual API functions and types for AI live in ai.ts — query key constants and API logic are split across two files.

**Suggestion:** Co-locate AI query keys (qk.aiSessions, qk.aiMessages) with the AI API functions in ai.ts, or move ai.ts exports into queries.ts to keep the AI layer together.

frontend/src/lib/components/ai-panel.tsx
```diff
@@ -1,89 +1,123 @@
...
-      addMessage(
-        "assistant",
-        `Based on your CRM data: "${msg}" — I'm still learning your data patterns. A full AI integration is coming soon.`
-      );
+  // Scroll to bottom on new messages
+  useEffect(() => {
+    bottomRef.current?.scrollIntoView({ behavior: "smooth" });
+  }, [messages.length]);
+
+  async function handleSend(text: string) {
```
yihua-zhuo: **INFO** (suggestion)

**Issue:** handleSend is a plain function, not wrapped in useCallback, so it recreates on every render and breaks the memoization guarantee of child components.

**Suggestion:** Wrap with useCallback: `const handleSend = useCallback(async (text: string) => { ... }, [activeSessionId, isLoading, sendMessageMutation])`

frontend/src/lib/components/ai-panel.tsx
```diff
@@ -1,89 +1,123 @@
...
-  // Cleanup timeout on unmount or when open closes
-  useEffect(() => {
-    return () => {
-      if (aiTimeoutRef.current) clearTimeout(aiTimeoutRef.current);
-    };
-  }, []);
-
-  function handleSuggestion(text: string) {
-    handleSend(text);
+  function handleClear() {
```
yihua-zhuo: **INFO** (suggestion)

**Issue:** handleClear is a plain function recreated on every render.

**Suggestion:** Wrap with useCallback: `const handleClear = useCallback(() => { ... }, [activeSessionId, clearSession])`

frontend/src/lib/components/chat-input.tsx
```diff
@@ -0,0 +1,92 @@
+"use client";
+import { useState, KeyboardEvent } from "react";
+import { Send, Trash2 } from "lucide-react";
+import { cn } from "@/lib/utils";
+import { Button } from "@/components/ui/button";
+import { Textarea } from "@/components/ui/textarea";
+
+const SUGGESTIONS = [
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** SUGGESTIONS is a module-level mutable array (Rule 28). If the module is ever imported server-side or in a non-component context, mutations could leak across invocations.

**Suggestion:** Move SUGGESTIONS inside the component as a const, or declare as `const SUGGESTIONS: readonly string[] = [...]` to prevent accidental mutation.

frontend/src/lib/components/chat-input.tsx
```diff
@@ -0,0 +1,92 @@
...
+  onSend: (text: string) => void;
+  onClear?: () => void;
+  disabled?: boolean;
+  placeholder?: string;
+}
+
+export function ChatInput({ onSend, onClear, disabled, placeholder = "Ask about your CRM…" }: ChatInputProps) {
+  const [input, setInput] = useState("");
+
+  function handleKeyDown(e: KeyboardEvent<HTMLTextAreaElement>) {
```
yihua-zhuo: **INFO** (suggestion)

**Issue:** handleKeyDown, handleSend, and handleSuggestion are plain functions recreated on every render — inefficient when ChatInput re-renders from parent changes.

**Suggestion:** Wrap in useCallback hooks to preserve function identity across renders and avoid downstream effects firing unnecessarily.

frontend/src/lib/components/chat-message.tsx
```diff
@@ -0,0 +1,142 @@
...
+      {message.role === "assistant" && (
+        <div className="mt-0.5 flex h-6 w-6 flex-shrink-0 items-center justify-center rounded-full bg-primary/10">
+          <Bot className="h-3.5 w-3.5 text-primary" />
+        </div>
+      )}
+      <div className="max-w-[80%] rounded-xl px-3 py-2 text-sm">
+        {message.role === "assistant" ? (
+          <div className="space-y-1">
+            <div className="space-y-2">
+              <ReactMarkdown
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** ReactMarkdown renders markdown directly from AI responses. If the backend stores content with raw HTML (e.g., <script> or ), react-markdown will emit it as-is — XSS risk (Rule 52).

**Suggestion:** Add an HTML sanitization step (e.g., DOMPurify) on message.content before passing it to ReactMarkdown, or configure remarkPlugins to strip raw HTML.

frontend/src/lib/components/chat-message.tsx
```diff
@@ -0,0 +1,142 @@
...
+                          }}
+                          className="absolute right-2 top-2 rounded p-1 opacity-0 transition-opacity bg-muted hover:bg-muted/80 cursor-pointer group-hover:opacity-100 group-focus-within:opacity-100 focus-visible:opacity-100 focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-primary/50"
+                          aria-label="Copy code"
+                        >
+                          <Copy className="h-3 w-3" />
+                        </button>
+                        <SyntaxHighlighter
+                          style={oneDark}
+                          language={match[1]}
+                          PreTag="div"
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** oneDark is a large bundled style imported directly from react-syntax-highlighter/dist/esm — not tree-shakeable, adds significant bundle size for a cosmetic feature (Rule 29).

**Suggestion:** Import just the required style object or define a minimal custom theme inline: `const myStyle = { 'pre[class*="language-"]': { background: 'transparent', margin: 0 } }`

frontend/src/lib/components/chat-session-sidebar.tsx
```diff
@@ -0,0 +1,150 @@
...
+export function ChatSessionSidebar({ activeSessionId, onClose }: ChatSessionSidebarProps) {
+  const [deletingId, setDeletingId] = useState<number | null>(null);
+  const router = useRouter();
+  const { data, isLoading } = useChatSessions();
+  const deleteMutation = useDeleteChatSession();
+  const createMutation = useCreateChatSession();
+
+  const sessions = data?.data?.items ?? [];
+
+  async function handleNewConversation() {
```
yihua-zhuo: **INFO** (suggestion)

**Issue:** handleNewConversation is a plain async function, not useCallback — recreates on every render, breaking memoization for any child depending on it.

**Suggestion:** Wrap with useCallback: `const handleNewConversation = useCallback(async () => { ... }, [createMutation, router, onClose])`

frontend/src/lib/components/chat-session-sidebar.tsx
```diff
@@ -0,0 +1,150 @@
...
+      const result = await createMutation.mutateAsync();
+      const newSession = result.data;
+      router.push(`/ai/sessions/${newSession.id}`);
+      onClose?.();
+    } catch (err) {
+      console.error("[ChatSessionSidebar] createSession failed:", err);
+    }
+  }
+
+  async function handleDelete(e: React.MouseEvent, sessionId: number) {
```
yihua-zhuo: **INFO** (suggestion)

**Issue:** handleDelete is a plain async function, not useCallback — same issue.

**Suggestion:** Wrap with useCallback: `const handleDelete = useCallback(async (e, sessionId) => { ... }, [deleteMutation, activeSessionId, router])`

frontend/src/lib/components/chat-session-sidebar.tsx
```diff
@@ -0,0 +1,150 @@
...
+  if (diffHr < 24) return `${diffHr}h ago`;
+  const diffDay = Math.floor(diffHr / 24);
+  if (diffDay < 7) return `${diffDay}d ago`;
+  return date.toLocaleDateString();
+}
+
+interface ChatSessionSidebarProps {
+  activeSessionId?: number | null;
+  onClose?: () => void;
+}
```
yihua-zhuo: **INFO** (suggestion)

**Issue:** ChatStore Message.id is string while activeSessionId is number — type inconsistency across the store boundary; not a runtime bug but indicates incomplete typing.

**Suggestion:** Align Message.id in chat-store.ts with the actual backend type (number) or add an explicit comment documenting the string-coercion contract used in AIPanel.

frontend/src/lib/components/typing-indicator.tsx
```diff
@@ -0,0 +1,22 @@
...
+  className?: string;
+}
+
+export function TypingIndicator({ className }: TypingIndicatorProps) {
+  return (
+    <div className={cn("flex items-center gap-1", className)}>
+      {DOT_DELAYS.map((delay) => (
+        <div
+          key={delay}
+          className="h-1.5 w-1.5 rounded-full bg-muted-foreground animate-bounce"
```
yihua-zhuo: **INFO** (suggestion)

**Issue:** `animate-bounce` performs a vertical bounce animation, not the horizontal oscillation expected from a typing indicator; `animate-pulse` or a custom horizontal keyframe is more semantically correct.

**Suggestion:** Replace `animate-bounce` with `animate-pulse` or define a custom `@keyframes` that oscillates horizontally.

frontend/src/lib/store/chat-store.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  setLoading: (loading) => set({ isLoading: loading }),
+
+  clearSession: (sessionId) =>
+    set((state) => ({
+      messages: {
+        ...state.messages,
+        [sessionId]: [],
+      },
+      ...(state.activeSessionId === sessionId ? { activeSessionId: null, isLoading: false } : {}),
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** `isLoading` has no error-boundary reset — if an API call sets it to `true` and then throws, `isLoading` stays `true` permanently.

**Suggestion:** Reset `isLoading: false` in every error path of calling code, or add a `resetLoading()` action.

src/api/routers/ai.py
```diff
@@ -0,0 +1,194 @@
+"""AI Chat router — session management, messages, chat completion.
+
+Services raise AppException on errors (caught by global handler in main.py).
+Models have .to_dict(); router calls it before returning.
+"""
+
+import json
+
+from fastapi import APIRouter, Depends, HTTPException, Path, Query
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** `Path` is imported but never used in this file.

**Suggestion:** Remove `Path` from the import statement.

src/api/routers/ai.py
```diff
@@ -0,0 +1,194 @@
...
+
+
+# ---------------------------------------------------------------------------
+# Request schemas
+# ---------------------------------------------------------------------------
+
+
+class SendMessageRequest(BaseModel):
+    content: str = Field(..., min_length=1, max_length=MAX_MESSAGE_LENGTH)
+    regenerate: bool = Field(default=False)
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** `regenerate: bool` is defined in `SendMessageRequest` and accepted from the client but completely ignored by the `send_message` handler.

**Suggestion:** Either use `regenerate` to branch the logic or remove it from the request model until it is implemented.

src/api/routers/ai.py
```diff
@@ -0,0 +1,194 @@
...
+        "tenant_id": session.tenant_id,
+        "user_id": session.user_id,
+        "title": session.title,
+        "created_at": session.created_at.isoformat() if session.created_at else None,
+        "updated_at": session.updated_at.isoformat() if session.updated_at else None,
+    }
+
+
+def _message_to_dict(msg) -> dict:
+    return {
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** `tenant_id` is included in the serialized `_message_to_dict` output. The client already knows its own tenant; exposing it in the API response is unnecessary data leakage.

**Suggestion:** Remove `"tenant_id": msg.tenant_id` from `_message_to_dict`.

src/api/routers/ai.py
```diff
@@ -0,0 +1,194 @@
...
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+
+def _session_to_dict(session) -> dict:
+    return {
+        "id": session.id,
+        "tenant_id": session.tenant_id,
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** `tenant_id` is included in the serialized `_session_to_dict` output, serving no client-side purpose.

**Suggestion:** Remove `"tenant_id": session.tenant_id` from `_session_to_dict`.

src/api/routers/ai.py
```diff
@@ -0,0 +1,194 @@
...
+
+@ai_router.post("/sessions/{session_id}/messages", status_code=201)
+async def send_message(
+    session_id: int,
+    body: SendMessageRequest,
+    ctx: AuthContext = Depends(require_auth),
+    session: AsyncSession = Depends(get_db),
+):
+    """Send a user message, save it, and return a placeholder AI response.
+
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** Docstring says 'Replace the body of this handler with a real LLM call (e.g. via AIService) when available' but the actual implementation saves and returns a hardcoded string — the docstring is stale and misleading.

**Suggestion:** Update the docstring to accurately describe the current placeholder behavior.

src/services/ai_service.py
```diff
@@ -0,0 +1,154 @@
...
+            raise ValidationException("role must be 'user' or 'assistant'")
+        await self.get_session(session_id, tenant_id)  # validate ownership
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** save_message calls get_session() for ownership validation, then immediately re-queries the session in the same scope, making the first call pure overhead.

**Suggestion:** Reuse the session object fetched by get_session() instead of running a second identical query. One option: refactor get_session() to return the session object so callers can use it directly.

tests/integration/test_ai_chat_integration.py
```diff
@@ -0,0 +1,238 @@
+"""Integration tests for AI Chat endpoints.
+
+Uses db_schema, tenant_id, async_session fixtures from conftest.py.
+Tests: session create → message send → message list → session delete.
+Rule: integration test services must not call commit() — get_db already commits.
```
yihua-zhuo: **INFO** (suggestion)

**Issue:** Docstring comment 'Rule: integration test services must not call commit()' is misleading — this is guidance, not a runtime rule, and the file's tests don't call commit directly anyway.

**Suggestion:** Remove or rephrase the comment to describe what the tests actually cover: 'Tests session create → message send → message list → session delete using real DB via async_session fixture.'

tests/unit/test_ai_router.py
```diff
@@ -0,0 +1,139 @@
...
+from fastapi.testclient import TestClient
+
+from api.routers.ai import ai_router
+from db.connection import get_db
+
+
+class TestAIEndpoints:
+    @pytest.fixture
+    def mock_db_session(self):
+        async def _session():
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** `mock_db_session` returns a synchronous generator function (`def _session(): yield MagicMock()`), but `get_db` is an async generator — the override will produce a bare `MagicMock` at yield time rather than an `AsyncIterator`.

**Suggestion:** Make the fixture return an async generator: `async def _session(): yield MagicMock()` and mark `mock_db_session` with `@pytest.fixture` + `@pytest.mark.asyncio`, or use `@pytest_asyncio.fixture`.

tests/unit/test_ai_service.py
```diff
@@ -0,0 +1,208 @@
...
+    """Build a mock AsyncSession whose execute() returns *result*."""
+    session = MagicMock()
+    session.execute = AsyncMock(return_value=result)
+    session.add = MagicMock()
+    session.flush = AsyncMock()
+    session.refresh = AsyncMock()
+    return session
+
+
+@pytest.fixture
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** `mock_db_session` passes an empty handler list `[]` to `make_mock_session`, so every SQL execution returns `MockResult([])`. The `TestListSessions` test passes 'for the wrong reason' — it doesn't exercise the service's count+select logic, it just benefits from the empty fallback.

**Suggestion:** Add a stateless AI-session handler to the mock so `list_sessions` exercises real code paths: `from tests.unit.conftest import make_mock_session, MockResult, MockRow` and define a minimal `ai_session_handler` that returns expected rows for the AI service queries.

yihua-zhuo: **INFO** (suggestion)

**Issue:** No AI-specific SQL handler exists in conftest.py; the empty handler list means every query to the AI service hits the fallback `MockResult([])`, so `TestSaveAndGetMessages`, `TestDeleteSession`, and `TestUpdateSessionTitle` all pass incidentally without exercising their real SQL paths.

**Suggestion:** Add a stateless `ai_session_handler` to `tests/unit/conftest.py` (pattern matching `from ai_sessions` / `from chat_messages`) so these tests verify actual query behaviour rather than just 'session method called X on mock'.

PR #372, opened 2026-05-13
Auto: implement issue #63

### LINKED ISSUE #63: [UI] Task Board — Kanban board for user tasks and to-dos
## Problem

task_service.py has task APIs but no UI. Users manage tasks via API only.

## Dependencies

### Task Board (`/tasks`)
- [ ] Kanban columns: To Do / In Progress / Done
- [ ] Drag-and-drop between columns
- [ ] Task cards: title, due date, priority badge, assignee avatar
- [ ] Column task counts
- [ ] "Add Task" button in each column

### Task List View (`/tasks/list`)
- [ ] Table view with all tasks
- [ ] Filters: status, assignee, due date, priority
- [ ] Bulk complete / delete
- [ ] Sort by due date, priority, created date

### Task Detail
- [ ] Title, description (markdown)
- [ ] Status selector (dropdown)
- [ ] Due date picker
- [ ] Priority selector
- [ ] Assignee selector
- [ ] Related customer / opportunity
- [ ] Comments / notes section
- [ ] Edit / Delete / Complete buttons

### Quick Add
- [ ] Global "Add Task" in header
- [ ] Inline create in Kanban column

yihua-zhuo:
Closes #63

Plan: `.plans/issue-63.md`

1. **Plan** — Claude wrote a detailed implementation plan to `.plans/issue-63.md`.

Please review before merging.

## Summary by CodeRabbit

* **New Features**
  * Kanban drag-and-drop for tasks and tickets; global quick-add task button; task modal with assignee/customer/opportunity selectors and markdown preview; ticket detail page with replies and edit/delete flows; SLA timers and Kanban/table views for tickets.

* **Improvements**
  * Task list: multi-select checkboxes, header select-all, bulk Complete/Delete with confirmation, assignee/date/status/priority filters, loading/error/empty states.

* **API**
  * Public Tasks REST endpoints (CRUD, complete, list with filters & pagination).

* **Tests**
  * New unit + integration suites covering tasks, tickets, and auth flows.

* **Chores**
  * Formatting, linting, test tooling and dependency updates; docs and implementation plans added.

[Review Change Stack](https://app.coderabbit.ai/change-stack/yihua-zhuo/agent-job/pull/372)

### DIFF
--- a/.plans/issue-63.md
+++ b/.plans/issue-63.md
@@ -1,5 +1,3 @@
-Now I have a thorough picture of the codebase. Let me compile the implementation plan.
-
 # Implementation Plan — Issue #63
 
 ## Goal
--- a/.plans/issue-63.md
+++ b/.plans/issue-63.md
@@ -1,0 +1,80 @@
+Now I have a thorough picture of the codebase. Let me compile the implementation plan.
+
+# Implementation Plan — Issue #63
+
+## Goal
+
+The frontend already has a Kanban board, list view, and task modal for `/tasks` and `/tasks/list`, backed by React Query hooks. The **backend has no tasks router** — `TaskService` and `TaskModel` exist but no `/api/v1/tasks` endpoints are wired up. This issue completes the backend API and enhances the UI with drag-and-drop, assignee picker, markdown preview, quick-add header, and bulk operations.
+
+## Affected Files
+
+- `src/api/routers/tasks.py` — **New** — FastAPI router for all task CRUD endpoints
+- `src/api/__init__.py` — **Modify** — add `tasks_router` export
+- `src/main.py` — **Modify** — register `tasks_router`
+- `frontend/src/app/(app)/tasks/page.tsx` — **Modify** — add drag-and-drop, column task counts from correct total
+- `frontend/src/app/(app)/tasks/list/page.tsx` — **Modify** — add bulk complete/delete, assignee filter, due date range sort
+- `frontend/src/app/(app)/tasks/task-modal.tsx` — **Modify** — add markdown preview, assignee selector, customer/opportunity link selector
+- `frontend/src/components/layout/app-sidebar.tsx` — **Modify** — add global quick-add task button in header
+- `tests/unit/test_tasks.py` — **New** — unit tests for task service + router
+- `tests/integration/test_tasks_integration.py` — **New** — integration tests for task endpoints
+
+## Implementation Steps
+
+### Backend
+1. Create `src/api/routers/tasks.py` with these endpoints, matching the patterns in `tickets.py` and the hook calls in `frontend/src/lib/api/queries.ts`:
+   - `GET /api/v1/tasks` — list with `page`, `page_size`, `status`, `assigned_to` query params; returns `{success, data: {items, total, page, page_size, has_next}}`
+   - `POST /api/v1/tasks` — create; body fields `title`, `description`, `priority`, `status`, `due_date`, `assigned_to`
+   - `GET /api/v1/tasks/{task_id}` — single task
+   - `PATCH /api/v1/tasks/{task_id}` — update fields
+   - `DELETE /api/v1/tasks/{task_id}` — delete
+   - `POST /api/v1/tasks/{task_id}/complete` — sets `status=completed`
+   - All endpoints use `ctx: AuthContext = Depends(require_auth)` and `session: AsyncSession = Depends(get_db)`
+2. Export `tasks_router` from `src/api/__init__.py`
+3. Register `tasks_router` in `src/main.py` via `app.include_router(tasks_router)`
+
+### Frontend — Kanban Enhancements
+4. In `frontend/src/app/(app)/tasks/page.tsx`:
+   - Add `@dnd-kit/core` + `@dnd-kit/sortable` for drag-and-drop between columns (install if not in `package.json`)
+   - Make task cards draggable with `useSortable`, columns are droppable zones keyed by `col.id`
+   - On drop, call `useUpdateTask().mutateAsync({ id, data: { status: targetColumnId } })`
+   - Display accurate column counts by computing `allTasks.filter(t => t.status === col.id).length` from the full dataset; remove the redundant per-status API calls and use only `useTasks(1, "")`
+
+### Frontend — TaskModal Enhancements
+5. In `frontend/src/app/(app)/tasks/task-modal.tsx`:
+   - Add `react-markdown` or `@uiw/react-md-editor` for description preview/edit toggle
+   - Add assignee `Select` dropdown using `useUsers()` hook to populate user list
+   - Add optional customer and opportunity link selectors using `useSearchCustomers()` and `useOpportunities()` (or `usePipelines()`) — these store `customer_id` and `opportunity_id` in the task payload but the backend TaskModel stores them via extra fields or a relation
+
+### Frontend — List View Enhancements
+6. In `frontend/src/app/(app)/tasks/list/page.tsx`:
+   - Add checkbox column for bulk selection
+   - Add bulk "Complete" and "Delete" buttons that call `useCompleteTask()` / `useDeleteTask()` on the selected array
+   - Add assignee filter dropdown (uses `useUsers()` to populate options)
+   - Add created-by date range filter (two date inputs: `created_after`, `created_before`)
+
+### Frontend — Global Quick-Add
+7. In `frontend/src/components/layout/app-sidebar.tsx` (or the app layout header):
+   - Add a `+` icon button in the top nav bar next to the page title
+   - On click, open `TaskModal` with `initialStatus="pending"` pre-selected
+   - Pass `onClose` to reset the button state
+
+## Test Plan
+
+- **Unit tests in `tests/unit/test_tasks.py`**: Create a `mock_db_session` fixture using `make_mock_session` with a `task_handler` (add this handler to `conftest.py` following the pattern of `customer_handler`). Test `TaskService` methods: `create_task`, `get_task`, `update_task`, `complete_task`, `delete_task`, `list_tasks`. Also test the router functions directly by constructing a mock `App` with `tasks_router` mounted and calling endpoint handlers with mocked session/auth.
+
+- **Integration tests in `tests/integration/test_tasks_integration.py`**: Use `db_schema`, `tenant_id`, `async_session` fixtures. Test full round-trip: create task via `POST /api/v1/tasks`, retrieve via `GET /api/v1/tasks/{id}`, list via `GET /api/v1/tasks?status=pending`, update via `PATCH /api/v1/tasks/{id}`, complete via `POST /api/v1/tasks/{id}/complete`, delete via `DELETE /api/v1/tasks/{id}`. Verify response envelope shape `{success, data}` and status codes.
+
+## Acceptance Criteria
+
+- `GET /api/v1/tasks` returns `{success: true, data: {items: [...], total: N, page, page_size, has_next}}`
+- `POST /api/v1/tasks` returns 201 with `{"success": true, "data": <task dict>}`
+- `PATCH /api/v1/tasks/{id}` updates and returns the task
+- `POST /api/v1/tasks/{id}/complete` sets status to `completed`
+- `DELETE /api/v1/tasks/{id}` returns 200
+- Kanban board renders 3 columns (To Do / In Progress / Done) with accurate task counts
+- Drag-and-drop moves a card from one Kanban column to another and persists via API
+- Task detail modal shows title, description with markdown toggle, status, priority, due date, assignee dropdown
+- Task list view has filters for status, priority, assignee, and sort by due date/priority/created date
+- Bulk complete and bulk delete work on the list view with checkbox selection
+- Unit tests for `TaskService` pass with mocked DB
+- Integration tests for all task API endpoints pass against real PostgreSQL
--- a/frontend/src/app/(app)/layout.tsx
+++ b/frontend/src/app/(app)/layout.tsx
@@ -84,8 +84,12 @@ export default function AppLayout({ children }: { children: React.ReactNode }) {
           onClose={quickAdd.close}
           initialStatus={quickAdd.initialStatus}
           onSubmit={async (data) => {
-            await create.mutateAsync({ ...data, status: (data.status as string | undefined) ?? quickAdd.initialStatus });
-            quickAdd.close();
+            try {
+              await create.mutateAsync({ ...data, status: (data.status as string | undefined) ?? quickAdd.initialStatus });
+              quickAdd.close();
+            } catch {
+              // Error surfaced via React Query onError callback
+            }
           }}
           isSubmitting={create.isPending}
         />
--- a/frontend/src/app/(app)/layout.tsx
+++ b/frontend/src/app/(app)/layout.tsx
@@ -26,7 +26,7 @@ export default function AppLayout({ children }: { children: React.ReactNode }) {
 
   /* eslint-disable react-hooks/set-state-in-effect -- intentional: synchronizing sidebar open state with pathname change */
   useEffect(() => {
-    setSidebarOpen(false);
+    if (sidebarOpen) setSidebarOpen(false);
   }, [pathname]);
   /* eslint-enable react-hooks/set-state-in-effect */
 
@@ -84,12 +84,8 @@ export default function AppLayout({ children }: { children: React.ReactNode }) {
           onClose={quickAdd.close}
           initialStatus={quickAdd.initialStatus}
           onSubmit={async (data) => {
-            try {
-              await create.mutateAsync({ ...data, status: (data.status as string | undefined) ?? quickAdd.initialStatus });
-              quickAdd.close();
-            } catch {
-              // Error surfaced via React Query onError callback
-            }
+            await create.mutateAsync({ ...data, status: (data.status as string | undefined) ?? quickAdd.initialStatus });
+            quickAdd.close();
           }}
           isSubmitting={create.isPending}
         />
--- a/frontend/src/app/(app)/layout.tsx
+++ b/frontend/src/app/(app)/layout.tsx
@@ -24,11 +24,9 @@ export default function AppLayout({ children }: { children: React.ReactNode }) {
   const create = useCreateTask();
   const update = useUpdateTask();
 
-  /* eslint-disable react-hooks/set-state-in-effect -- intentional: synchronizing sidebar open state with pathname change */
   useEffect(() => {
     if (sidebarOpen) setSidebarOpen(false);
-  }, [pathname]);
-  /* eslint-enable react-hooks/set-state-in-effect */
+  }, [pathname, sidebarOpen]);
 
   function toggleSidebar() {
     setCollapsed((c) => !c);
--- a/frontend/src/app/(app)/layout.tsx
+++ b/frontend/src/app/(app)/layout.tsx
@@ -84,7 +84,7 @@ export default function AppLayout({ children }: { children: React.ReactNode }) {
           onClose={quickAdd.close}
           initialStatus={quickAdd.initialStatus}
           onSubmit={async (data) => {
-            await create.mutateAsync({ ...data, status: quickAdd.initialStatus });
+            await create.mutateAsync({ ...data, status: (data.status as string | undefined) ?? quickAdd.initialStatus });
             quickAdd.close();
           }}
           isSubmitting={create.isPending}
--- a/frontend/src/app/(app)/layout.tsx
+++ b/frontend/src/app/(app)/layout.tsx
@@ -6,6 +6,9 @@ import { SessionTimeoutGuard } from "@/lib/components/session-timeout-guard";
 import { OfflineBanner } from "@/lib/components/offline-banner";
 import { AIPanel } from "@/lib/components/ai-panel";
 import { AppSidebar } from "@/components/layout/app-sidebar";
+import { TaskModal } from "@/app/(app)/tasks/task-modal";
+import { useQuickAddTask } from "@/lib/store/task-store";
+import { useCreateTask, useUpdateTask } from "@/lib/api/queries";
 import { Menu } from "lucide-react";
 import { cn } from "@/lib/utils";
 
@@ -17,6 +20,9 @@ export default function AppLayout({ children }: { children: React.ReactNode }) {
     try { return localStorage.getItem(STORAGE_KEY) === "true"; } catch { return false; }
   });
   const pathname = usePathname();
+  const quickAdd = useQuickAddTask();
+  const create = useCreateTask();
+  const update = useUpdateTask();
 
   /* eslint-disable react-hooks/set-state-in-effect -- intentional: synchronizing sidebar open state with pathname change */
   useEffect(() => {
@@ -71,6 +77,18 @@ export default function AppLayout({ children }: { children: React.ReactNode }) {
 
         <AIPanel />
         <SessionTimeoutGuard />
+
+        <TaskModal
+          open={quickAdd.isOpen}
+          title="Create Task"
+          onClose={quickAdd.close}
+          initialStatus={quickAdd.initialStatus}
+          onSubmit={async (data) => {
+            await create.mutateAsync({ ...data, status: quickAdd.initialStatus });
+            quickAdd.close();
+          }}
+          isSubmitting={create.isPending}
+        />
       </div>
     </AuthGuard>
   );
--- a/frontend/src/app/(app)/layout.tsx
+++ b/frontend/src/app/(app)/layout.tsx
@@ -82,8 +82,12 @@ export default function AppLayout({ children }: { children: React.ReactNode }) {
           onClose={quickAdd.close}
           initialStatus={quickAdd.initialStatus}
           onSubmit={async (data) => {
-            await create.mutateAsync({ ...data, status: (data.status as string | undefined) ?? quickAdd.initialStatus });
-            quickAdd.close();
+            try {
+              await create.mutateAsync({ ...data, status: (data.status as string | undefined) ?? quickAdd.initialStatus });
+              quickAdd.close();
+            } catch {
+              // error is handled by the mutation hook (e.g. toast)
+            }
           }}
           isSubmitting={create.isPending}
         />
--- a/frontend/src/app/(app)/tasks/list/page.tsx
+++ b/frontend/src/app/(app)/tasks/list/page.tsx
@@ -46,10 +46,11 @@ export default function TaskListPage() {
   const [editTask, setEditTask] = useState<Record<string, unknown> | null>(null);
   const [selectedIds, setSelectedIds] = useState<Set<number>>(new Set());
 
-  const { data, isLoading, isError } = useTasks(page, statusFilter);
-  const { data: usersData } = useUsers(1);
-  const items = (data?.data?.items ?? []) as Record<string, unknown>[];
-  const info = data?.data;
+  const { data: allData, isLoading, isError } = useTasks(1, "");
+  const { data: filteredData } = useTasks(page, statusFilter);
+  const { data: usersData } = useUsers(1, 100);
+  const items = (allData?.data?.items ?? []) as Record<string, unknown>[];
+  const info = filteredData?.data;
   const users = (usersData?.data?.items ?? []) as Record<string, unknown>[];
 
   const complete = useCompleteTask();
--- a/frontend/src/app/(app)/tasks/list/page.tsx
+++ b/frontend/src/app/(app)/tasks/list/page.tsx
@@ -73,7 +73,7 @@ export default function TaskListPage() {
       });
     }
     if (createdBefore) {
-      const beforeMs = new Date(createdBefore).getTime() + 86400000;
+      const beforeMs = new Date(createdBefore + "T23:59:59").getTime();
       result = result.filter((t) => {
         if (!t.created_at) return false;
         return new Date(String(t.created_at)).getTime() <= beforeMs;
@@ -129,17 +129,13 @@ export default function TaskListPage() {
   }
 
   async function bulkComplete() {
-    for (const id of selectedIds) {
-      await complete.mutateAsync(id);
-    }
+    await Promise.all([...selectedIds].map((id) => complete.mutateAsync(id)));
     setSelectedIds(new Set());
   }
 
   async function bulkDelete() {
     if (!window.confirm(`Delete ${selectedIds.size} task(s)?`)) return;
-    for (const id of selectedIds) {
-      await remove.mutateAsync(id);
-    }
+    await Promise.all([...selectedIds].map((id) => remove.mutateAsync(id)));
     setSelectedIds(new Set());
   }
 
--- a/frontend/src/app/(app)/tasks/list/page.tsx
+++ b/frontend/src/app/(app)/tasks/list/page.tsx
@@ -30,10 +30,6 @@ const PRIORITY_COLORS: Record<string, string> = {
 };
 const PRIORITY_ORDER: Record<string, number> = { urgent: 0, high: 1, normal: 2, low: 3 };
 
-function SortIconStatic() {
-  return <span className="text-muted-foreground">↕</span>;
-}
-
 export default function TaskListPage() {
   const [page, setPage] = useState(1);
   const [statusFilter, setStatusFilter] = useState("");
@@ -46,11 +42,9 @@ export default function TaskListPage() {
   const [editTask, setEditTask] = useState<Record<string, unknown> | null>(null);
   const [selectedIds, setSelectedIds] = useState<Set<number>>(new Set());
 
-  const { data: allData, isLoading, isError } = useTasks(1, "");
-  const { data: filteredData } = useTasks(page, statusFilter);
-  const { data: usersData } = useUsers(1, 100);
-  const items = (allData?.data?.items ?? []) as Record<string, unknown>[];
-  const info = filteredData?.data;
+  const { data, isLoading, isError } = useTasks(page, statusFilter, priorityFilter, assigneeFilter, createdAfter, createdBefore);
+  const items = (data?.data?.items ?? []) as Record<string, unknown>[];
+  const info = data?.data;
   const users = (usersData?.data?.items ?? []) as Record<string, unknown>[];
 
   const complete = useCompleteTask();
--- a/frontend/src/app/(app)/tasks/list/page.tsx
+++ b/frontend/src/app/(app)/tasks/list/page.tsx
@@ -1,6 +1,6 @@
 "use client";
 import { useState } from "react";
-import { useTasks, useDeleteTask, useCompleteTask, useUpdateTask } from "@/lib/api/queries";
+import { useTasks, useDeleteTask, useCompleteTask, useUpdateTask, useUsers } from "@/lib/api/queries";
 import { Button } from "@/components/ui/button";
 import { TaskModal } from "../task-modal";
 import { cn } from "@/lib/utils";
@@ -47,8 +47,10 @@ export default function TaskListPage() {
   const [selectedIds, setSelectedIds] = useState<Set<number>>(new Set());
 
   const { data, isLoading, isError } = useTasks(page, statusFilter);
+  const { data: usersData } = useUsers(1);
   const items = (data?.data?.items ?? []) as Record<string, unknown>[];
   const info = data?.data;
+  const users = (usersData?.data?.items ?? []) as Record<string, unknown>[];
 
   const complete = useCompleteTask();
   const remove = useDeleteTask();
@@ -191,6 +193,11 @@ export default function TaskListPage() {
           </SelectTrigger>
           <SelectContent>
             <SelectItem value="__none__">All Assignees</SelectItem>
+            {users.map((u) => (
+              <SelectItem key={String(u.id)} value={String(u.id)}>
+                {String(u.full_name ?? u.username ?? u.email ?? u.id)}
+              </SelectItem>
+            ))}
           </SelectContent>
         </Select>
         <div className="flex items-center gap-1 text-xs text-muted-foreground">
--- a/frontend/src/app/(app)/tasks/list/page.tsx
+++ b/frontend/src/app/(app)/tasks/list/page.tsx
@@ -3,6 +3,7 @@ import { useState } from "react";
 import { useTasks, useDeleteTask, useCompleteTask, useUpdateTask } from "@/lib/api/queries";
 import { Button } from "@/components/ui/button";
 import { TaskModal } from "../task-modal";
+import { cn } from "@/lib/utils";
 import {
   Select,
   SelectContent,
@@ -37,9 +38,13 @@ export default function TaskListPage() {
   const [page, setPage] = useState(1);
   const [statusFilter, setStatusFilter] = useState("");
   const [priorityFilter, setPriorityFilter] = useState("");
+  const [assigneeFilter, setAssigneeFilter] = useState("");
+  const [createdAfter, setCreatedAfter] = useState("");
+  const [createdBefore, setCreatedBefore] = useState("");
   const [sortKey, setSortKey] = useState<SortKey>("created_at");
   const [sortDir, setSortDir] = useState<SortDir>("desc");
   const [editTask, setEditTask] = useState<Record<string, unknown> | null>(null);
+  const [selectedIds, setSelectedIds] = useState<Set<number>>(new Set());
 
   const { data, isLoading, isError } = useTasks(page, statusFilter);
   const items = (data?.data?.items ?? []) as Record<string, unknown>[];
@@ -53,6 +58,24 @@ export default function TaskListPage() {
     let result = items;
     if (priorityFilter) {
       result = result.filter((t) => t.priority === priorityFilter);
+    }
+    if (assigneeFilter) {
+      const aid = Number(assigneeFilter);
+      result = result.filter((t) => Number(t.assigned_to) === aid);
+    }
+    if (createdAfter) {
+      const afterMs = new Date(createdAfter).getTime();
+      result = result.filter((t) => {
+        if (!t.created_at) return false;
+        return new Date(String(t.created_at)).getTime() >= afterMs;
+      });
+    }
+    if (createdBefore) {
+      const beforeMs = new Date(createdBefore).getTime() + 86400000;
+      result = result.filter((t) => {
+        if (!t.created_at) return false;
+        return new Date(String(t.created_at)).getTime() <= beforeMs;
+      });
     }
     return result.slice().sort((a, b) => {
       if (sortKey === "due_date") {
@@ -85,12 +108,56 @@ export default function TaskListPage() {
     }
   }
 
+  function toggleSelect(id: number) {
+    setSelectedIds((prev) => {
+      const next = new Set(prev);
+      if (next.has(id)) next.delete(id);
+      else next.add(id);
+      return next;
+    });
+  }
+
+  function toggleSelectAll() {
+    const displayed = filteredAndSorted();
+    if (selectedIds.size === displayed.length) {
+      setSelectedIds(new Set());
+    } else {
+      setSelectedIds(new Set(displayed.map((t) => Number(t.id))));
+    }
+  }
+
+  async function bulkComplete() {
+    for (const id of selectedIds) {
+      await complete.mutateAsync(id);
+    }
+    setSelectedIds(new Set());
+  }
+
+  async function bulkDelete() {
+    if (!window.confirm(`Delete ${selectedIds.size} task(s)?`)) return;
+    for (const id of selectedIds) {
+      await remove.mutateAsync(id);
+    }
+    setSelectedIds(new Set());
+  }
+
   const displayed = filteredAndSorted();
+  const allSelected = displayed.length > 0 && displayed.every((t) => selectedIds.has(Number(t.id)));
 
   return (
     <div className="space-y-4">
       <div className="flex items-center justify-between">
         <h1 className="text-2xl font-bold">Task List</h1>
+        {selectedIds.size > 0 && (
+          <div className="flex gap-2">
+            <Button size="sm" variant="default" onClick={bulkComplete}>
+              Complete ({selectedIds.size})
+            </Button>
+            <Button size="sm" variant="destructive" onClick={bulkDelete}>
+              Delete ({selectedIds.size})
+            </Button>
+          </div>
+        )}
       </div>
 
       <div className="flex flex-wrap gap-2">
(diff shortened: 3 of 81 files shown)


frontend/src/app/(app)/tasks/page.tsx
```diff
@@ -143,46 +253,112 @@ export default function TasksPage() {
...
-              <span className={`font-semibold text-sm ${col.color}`}>{col.label}</span>
-              <div className="flex items-center gap-2">
-                <span className="text-xs font-medium text-muted-foreground">{countByStatus(col.id)}</span>
-                <Button variant="ghost" size="icon" className="h-6 w-6" onClick={() => { setCreateStatus(col.id); setShowCreate(true); }} aria-label="Add task" title="Add task">
-                  +
-                </Button>
+      <DndContext
+        sensors={sensors}
+        onDragStart={handleDragStart}
+        onDragEnd={handleDragEnd}
```
yihua-zhuo, line 297: **CRITICAL** (critical)

**Issue:** DndContext onDragEnd references handleDragEnd which is never defined, causing an uncaught ReferenceError when a drag ends.

**Suggestion:** Define handleDragEnd: async function handleDragEnd(event: DragEndEvent) { const { active, over } = event; if (!over) return; const taskId = Number(active.id); const newStatus = over.id as string; await update.mutateAsync({ id: taskId, data: { status: newStatus } }); }

.plans/issue-63.md
```diff
@@ -0,0 +1,78 @@
+# Implementation Plan — Issue #63
```
yihua-zhuo, line 1: **INFO** (suggestion)

**Issue:** Plan references `@uiw/react-md-editor` but `package.json` has `react-markdown` instead — plan needs updating to reflect installed dependency

**Suggestion:** Replace `@uiw/react-md-editor` reference with `react-markdown` in step 5, or add `@uiw/react-md-editor` to `frontend/package.json`

yihua-zhuo, line 1: **INFO** (suggestion)

**Issue:** Plan references old `radix-ui` package names (e.g. `Select`) but `package.json` uses `^1.4.3` which is the deprecated monolithic `radix-ui` package — plan references may be outdated

**Suggestion:** Verify `frontend/src/components/ui/select.tsx` exists and uses the correct `@radix-ui/react-select` package, then update the plan if needed

frontend/src/app/(app)/layout.tsx
```diff
@@ -17,6 +20,9 @@ export default function AppLayout({ children }: { children: React.ReactNode }) {
     try { return localStorage.getItem(STORAGE_KEY) === "true"; } catch { return false; }
   });
   const pathname = usePathname();
+  const quickAdd = useQuickAddTask();
+  const create = useCreateTask();
+  const update = useUpdateTask();
 
   /* eslint-disable react-hooks/set-state-in-effect -- intentional: synchronizing sidebar open state with pathname change */
   useEffect(() => {
```
yihua-zhuo, line 27: **WARNING** (suggestion)

**Issue:** The pathname effect runs on every navigation regardless of sidebar open state, causing a superfluous re-render when sidebar is already closed

**Suggestion:** Guard with `if (sidebarOpen) setSidebarOpen(false);` to skip the setState call when sidebar is already closed

frontend/src/app/(app)/layout.tsx
```diff
@@ -71,6 +77,22 @@ export default function AppLayout({ children }: { children: React.ReactNode }) {
...
+        <TaskModal
+          open={quickAdd.isOpen}
+          title="Create Task"
+          onClose={quickAdd.close}
+          initialStatus={quickAdd.initialStatus}
+          onSubmit={async (data) => {
+            try {
+              await create.mutateAsync({ ...data, status: (data.status as string | undefined) ?? quickAdd.initialStatus });
+              quickAdd.close();
+            } catch {
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** Empty `catch {}` block in async `onSubmit` — if React Query's `onError` is truly the only error path, the try/catch is redundant and masks bugs if the assumption changes

**Suggestion:** Remove the try/catch entirely and let exceptions propagate to React Query's error handling; or add explicit error state/logging to avoid silent swallow

frontend/src/app/(app)/tasks/list/page.tsx
```diff
@@ -134,30 +241,46 @@ export default function TaskListPage() {
...
+                  </td>
+                  <td className="px-3 py-2.5">
+                    <Badge colorClass={PRIORITY_COLORS[String(t.priority)] ?? "bg-gray-100 text-gray-600"}>
+                      {String(t.priority ?? "normal")}
+                    </Badge>
+                  </td>
+                  <td className="px-3 py-2.5 text-muted-foreground">
+                    {t.due_date ? new Date(String(t.due_date)).toLocaleDateString() : "—"}
+                  </td>
+                  <td className="px-3 py-2.5 text-muted-foreground">
```
yihua-zhuo, line 274: **INFO** (suggestion)

**Issue:** Rule 27 — TypeScript types for `editTask` use `Record<string, unknown>` losing all type safety; casts like `Number(t.id)` repeated throughout

**Suggestion:** Define a `Task` interface or use the inferred type from the API response and use a typed state

frontend/src/app/(app)/tasks/page.tsx
```diff
@@ -19,19 +34,36 @@ const PRIORITY_COLORS: Record<string, string> = {
...
 }) {
   const due = task.due_date ? new Date(String(task.due_date)) : null;
   const isOverdue = due && due < new Date() && task.status !== "completed";
   const rawId = task.id;
-  const taskId = typeof rawId === "number" && Number.isFinite(rawId)
-    ? rawId
-    : typeof rawId === "string" && /^\d+$/.test(rawId.trim())
-    ? Number(rawId)
-    : null;
+  const taskId =
```
yihua-zhuo, line 61: **WARNING** (suggestion)

**Issue:** tasksByStatus is called 7 times per render without memoization, causing O(n) scans of allTasks per column per render.

**Suggestion:** Memoize with useMemo: const tasksByStatus = useMemo(() => { const m: Record<string, TaskData[]> = {}; for (const col of COLUMNS) m[col.id] = allTasks.filter(t => t.status === col.id); return m; }, [allTasks]);

frontend/src/app/(app)/tasks/page.tsx
```diff
@@ -19,19 +34,36 @@ const PRIORITY_COLORS: Record<string, string> = {
...
+interface TaskData {
+  id: string | number;
+  title: string | null;
+  description?: string | null;
+  status: string;
+  priority?: string | null;
+  due_date?: string | null;
+  [key: string]: unknown;
+}
+
```
yihua-zhuo, line 46: **INFO** (suggestion)

**Issue:** PRIORITY_COLORS and COLUMNS are defined at module scope but could be typed as const for extra safety and tree-shaking.

**Suggestion:** Add 'as const' suffix: const COLUMNS = [{ id: "pending", ... }] as const;

frontend/src/app/(app)/tasks/task-modal.tsx
```diff
@@ -127,15 +185,88 @@ export function TaskModal({
...
+                    <SelectItem key={String(o.id)} value={String(o.id)}>
+                      {String(o.title ?? o.id)}
+                    </SelectItem>
+                  ))}
+                </SelectContent>
+              </Select>
+            </div>
+          </div>
+
           <DialogFooter className="flex-row! justify-between gap-2 sm:justify-between">
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** flex-row! uses invalid Tailwind syntax — the exclamation point modifier is a V4 feature that requires Tailwind V4 to work and is not universally supported.

**Suggestion:** Replace className="flex-row! justify-between" with className="flex flex-row justify-between gap-2 sm:justify-between" or simply flex-row (which is already the default).

frontend/src/app/(app)/tasks/task-modal.tsx
```diff
@@ -47,10 +50,30 @@ export function TaskModal({
...
     status: task ? String(task.status ?? "pending") : initialStatus ?? "pending",
     due_date: task && task.due_date ? String(task.due_date).slice(0, 10) : "",
+    assigned_to: task ? String(task.assigned_to ?? "0") : "0",
+    customer_id: task ? String(task.customer_id ?? "") : "",
+    opportunity_id: task ? String(task.opportunity_id ?? "") : "",
   });
+  const [previewDesc, setPreviewDesc] = useState(false);
+  const [customerSearch, setCustomerSearch] = useState<string | undefined>(undefined);
+
+  const { data: usersData } = useUsers(1);
```
yihua-zhuo: **INFO** (suggestion)

**Issue:** useUsers and useOpportunities are called with page=1 and no search filter, potentially fetching large unpaginated lists that are used only for dropdown population.

**Suggestion:** Consider passing a reasonable page_size limit (e.g., 100) to avoid fetching unnecessary data when only populating a dropdown.

frontend/src/app/(app)/tasks/task-modal.tsx
```diff
@@ -61,6 +84,11 @@ export function TaskModal({
       status: form.status,
     };
     if (form.due_date) payload.due_date = form.due_date;
+    if (form.assigned_to && form.assigned_to !== "0") {
+      payload.assigned_to = Number(form.assigned_to);
+    }
+    if (form.customer_id) payload.customer_id = Number(form.customer_id);
+    if (form.opportunity_id) payload.opportunity_id = Number(form.opportunity_id);
     await onSubmit(payload);
```
yihua-zhuo, line 101: **INFO** (suggestion)

**Issue:** The initialStatus prop is accepted but not used in the initial form state when task is provided — the form always uses task.status when editing, ignoring initialStatus after first mount.

**Suggestion:** Ensure initialStatus is consumed in the useEffect that syncs form state on task prop changes.

src/api/routers/tasks.py
```diff
@@ -0,0 +1,164 @@
...
+    session: AsyncSession = Depends(get_db),
+):
+    service = TaskService(session)
+    task = await service.create_task(
+        title=body.title,
+        description=body.description,
+        priority=body.priority,
+        tenant_id=ctx.tenant_id,
+        created_by=ctx.user_id,
+        assigned_to=body.assigned_to or 0,
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** body.assigned_to or 0 collapses both 0 and None into 0; a tenant_id of 0 silently bypasses multi-tenancy isolation in the SQL WHERE clause (Rule 6)

**Suggestion:** Pass body.assigned_to directly; let the service decide how to handle None vs 0

src/api/routers/tasks.py
```diff
@@ -0,0 +1,164 @@
...
+@tasks_router.get("/tasks")
+async def list_tasks(
+    page: int = Query(1, ge=1),
+    page_size: int = Query(20, ge=1, le=100),
+    status: str | None = Query(None),
+    assigned_to: int | None = Query(None, ge=0),
+    ctx: AuthContext = Depends(require_auth),
+    session: AsyncSession = Depends(get_db),
+):
+    service = TaskService(session)
```
yihua-zhuo, line 105: **INFO** (suggestion)

**Issue:** list_tasks endpoint discards the second return value (total count) from service.list_tasks — count is re-fetched separately at line 111, creating an extra DB round-trip

**Suggestion:** Change service.list_tasks to return (items, total) and eliminate the redundant count_tasks call in the router

tests/integration/test_tasks_integration.py
```diff
@@ -0,0 +1,238 @@
...
+        assert resp.status_code == 200, f"Body: {resp.text}"
+        data = resp.json()
+        assert data["success"] is True
+        assert data["data"]["status"] == "completed"
+
+    async def test_complete_task_not_found(self, api_client: AsyncClient, tenant_id_web: int):
+        resp = await api_client.post("/api/v1/tasks/999999999/complete")
+        assert resp.status_code == 404
+
+    async def test_delete_task(self, api_client: AsyncClient):
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** test_delete_task declares tenant_id_web fixture parameter but never uses it

**Suggestion:** Remove tenant_id_web from the function signature: async def test_delete_task(self, api_client: AsyncClient):

tests/integration/test_tasks_integration.py
```diff
@@ -0,0 +1,238 @@
...
+        resp = await api_client.delete(f"/api/v1/tasks/{created_id}")
+        assert resp.status_code == 200, f"Body: {resp.text}"
+        data = resp.json()
+        assert data["success"] is True
+
+        # Verify it's gone
+        get_resp = await api_client.get(f"/api/v1/tasks/{created_id}")
+        assert get_resp.status_code == 404
+
+    async def test_delete_task_not_found(self, api_client: AsyncClient):
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** test_delete_task_not_found declares tenant_id_web fixture parameter but never uses it

**Suggestion:** Remove tenant_id_web from the function signature: async def test_delete_task_not_found(self, api_client: AsyncClient):

frontend/src/app/(app)/tasks/list/page.tsx
```diff
@@ -37,13 +38,20 @@ export default function TaskListPage() {
...
+  const [createdBefore, setCreatedBefore] = useState("");
   const [sortKey, setSortKey] = useState<SortKey>("created_at");
   const [sortDir, setSortDir] = useState<SortDir>("desc");
   const [editTask, setEditTask] = useState<Record<string, unknown> | null>(null);
+  const [selectedIds, setSelectedIds] = useState<Set<number>>(new Set());
 
-  const { data, isLoading, isError } = useTasks(page, statusFilter);
-  const items = (data?.data?.items ?? []) as Record<string, unknown>[];
-  const info = data?.data;
+  const { data: allData, isLoading, isError } = useTasks(1, "");
```
yihua-zhuo: **CRITICAL** (critical)

**Issue:** Tenant ID is hardcoded to 1 — breaks multi-tenancy entirely for any user whose tenant is not ID 1.

**Suggestion:** Replace the hardcoded `1` with the actual authenticated tenant ID from auth context (e.g., `useAuth()` or a tenant store), passed through the route or a context provider.

yihua-zhuo: **CRITICAL** (critical)

**Issue:** useTasks(1, '') fetches all tasks ignoring page/search params, while filtered server data (page, statusFilter) is fetched separately but only used for pagination metadata — pagination info will be incorrect when total > 20.

**Suggestion:** Pass all active filters (statusFilter, priorityFilter, assigneeFilter, createdAfter, createdBefore) to the server via useTasks so the server handles pagination correctly. Remove the allData fetch or use it only when no filters are active.

frontend/src/components/layout/app-sidebar.tsx
```diff
@@ -2,10 +2,11 @@
 import Link from "next/link";
 import { usePathname } from "next/navigation";
 import { useAuthStore } from "@/lib/store/auth-store";
+import { useQuickAddTask } from "@/lib/store/task-store";
 import {
   Users, TrendingUp, Ticket, BarChart3, Sparkles, CheckSquare, Bell,
   Activity, UsersRound, Settings as SettingsIcon, Shield, Upload,
```
yihua-zhuo: **CRITICAL** (critical)

**Issue:** SettingsIcon is used in JSX (systemItems) but imported as Settings — this will cause a runtime ReferenceError unless Settings is re-exported as SettingsIcon elsewhere.

**Suggestion:** Change import to: `Settings as SettingsIcon` so the alias matches the usage in systemItems.

frontend/src/app/(app)/layout.tsx
```diff
@@ -71,6 +75,18 @@ export default function AppLayout({ children }: { children: React.ReactNode }) {
 
         <AIPanel />
         <SessionTimeoutGuard />
+
+        <TaskModal
+          open={quickAdd.isOpen}
+          title="Create Task"
+          onClose={quickAdd.close}
+          initialStatus={quickAdd.initialStatus}
+          onSubmit={async (data) => {
```
yihua-zhuo, line 84: **WARNING** (suggestion)

**Issue:** TaskModal onSubmit has no error handling — if create.mutateAsync rejects, the modal stays open silently and the user gets no feedback.

**Suggestion:** Wrap in try/await and call quickAdd.close() only on success; surface the error via toast or setError state on failure.

frontend/src/app/(app)/tasks/list/page.tsx
yihua-zhuo: **WARNING** (suggestion)

**Issue:** useUsers(1, 100) is called for the assignee dropdown but ignores the search parameter — a tenant with many users will have an unfiltered 100-user payload.

**Suggestion:** Add a search input for assignees or paginate with a search term if user list is large.

frontend/src/app/(app)/tasks/list/page.tsx
```diff
@@ -1,8 +1,9 @@
 "use client";
 import { useState } from "react";
-import { useTasks, useDeleteTask, useCompleteTask, useUpdateTask } from "@/lib/api/queries";
+import { useTasks, useDeleteTask, useCompleteTask, useUpdateTask, useUsers } from "@/lib/api/queries";
```
yihua-zhuo, line 3: **INFO** (suggestion)

**Issue:** SortIconStatic is defined but never called in this file — the static ↕ character is rendered directly in JSX instead.

**Suggestion:** Remove the unused SortIconStatic function.

frontend/src/app/(app)/tasks/page.tsx
```diff
@@ -10,28 +25,45 @@ const COLUMNS = [
...
+  isDragging = false,
+}: {
+  task: TaskData;
   onClick: () => void;
   onComplete?: (id: number) => void;
+  isDragging?: boolean;
 }) {
   const due = task.due_date ? new Date(String(task.due_date)) : null;
   const isOverdue = due && due < new Date() && task.status !== "completed";
   const rawId = task.id;
```
yihua-zhuo, line 60: **WARNING** (suggestion)

**Issue:** filteredAndSorted() / useMemo recalculates on every render — called multiple times per render cycle for tasksByStatus, statusCounts, and countByStatus.

**Suggestion:** Memoize the derived values: tasksByStatus and statusCounts already use useMemo but the returned functions (countByStatus, displayed) are recomputed inline; ensure the render-time calls are derived from memoized values.

frontend/src/app/(app)/tasks/page.tsx
```diff
@@ -95,32 +157,92 @@ function TaskCard({ task, onClick, onComplete }: {
...
+    useSensor(PointerSensor, {
+      activationConstraint: { distance: 8 },
+    })
+  );
+
+  function handleDragStart(event: DragStartEvent) {
+    setActiveTaskId(event.active.id);
+  }
+
+  async function handleDragEnd(event: DragEndEvent) {
```
yihua-zhuo, line 241: **WARNING** (suggestion)

**Issue:** handleDragEnd calls update.mutateAsync without disabling the UI — a second drag-drop during the in-flight request will overwrite the first update with a stale task position.

**Suggestion:** Track a `isDragging` boolean state; disable drop targets or show a loading indicator while update.isPending is true.

frontend/src/app/(app)/tasks/page.tsx
```diff
@@ -1,5 +1,20 @@
 "use client";
-import { useState } from "react";
+import { useState, useMemo } from "react";
+import {
```
yihua-zhuo, line 3: **INFO** (suggestion)

**Issue:** ReactMarkdown renders task.description as HTML — if descriptions are user-supplied, this is an XSS vector despite react-markdown's default HTML stripping (plugins can re-enable raw HTML).

**Suggestion:** Pass a remarkPlugins array that explicitly disables dangerouslySetInnerHTML, or use a sanitizer (e.g., DOMPurify) on the rendered output.

frontend/src/app/(app)/tasks/task-modal.tsx
```diff
@@ -1,5 +1,6 @@
 "use client";
-import { useState } from "react";
+import { useState, useEffect } from "react";
+import ReactMarkdown from "react-markdown";
```
yihua-zhuo, line 3: **INFO** (suggestion)

**Issue:** ReactMarkdown renders task.description as HTML — same XSS consideration as in tasks/page.tsx.

**Suggestion:** Use a sanitizer like DOMPurify on the description before passing to ReactMarkdown, or restrict remarkPlugins to prevent raw HTML injection.

frontend/src/app/(app)/tasks/task-modal.tsx
```diff
@@ -127,16 +192,89 @@ export function TaskModal({
...
+            <div className="space-y-1.5">
+              <label className="text-sm font-medium" htmlFor="task-customer">Customer</label>
+              <Select
+                value={form.customer_id || "__none__"}
+                onValueChange={(v) => setFormField("customer_id", v === "__none__" ? "" : v)}
+              >
+                <SelectTrigger id="task-customer">
+                  <SelectValue placeholder="Search customer..." />
+                </SelectTrigger>
+                <SelectContent>
```
yihua-zhuo, line 235: **INFO** (suggestion)

**Issue:** Customer search input is rendered inside SelectContent (Radix UI) — Radix portals SelectContent on each open, which can cause excessive re-renders or state loss with large customer lists.

**Suggestion:** Lift the customer search into a separate combobox pattern or a dedicated popover outside the Select to avoid Radix rendering lifecycle issues.

frontend/src/app/(app)/tasks/task-modal.tsx
```diff
@@ -45,12 +48,39 @@ export function TaskModal({
...
+  const { data: opportunitiesData } = useOpportunities(1, 100);
+
+  const [debouncedSearch, setDebouncedSearch] = useState<string | undefined>(undefined);
+  useEffect(() => {
+    const timer = setTimeout(() => setDebouncedSearch(customerSearch), 300);
+    return () => clearTimeout(timer);
+  }, [customerSearch]);
+
+  // Sync initialStatus into form status when task is not being edited
+  useEffect(() => {
```
yihua-zhuo, line 71: **INFO** (suggestion)

**Issue:** Effect that syncs initialStatus into form.status has task as a dependency but task never changes (only one effect run on mount) — fragile if task prop reference changes.

**Suggestion:** Use initialStatus directly in the form state initializer or add a comment explaining the one-time sync intent, or guard with task === undefined.

frontend/src/components/layout/app-sidebar.tsx
```diff
@@ -86,7 +87,8 @@ export function AppSidebar({ collapsed, onToggle }: { collapsed: boolean; onTogg
   const { user, clearAuth } = useAuthStore();
   const router = useRouter();
   const { data: notifData } = useNotifications(1, true);
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** useNotifications(1, true) passes read=true to fetch read notifications, but the sidebar badge is titled 'unreadCount' and is meant to show unread notifications.

**Suggestion:** Change to useNotifications(1, false) if unread notifications should appear as the badge count, or verify the API contract for the read parameter.

frontend/src/components/layout/app-sidebar.tsx
yihua-zhuo: **INFO** (suggestion)

**Issue:** Shield and Upload are imported but not used in crmItems or systemItems.

**Suggestion:** Remove unused imports: Shield, Upload.

frontend/src/lib/store/auth-store.test.ts
```diff
@@ -1,31 +1,32 @@
 import { describe, it, expect, vi, beforeEach } from "vitest";
 import { useAuthStore } from "./auth-store";
 
+vi.mock("crypto-js", () => ({
```
yihua-zhuo, line 4: **WARNING** (suggestion)

**Issue:** vi.mock for crypto-js is module-level and not reset in beforeEach — if the crypto-js mock state or implementation changes in a future test, it could bleed into other test runs.

**Suggestion:** Ensure beforeEach or afterEach explicitly resets or verifies mock state, or isolate tests that depend on specific mock behavior.

frontend/src/lib/store/auth-store.test.ts
```diff
@@ -1,31 +1,32 @@
...
+        },
+      }),
+    },
+    enc: {
+      Utf8: {},
+    },
+  },
+}));
+
 describe("auth-store", () => {
```
yihua-zhuo, line 28: **INFO** (suggestion)

**Issue:** Tests only cover setAuth, clearAuth, and isAuthenticated — clearAuth also relies on the mock and is not tested independently of setAuth.

**Suggestion:** Add a test for clearAuth without a prior setAuth call to verify it is idempotent and handles the null state correctly.

frontend/src/lib/store/task-store.ts
```diff
@@ -0,0 +1,16 @@
+import { create } from "zustand";
+
+interface QuickAddState {
+  isOpen: boolean;
+  initialStatus: string;
+  open: (initialStatus?: string) => void;
```
yihua-zhuo, line 6: **INFO** (suggestion)

**Issue:** open accepts initialStatus as optional (?) but also assigns a default ='pending', making the ? redundant — TypeScript will treat it as optional string anyway.

**Suggestion:** Remove the ?: void union and write open: (initialStatus?: string) => void, or simplify to open: (initialStatus = 'pending') => void if callers never pass undefined explicitly.

tests/unit/conftest.py
```diff
@@ -593,23 +817,125 @@ def campaign_handler(sql_text, params):
...
+                        "created_by": params.get("created_by", 0),
+                        "completed_at": None,
+                        "created_at": params.get("created_at"),
+                        "updated_at": params.get("updated_at"),
+                    }
+                )
+            ]
+        )
+
+    if "from tasks where id" in sql_text:
```
yihua-zhuo, line 884: **WARNING** (suggestion)

**Issue:** task_handler checks 'insert into tasks' without an early return, and falls through to the 'from tasks where id' guard which would not match INSERT — but the guard ordering is non-obvious and fragile (rule 15).

**Suggestion:** Add an explicit 'return MockResult(...)' inside the 'insert into tasks' block and a final bare 'return None' after all guards, to make control flow unambiguous.

tests/unit/test_tasks.py
```diff
@@ -0,0 +1,315 @@
...
+    "updated_at": None,
+}
+
+
+# ---------------------------------------------------------------------------
+# Test fixture
+# ---------------------------------------------------------------------------
+
+
+@pytest.fixture
```
yihua-zhuo, line 74: **WARNING** (suggestion)

**Issue:** client_with_service creates a single mock_service instance shared across every test method via a lambda closure — when one test assigns svc.create_task = AsyncMock(...) it mutates the shared object, risking state bleed between tests if execution order changes (rule 65, 63).

**Suggestion:** Make client_with_service a factory (accept no arguments, create fresh mocks inside) so each test gets an independently constructed mock_service.

tests/unit/test_tasks.py
```diff
@@ -0,0 +1,315 @@
...
+        )
+        assert resp.status_code == 201
+
+
+# ---------------------------------------------------------------------------
+# GET /api/v1/tasks — list tasks
+# ---------------------------------------------------------------------------
+
+
+class TestListTasksEndpoint:
```
yihua-zhuo, line 186: **INFO** (suggestion)

**Issue:** test_delete_task omits tenant_id_web from the signature, while every other test method in the class includes it — inconsistent with the established pattern (rule 20).

**Suggestion:** Add tenant_id_web: int to the method signature for consistency: async def test_delete_task(self, api_client: AsyncClient, tenant_id_web: int).

tests/unit/test_tasks.py
```diff
@@ -0,0 +1,315 @@
...
+        body = resp.json()
+        assert body["success"] is True
+        assert "items" in body["data"]
+        assert "total" in body["data"]
+        assert "has_next" in body["data"]
+
+    async def test_filter_by_status(self, client_with_service):
+        client, svc = client_with_service
+        svc.list_tasks = AsyncMock(return_value=([], 0))
+        resp = await client.get("/api/v1/tasks?status=pending")
```
yihua-zhuo, line 202: **INFO** (suggestion)

**Issue:** test_delete_task_not_found omits tenant_id_web from the signature for the same reason as test_delete_task — inconsistent with the class pattern (rule 20).

**Suggestion:** Add tenant_id_web: int to the method signature for consistency.

frontend/src/app/(app)/tasks/list/page.tsx
```diff
@@ -29,21 +30,22 @@ const PRIORITY_COLORS: Record<string, string> = {
...
   const [sortKey, setSortKey] = useState<SortKey>("created_at");
   const [sortDir, setSortDir] = useState<SortDir>("desc");
   const [editTask, setEditTask] = useState<Record<string, unknown> | null>(null);
+  const [selectedIds, setSelectedIds] = useState<Set<number>>(new Set());
 
-  const { data, isLoading, isError } = useTasks(page, statusFilter);
+  const { data, isLoading, isError } = useTasks(page, statusFilter, priorityFilter, assigneeFilter, createdAfter, createdBefore);
   const items = (data?.data?.items ?? []) as Record<string, unknown>[];
   const info = data?.data;
+  const users = (usersData?.data?.items ?? []) as Record<string, unknown>[];
```
yihua-zhuo, line 48: **CRITICAL** (critical)

**Issue:** Variable `usersData` is referenced but never defined; `useUsers` is imported and called via `useTasks` but its result is stored in an undefined variable.

**Suggestion:** Add `const { data: usersData } = useUsers(1, 100);` alongside the `useTasks` call on line 45, or destructure it there.

frontend/src/app/(app)/tasks/page.tsx
```diff
@@ -95,32 +157,98 @@ function TaskCard({ task, onClick, onComplete }: {
...
+  }
+
+  async function handleDragEnd(event: DragEndEvent) {
+    const { active, over } = event;
+    if (!over) return;
+    setIsDragging(false);
+    const taskId = Number(active.id);
+    const newStatus = over.id as string;
+    try {
+      await update.mutateAsync({ id: taskId, data: { status: newStatus } });
```
yihua-zhuo, line 248: **CRITICAL** (critical)

**Issue:** Drag-end status update silently swallows errors with an empty catch block; the user receives no feedback if the API call fails.

**Suggestion:** Log the error or call a toast/notification utility inside the catch block so the user knows the move failed.

tests/integration/test_services_integration.py
```diff
@@ -310,8 +310,9 @@ async def test_list_tasks(self, db_schema, tenant_id, async_session):
             tenant_id=tenant_id,
         )
 
-        result = await svc.list_tasks(tenant_id)
-        titles = [t.title for t in result]
+        items, total = await svc.list_tasks(tenant_id)
```
yihua-zhuo, line 313: **CRITICAL** (critical)

**Issue:** Rule 100: list_tasks called as svc.list_tasks(tenant_id) with only one positional arg — the signature is list_tasks(tenant_id, assigned_to, status, page, page_size), so this passes tenant_id to the assigned_to parameter and relies on defaults for the rest.

**Suggestion:** Use keyword argument: svc.list_tasks(tenant_id=tenant_id)

tests/unit/conftest.py
```diff
@@ -593,23 +817,126 @@ def campaign_handler(sql_text, params):
...
+                        "priority": params.get("priority", "normal"),
+                        "created_by": params.get("created_by", 0),
+                        "completed_at": None,
+                        "created_at": params.get("created_at"),
+                        "updated_at": params.get("updated_at"),
+                    }
+                )
+            ]
+        )
+        return MockResult([])
```
yihua-zhuo, line 882: **CRITICAL** (critical)

**Issue:** Unreachable `return MockResult([])` immediately after `return MockResult([...])` in task_handler INSERT block — line 882 can never execute.

**Suggestion:** Delete line 882. The INSERT branch already returns at line 862.

tests/unit/conftest.py
```diff
@@ -593,23 +817,126 @@ def campaign_handler(sql_text, params):
...
+                        "updated_at": None,
+                    }
+                )
+            ]
+        )
+
+    return None
+
+
+def task_handler(sql_text, params):
```
yihua-zhuo, line 858: **CRITICAL** (critical)

**Issue:** task_handler (used by all_handlers) implements only INSERT and SELECT; UPDATE and DELETE for tasks are not handled. Integration tests that call PATCH /api/v1/tasks/{id} or DELETE /api/v1/tasks/{id} will get MockResult([]) from the catch-all, causing downstream test failures.

**Suggestion:** Add UPDATE and DELETE branches to task_handler, e.g.:
  if sql_text.startswith('update') and 'tasks' in sql_text: ...
  if sql_text.startswith('delete') and 'tasks' in sql_text: ...

.plans/issue-63.md
```diff
@@ -0,0 +1,78 @@
...
+- `frontend/src/components/layout/app-sidebar.tsx` — **Modify** — add global quick-add task button in header
+- `tests/unit/test_tasks.py` — **New** — unit tests for task service + router
+- `tests/integration/test_tasks_integration.py` — **New** — integration tests for task endpoints
+
+## Implementation Steps
+
+### Backend
+1. Create `src/api/routers/tasks.py` with these endpoints, matching the patterns in `tickets.py` and the hook calls in `frontend/src/lib/api/queries.ts`:
+   - `GET /api/v1/tasks` — list with `page`, `page_size`, `status`, `assigned_to` query params; returns `{success, data: {items, total, page, page_size, has_next}}`
+   - `POST /api/v1/tasks` — create; body fields `title`, `description`, `priority`, `status`, `due_date`, `assigned_to`
```
yihua-zhuo, line 24: **WARNING** (suggestion)

**Issue:** POST /api/v1/tasks endpoint is described with no response status code, while other endpoints have explicit codes — inconsistent with the acceptance criteria stating 201.

**Suggestion:** Add 'returns 201' to the POST endpoint description to match the acceptance criteria on line 68.

.plans/issue-63.md
```diff
@@ -0,0 +1,78 @@
...
+   - `POST /api/v1/tasks/{task_id}/complete` — sets `status=completed`
+   - All endpoints use `ctx: AuthContext = Depends(require_auth)` and `session: AsyncSession = Depends(get_db)`
+2. Export `tasks_router` from `src/api/__init__.py`
+3. Register `tasks_router` in `src/main.py` via `app.include_router(tasks_router)`
+
+### Frontend — Kanban Enhancements
+4. In `frontend/src/app/(app)/tasks/page.tsx`:
+   - Add `@dnd-kit/core` + `@dnd-kit/sortable` for drag-and-drop between columns (install if not in `package.json`)
+   - Make task cards draggable with `useSortable`, columns are droppable zones keyed by `col.id`
+   - On drop, call `useUpdateTask().mutateAsync({ id, data: { status: targetColumnId } })`
```
yihua-zhuo, line 37: **INFO** (suggestion)

**Issue:** Plan references removing redundant per-status API calls but doesn't account for the tradeoff: fetching all tasks with no filter to compute counts may be costlier than dedicated status-count endpoints for large datasets.

**Suggestion:** Clarify that the full dataset fetch (useTasks(1, "")) is acceptable given expected task volume, or consider a dedicated counts endpoint as a future optimization.

.plans/issue-63.md
```diff
@@ -0,0 +1,78 @@
...
+   - Display accurate column counts by computing `allTasks.filter(t => t.status === col.id).length` from the full dataset; remove the redundant per-status API calls and use only `useTasks(1, "")`
+
+### Frontend — TaskModal Enhancements
+5. In `frontend/src/app/(app)/tasks/task-modal.tsx`:
+   - Add `react-markdown` or `@uiw/react-md-editor` for description preview/edit toggle
+   - Add assignee `Select` dropdown using `useUsers()` hook to populate user list
+   - Add optional customer and opportunity link selectors using `useSearchCustomers()` and `useOpportunities()` (or `usePipelines()`) — these store `customer_id` and `opportunity_id` in the task payload but the backend TaskModel stores them via extra fields or a relation
```
yihua-zhuo, line 44: **WARNING** (suggestion)

**Issue:** Plan says 'the backend TaskModel stores them via extra fields or a relation' — this ambiguous wording could lead to inconsistent implementations where customer_id/opportunity_id are added differently in the model vs. the API.

**Suggestion:** Specify exactly how customer_id and opportunity_id are stored on the backend (e.g. nullable columns on TaskModel) before frontend work begins.

frontend/src/app/(app)/layout.tsx
```diff
@@ -17,12 +20,13 @@ export default function AppLayout({ children }: { children: React.ReactNode }) {
     try { return localStorage.getItem(STORAGE_KEY) === "true"; } catch { return false; }
   });
   const pathname = usePathname();
+  const quickAdd = useQuickAddTask();
+  const create = useCreateTask();
+  const update = useUpdateTask();
```
yihua-zhuo, line 25: **WARNING** (suggestion)

**Issue:** useUpdateTask is imported and assigned but never used in the component — dead code clutters the module and may indicate a missing feature or accidental import.

**Suggestion:** Remove the unused `update = useUpdateTask()` line, or wire it up if bulk-update from this layout context is intended.

frontend/src/app/(app)/layout.tsx
```diff
@@ -17,12 +20,13 @@ export default function AppLayout({ children }: { children: React.ReactNode }) {
...
 
-  /* eslint-disable react-hooks/set-state-in-effect -- intentional: synchronizing sidebar open state with pathname change */
   useEffect(() => {
-    setSidebarOpen(false);
-  }, [pathname]);
-  /* eslint-enable react-hooks/set-state-in-effect */
+    if (sidebarOpen) setSidebarOpen(false);
+  }, [pathname, sidebarOpen]);
```
yihua-zhuo, line 29: **WARNING** (suggestion)

**Issue:** useEffect dependency array contains sidebarOpen which is also the state being set inside the effect body — this creates a confusing dependency that ESLint will flag as a rule-of-hooks or exhaustive-deps violation.

**Suggestion:** Move the setSidebarOpen(false) outside the effect and restructure: either use a different trigger (e.g. just pathname change) or eliminate the effect entirely if the intent is just to close the sidebar on navigation.

frontend/src/app/(app)/layout.tsx
```diff
@@ -71,6 +75,22 @@ export default function AppLayout({ children }: { children: React.ReactNode }) {
...
+
+        <TaskModal
+          open={quickAdd.isOpen}
+          title="Create Task"
+          onClose={quickAdd.close}
+          initialStatus={quickAdd.initialStatus}
+          onSubmit={async (data) => {
+            try {
+              await create.mutateAsync({ ...data, status: (data.status as string | undefined) ?? quickAdd.initialStatus });
+              quickAdd.close();
```
yihua-zhuo, line 87: **WARNING** (suggestion)

**Issue:** Empty catch block silently swallows all errors from create.mutateAsync — error is 'handled by the mutation hook' but the comment is misleading since errors would typically trigger the onError callback, not the hook itself.

**Suggestion:** Either remove the try/catch entirely (let errors propagate to the hook's error state) or log the error: console.error('Task creation failed', err). The current empty catch obscures debugging.

frontend/src/app/(app)/layout.tsx
```diff
@@ -17,12 +20,13 @@ export default function AppLayout({ children }: { children: React.ReactNode }) {
     try { return localStorage.getItem(STORAGE_KEY) === "true"; } catch { return false; }
```
yihua-zhuo, line 20: **INFO** (suggestion)

**Issue:** localStorage read in useState initializer runs on the server during SSR — Next.js 13+ App Router disables client-side checks by default so this may not crash, but the pattern is fragile if the file is ever moved or Next.js config changes.

**Suggestion:** Use a lazy initialization pattern or read localStorage only inside a useEffect to guarantee it only runs in the browser context.

frontend/src/app/(app)/tasks/list/page.tsx
```diff
@@ -85,12 +105,52 @@ export default function TaskListPage() {
...
+  function toggleSelect(id: number) {
+    setSelectedIds((prev) => {
+      const next = new Set(prev);
+      if (next.has(id)) next.delete(id);
+      else next.add(id);
+      return next;
+    });
+  }
+
+  function toggleSelectAll() {
```
yihua-zhuo, line 117: **WARNING** (suggestion)

**Issue:** `filteredAndSorted()` call inside `toggleSelectAll` is not memoized, so sorting re-runs on every checkbox toggle.

**Suggestion:** Use the memoized `displayed` value instead of calling the function directly.

frontend/src/app/(app)/tasks/list/page.tsx
```diff
@@ -134,30 +236,47 @@ export default function TaskListPage() {
...
-                </td>
-                <td className="px-3 py-2.5 text-muted-foreground">
-                  {t.due_date ? new Date(String(t.due_date)).toLocaleDateString() : "—"}
-                </td>
-                <td className="px-3 py-2.5 text-muted-foreground">
-                  {t.created_at ? new Date(String(t.created_at)).toLocaleDateString() : "—"}
-                </td>
-              </tr>
-            ))}
+            {isLoading && <tr><td colSpan={7} className="px-3 py-8 text-center text-muted-foreground">Loading…</td></tr>}
```
yihua-zhuo, line 239: **WARNING** (suggestion)

**Issue:** `colSpan={7}` does not match the actual column count (6 columns: checkbox, Title, Status, Priority, Due Date, Created), causing a malformed table layout.

**Suggestion:** Change `colSpan={7}` to `colSpan={6}` on lines 239, 240, and 241.

frontend/src/app/(app)/tasks/page.tsx
```diff
@@ -95,32 +157,98 @@ function TaskCard({ task, onClick, onComplete }: {
...
+  );
+}
+
 export default function TasksPage() {
-  const [editTask, setEditTask] = useState<Record<string, unknown> | null>(null);
+  const [editTask, setEditTask] = useState<TaskData | null>(null);
   const [showCreate, setShowCreate] = useState(false);
   const [createStatus, setCreateStatus] = useState("pending");
+  const [activeTaskId, setActiveTaskId] = useState<string | number | null>(null);
+  const [isDragging, setIsDragging] = useState(false);
```
yihua-zhuo, line 200: **WARNING** (suggestion)

**Issue:** `isDragging` state is declared and updated but never read — the drag-overlay and column opacity use `update.isPending` instead.

**Suggestion:** Remove the `isDragging` state declaration, or use it (e.g., remove `update.isPending` in favor of `isDragging`).

frontend/src/app/(app)/tasks/task-modal.tsx
```diff
@@ -127,16 +194,89 @@ export function TaskModal({
...
+              onChange={(e) => setFormField("due_date", e.target.value)}
             />
           </div>
-          <DialogFooter className="flex-row! justify-between gap-2 sm:justify-between">
+
+          <div className="grid grid-cols-2 gap-3">
+            <div className="space-y-1.5">
+              <label className="text-sm font-medium" htmlFor="task-customer">Customer</label>
+              <Select
+                value={form.customer_id || "__none__"}
```
yihua-zhuo, line 229: **INFO** (suggestion)

**Issue:** Customer search nests an HTML `<input>` directly inside `<SelectContent>`, which is an a11y anti-pattern (interactive elements inside selects are not announced by screen readers).

**Suggestion:** Use a separate Popover + Combobox pattern (e.g., @dnd-kit's combobox or a custom autocomplete) for customer search instead of nesting inputs in SelectContent.

frontend/src/components/layout/app-sidebar.tsx
```diff
@@ -2,10 +2,10 @@
 import Link from "next/link";
 import { usePathname } from "next/navigation";
 import { useAuthStore } from "@/lib/store/auth-store";
+import { useQuickAddTask } from "@/lib/store/task-store";
 import {
   Users, TrendingUp, Ticket, BarChart3, Sparkles, CheckSquare, Bell,
-  Activity, UsersRound, Settings as SettingsIcon, Shield, Upload,
-  LogOut, UserCog, ChevronLeft, ChevronRight,
+  Activity, UsersRound, Settings, LogOut, UserCog, ChevronLeft, ChevronRight, Plus,
 } from "lucide-react";
```
yihua-zhuo, line 9: **INFO** (suggestion)

**Issue:** `Activity` is imported from lucide-react twice in the same import statement (line 9 shows it listed, and line 7 also lists it), which is redundant.

**Suggestion:** Deduplicate the `Activity` import into a single occurrence.

frontend/src/lib/api/queries.ts
```diff
@@ -307,10 +307,14 @@ export function useCreateUser() {
 }
 // ── Tasks ───────────────────────────────────────────────────────────────────
 
-export function useTasks(page = 1, status = "") {
+export function useTasks(page = 1, status = "", priority = "", assigned_to = "", createdAfter = "", createdBefore = "") {
```
yihua-zhuo, line 310: **INFO** (suggestion)

**Issue:** useTasks constructs a queryKey with only page and status via qk.tasks, but the hook also filters by priority, assigned_to, createdAfter, createdBefore — meaning queries with different filter values will share a cache key and return stale data.

**Suggestion:** Extend qk.tasks to include all filter params and pass them to the qk call, e.g. qk.tasks(page, status, priority, assigned_to, createdAfter, createdBefore).

frontend/src/lib/api/queries.ts
```diff
@@ -307,10 +307,14 @@ export function useCreateUser() {
...
   const token = useAuthStore((s) => s.token);
   const params = new URLSearchParams({ page: String(page), page_size: "20" });
   if (status) params.set("status", status);
+  if (priority) params.set("priority", priority);
+  if (assigned_to) params.set("assigned_to", assigned_to);
+  if (createdAfter) params.set("created_after", createdAfter);
+  if (createdBefore) params.set("created_before", createdBefore);
   return useQuery({
     queryKey: qk.tasks(page, status),
```
yihua-zhuo, line 319: **INFO** (suggestion)

**Issue:** useActivities queryKey only includes page and type but not page_size, though the implementation hardcodes page_size=20 — this is fine but slightly misleading given other hooks expose page_size as a parameter.

**Suggestion:** Either expose page_size as a parameter in useActivities or remove it from consideration here.

frontend/src/lib/store/auth-store.test.ts
```diff
@@ -1,34 +1,39 @@
...
-        },
-        enc: {
-          Utf8: {},
-        },
-      },
-    }));
     useAuthStore.setState({ token: null, user: null, isHydrated: true });
   });
 
+  afterEach(() => {
```
yihua-zhuo, line 33: **INFO** (suggestion)

**Issue:** afterEach calls vi.restoreAllMocks() after every test, but vi.mock is module-level and the per-test beforeEach already resets store state — restoring all mocks is unnecessary overhead.

**Suggestion:** Remove the afterEach block entirely; it provides no value given beforeEach already isolates test state via setState.

frontend/src/lib/store/task-store.ts
```diff
@@ -0,0 +1,16 @@
...
+  close: () => void;
+}
+
+export const useQuickAddTask = create<QuickAddState>((set) => ({
+  isOpen: false,
+  initialStatus: "pending",
+  open: (initialStatus = "pending") =>
```
yihua-zhuo, line 13: **WARNING** (suggestion)

**Issue:** Mutable default argument initialStatus = 'pending' is a JavaScript antipattern — while primitives are safe here, the pattern is easy to get wrong if the default is later changed to an object or array (Rule 22).

**Suggestion:** Normalize the default inside the function body: open: (initialStatus?: string) => set({ isOpen: true, initialStatus: initialStatus ?? 'pending' })

src/api/routers/tasks.py
```diff
@@ -0,0 +1,161 @@
...
+from pydantic import BaseModel, Field
+from sqlalchemy.ext.asyncio import AsyncSession
+
+from db.connection import get_db
+from internal.middleware.fastapi_auth import AuthContext, require_auth
+from services.task_service import TaskService
+
+tasks_router = APIRouter(prefix="/api/v1", tags=["tasks"])
+
+UNASSIGNED = 0
```
yihua-zhuo, line 19: **INFO** (suggestion)

**Issue:** UNASSIGNED = 0 is a module-level constant but 0 is not self-documenting; it's effectively a magic number.

**Suggestion:** Use a named constant from the models layer (e.g. from db.models.task import UNASSIGNED_ID) or document why 0 is the sentinel.

src/api/routers/tasks.py
```diff
@@ -0,0 +1,161 @@
...
+    session: AsyncSession = Depends(get_db),
+):
+    service = TaskService(session)
+    task = await service.create_task(
+        title=body.title,
+        description=body.description,
+        priority=body.priority,
+        tenant_id=ctx.tenant_id,
+        created_by=ctx.user_id,
+        assigned_to=body.assigned_to or UNASSIGNED,
```
yihua-zhuo, line 90: **WARNING** (suggestion)

**Issue:** Body.assigned_to=0 is falsy so the `or UNASSIGNED` fallback is a no-op and obscures intent; explicit None-check reads clearer.

**Suggestion:** Use `assigned_to=body.assigned_to if body.assigned_to is not None else UNASSIGNED` for clarity.

tests/integration/test_services_integration.py
yihua-zhuo, line 313: **WARNING** (suggestion)

**Issue:** Rule 63: test_list_tasks has a shared-tenant assertion (total >= 2) that can be flaky if other tests run in the same session and create tasks for different suffixes.

**Suggestion:** Add a tenant_id filter or status filter to list_tasks call so the test is isolated to its own data. Use a unique status value or a named assigned_to user so cross-test pollution cannot affect the assertion.

tests/integration/test_tasks_integration.py
```diff
@@ -0,0 +1,238 @@
...
+
+        # Verify it's gone
+        get_resp = await api_client.get(f"/api/v1/tasks/{created_id}")
+        assert get_resp.status_code == 404
+
+    async def test_delete_task_not_found(self, api_client: AsyncClient, tenant_id_web: int):
+        resp = await api_client.delete("/api/v1/tasks/999999999")
+        assert resp.status_code == 404
+
+    async def test_task_lifecycle(self, api_client: AsyncClient):
```
yihua-zhuo, line 204: **INFO** (suggestion)

**Issue:** test_task_lifecycle accepts tenant_id_web: int parameter but never references it, making the parameter dead weight.

**Suggestion:** Remove `tenant_id_web: int` from the method signature, or use it to assert cross-tenant isolation within the lifecycle test.

PR #373, opened 2026-05-13
Auto: implement issue #65

### LINKED ISSUE #65: [UI] Automation Rules — Visual rule builder for workflow automation
## Problem

automation_rules.py has rule engine APIs but no UI. Admin cannot manage automation rules.

## Dependencies

**Blocked by:** #41 (AI Agent Framework — must complete first)
**Blocked by:** #37 (workflow_service — backend must be complete)

### Rules List (`/automation/rules`)
- [ ] Table: name, trigger, status, last run, actions
- [ ] Enable / disable toggle per rule
- [ ] Edit / delete actions
- [ ] Run history link (last 10 runs)
- [ ] "Create Rule" button

### Rule Builder (`/automation/rules/new`)
- [ ] Rule name
- [ ] IF trigger selector:
  - Ticket: created / updated / assigned
  - Opportunity: stage_changed / created
  - Customer: churn_risk_high / created
  - etc.
- [ ] Conditions (AND/OR):
  - Field / operator / value
  - Add/remove condition buttons
- [ ] THEN actions:
  - Notification: send to user/role
  - Assign: to user/team
  - Update: field → value
  - Task: create with title
  - Email: send template
- [ ] Save as draft / Activate

### Rule Templates
- [ ] Gallery of pre-built rules
- [ ] One-click activate
- [ ] Templates: "High value ticket alert", "Win-back campaign", etc.

### Execution History
- [ ] Log per rule execution
- [ ] Trigger event details
- [ ] Actions executed
- [ ] Errors if any

yihua-zhuo:
Closes #65

Plan: `.plans/issue-65.md`

1. **Plan** — Claude wrote a detailed implementation plan to `.plans/issue-65.md`.

Please review before merging.

## Summary by CodeRabbit

* **New Features**
  * Full Automation Rules UI: list, view, create, edit, delete, enable/disable, templates gallery, sidebar entry
  * Rule Builder: trigger selector, dynamic condition/action rows, validation, draft vs activate flow
  * Execution History: global logs and per-rule history with pagination, status filters, and status badges

* **UI**
  * Added additional badge color variants for statuses
  * Per-action parameter validation for common action types

* **Tests**
  * New integration and unit tests covering rule CRUD, toggling, triggering, and logs

* **Documentation**
  * Added implementation plan for rollout

[Review Change Stack](https://app.coderabbit.ai/change-stack/yihua-zhuo/agent-job/pull/373)


frontend/src/app/(app)/automation/rules/templates/page.tsx
```diff
@@ -0,0 +1,191 @@
...
+    </Card>
+  );
+}
+
+export default function TemplatesPage() {
+  const router = useRouter();
+  const createRule = useCreateAutomationRule();
+  const [activatingTemplate, setActivatingTemplate] = useState<string | null>(null);
+
+  async function handleActivate(template: (typeof TEMPLATES)[number]) {
```
yihua-zhuo, line 134: **CRITICAL** (critical)

**Issue:** The early-return guard checks createRule.isPending at call-time but does not re-check after setActivatingTemplate; two rapid clicks can both pass the guard before the state update flushes, causing a double-mutation race.

**Suggestion:** Gate with activatingTemplate immediately: if (activatingTemplate !== null) return; setActivatingTemplate(template.name); await createRule.mutateAsync(...); and always reset in finally.

frontend/src/app/(app)/automation/rules/templates/page.tsx
```diff
@@ -0,0 +1,191 @@
...
+import { ArrowLeft, Zap } from "lucide-react";
+import Link from "next/link";
+import { useCreateAutomationRule } from "@/lib/api/queries";
+
+const TEMPLATES = [
+  {
+    name: "High-Value Ticket Alert",
+    description: "Notify the team whenever a high-priority ticket is created.",
+    trigger: "ticket.created",
+    conditions: [{ field: "priority", operator: "eq", value: "high" }],
```
yihua-zhuo: **CRITICAL** (critical)

**Issue:** TEMPLATES[0].conditions uses {field, operator, value} but RuleBuilderValues expects {field: {key, label}, operator, value}; the mismatch will cause runtime errors when the RuleBuilder tries to access condition.field.key.

**Suggestion:** Align template conditions to match RuleBuilderValues, e.g. conditions: [{ field: { key: "priority", label: "Priority" }, operator: "eq", value: "high" }].

frontend/src/components/automation/action-row.tsx
```diff
@@ -0,0 +1,203 @@
...
+  function updateParam(key: string, val: string) {
+    onChange({ ...value.params, [key]: val });
+  }
+
+  switch (value.type) {
+    case "notification.send":
+      return (
+        <div className="flex gap-2 mt-2">
+          <Input
+            placeholder="user_id"
```
yihua-zhuo, line 55: **CRITICAL** (critical)

**Issue:** Email "to" field has no input validation or sanitization — malformed addresses are accepted silently.

**Suggestion:** Add basic email format validation (e.g. a regex or HTML5 type="email") and surface a validation error for invalid input.

frontend/src/components/automation/action-row.tsx
```diff
@@ -0,0 +1,203 @@
...
+            className="flex-1"
+          />
+        </div>
+      );
+    case "webhook.call":
+      return (
+        <div className="flex gap-2 mt-2">
+          <Input
+            placeholder="url"
+            value={value.params.url ?? ""}
```
yihua-zhuo: **CRITICAL** (critical)

**Issue:** Webhook URL field has no URL validation — arbitrary strings are accepted as the URL parameter.

**Suggestion:** Validate that the URL is a well-formed URI (e.g. using a URL constructor check or regex) before passing to the backend.

frontend/src/components/automation/condition-row.tsx
```diff
@@ -0,0 +1,94 @@
...
+    onChange({ ...value, operator });
+  }
+  function updateValue(newValue: string) {
+    onChange({ ...value, value: newValue });
+  }
+
+  return (
+    <div className="flex items-center gap-2">
+      <Input
+        placeholder="field name"
```
yihua-zhuo: **CRITICAL** (critical)

**Issue:** Field name input accepts arbitrary user strings with no sanitization — if rendered directly in the UI or sent to a backend without encoding, this is an XSS vector.

**Suggestion:** Sanitize field names before rendering or sending (escape HTML entities) and validate against an allowed character set (e.g. alphanumeric + dots/underscores).

frontend/src/components/automation/condition-row.tsx
```diff
@@ -0,0 +1,94 @@
...
+        </SelectTrigger>
+        <SelectContent>
+          {OPERATORS.map((op) => (
+            <SelectItem key={op.value} value={op.value}>
+              {op.label}
+            </SelectItem>
+          ))}
+        </SelectContent>
+      </Select>
+      <Input
```
yihua-zhuo, line 74: **CRITICAL** (critical)

**Issue:** Condition value input accepts arbitrary strings with no sanitization — same XSS risk as the field name input.

**Suggestion:** Sanitize the value field on output/render and validate input format where the domain expects a specific type (e.g. numeric comparison fields).

frontend/src/components/automation/rule-builder.tsx
```diff
@@ -0,0 +1,267 @@
...
+            <p className="text-sm font-medium">
+              {enabled ? "Active" : "Disabled"}
+            </p>
+            <p className="text-xs text-muted-foreground">
+              {enabled ? "Rule will run automatically when triggered" : "Save as draft without activating"}
+            </p>
+          </div>
+        </div>
+        <div className="flex gap-3">
+          <Button type="button" onClick={doSubmit} disabled={disabled}>
```
yihua-zhuo: **CRITICAL** (critical)

**Issue:** The primary submit Button has type="button" instead of type="submit", preventing form submission via the Enter key and bypassing native HTML form validation.

**Suggestion:** Change type="button" to type="submit" on the submit Button. The onClick handler is redundant since the form onSubmit already calls doSubmit.

tests/integration/test_automation_rules_ui_integration.py
```diff
@@ -0,0 +1,262 @@
...
+                "conditions": [],
+                "actions": [{"type": "notification.send", "params": {"message": "New customer!"}}],
+                "enabled": True,
+            },
+        )
+
+        # Trigger the event
+        trigger_resp = await api_client.post(
+            "/api/v1/automation/trigger",
+            json={
```
yihua-zhuo, line 173: **CRITICAL** (critical)

**Issue:** automation_rules and automation_logs tables are missing from _TABLES in tests/integration/conftest.py, so TRUNCATE CASCADE and fresh_schema DROP will never clean those tables between tests.

**Suggestion:** Add 'automation_rules' and 'automation_logs' to the _TABLES list in tests/integration/conftest.py (line ~171).

frontend/src/app/(app)/automation/rules/[id]/edit/page.tsx
```diff
@@ -0,0 +1,149 @@
...
+          name: values.name,
+          description: values.description,
+          trigger_event: values.trigger_event,
+          conditions: values.conditions,
+          actions: values.actions.map((a) => ({ type: a.type, params: a.params })),
+          enabled: values.enabled,
+        },
+      });
+      router.push("/automation/rules");
+    } catch {
```
Comment (yihua-zhuo) [inline frontend/src/app/(app)/automation/rules/[id]/edit/page.tsx]: **WARNING** (suggestion)

**Issue:** Bare catch block silently swallows the error with no logging or user-visible feedback (beyond a pre-set message). If the mutation rejects with an unexpected error shape, the user sees only the generic message.

**Suggestion:** Log the error internally (console.error) and ensure the catch block only falls through to the generic message when the error is truly expected (e.g. network failure, 422).

frontend/src/app/(app)/automation/rules/[id]/page.tsx
```diff
@@ -0,0 +1,161 @@
+"use client";
+import { useRouter, useParams } from "next/navigation";
```
Comment (yihua-zhuo) [inline frontend/src/app/(app)/automation/rules/[id]/page.tsx:2]: **INFO** (suggestion)

**Issue:** useRouter is imported but router is never used in RuleDetailPageInner (line 32). Rule 11: variable names are meaningful; unused imports bloat the bundle.

**Suggestion:** Remove the unused import: import { useParams } from "next/navigation";

frontend/src/app/(app)/automation/rules/[id]/page.tsx
```diff
@@ -0,0 +1,161 @@
...
+export default function RuleDetailPage() {
+  return (
+    <AuthGuard>
+      <RuleDetailPageInner />
+    </AuthGuard>
+  );
+}
+
+function RuleDetailPageInner() {
+  const router = useRouter();
```
Comment (yihua-zhuo) [inline frontend/src/app/(app)/automation/rules/[id]/page.tsx:32]: **INFO** (suggestion)

**Issue:** router is destructured from useRouter() but never referenced. Rule 11.

**Suggestion:** Remove the unused variable: const params = useParams<{ id: string }>();

frontend/src/app/(app)/automation/rules/history/[rule_id]/page.tsx
```diff
@@ -0,0 +1,170 @@
...
+  rule_name: string;
+  trigger_event: string;
+  status: string;
+  executed_at: string;
+  error_message?: string;
+}
+
+function PerRuleHistoryContent() {
+  const params = useParams<{ rule_id: string }>();
+  const router = useRouter();
```
Comment (yihua-zhuo) [inline frontend/src/app/(app)/automation/rules/history/[rule_id]/page.tsx]: **INFO** (suggestion)

**Issue:** useRouter is called (router declared at line 43) but never used in PerRuleHistoryContent. Rule 11.

**Suggestion:** Remove const router = useRouter();

frontend/src/app/(app)/automation/rules/history/[rule_id]/page.tsx
```diff
@@ -0,0 +1,170 @@
...
+                  </td>
+                  <td className="px-3 py-3" />
+                </tr>
+              ))}
+            </tbody>
+          </table>
+        )}
+      </div>
+
+      {info && !isError && total > 0 && (
```
Comment (yihua-zhuo) [inline frontend/src/app/(app)/automation/rules/history/[rule_id]/page.tsx:138]: **WARNING** (suggestion)

**Issue:** Pagination controls (prev/next and count display) are gated behind info && total > 0. If the API returns {data: {items: [], total: 0, ...}} (total = 0), the controls are hidden — fine. But if the API returns a 200 with no data field at all, this silently shows nothing, even though the page should still render a 'no logs' state. Not critical but fragile.

**Suggestion:** Consider using total > 0 (not info && ...) to determine whether to show pagination, so the footer appears whenever there are results even if info structure is partial.

frontend/src/app/(app)/automation/rules/history/page.tsx
```diff
@@ -0,0 +1,235 @@
...
+  SelectContent,
+  SelectItem,
+  SelectTrigger,
+  SelectValue,
+} from "@/components/ui/select";
+import { ArrowLeft } from "lucide-react";
+import Link from "next/link";
+import { useRouter } from "next/navigation";
+
+const STATUS_COLORS: Record<string, string> = {
```
yihua-zhuo: **INFO** (suggestion)

**Issue:** STATUS_COLORS constant is defined but never consumed; getStatusVariant() returns string literals directly instead of using the map.

**Suggestion:** Remove STATUS_COLORS or wire getStatusVariant to use it: return STATUS_COLORS[status] ?? "gray".

frontend/src/app/(app)/automation/rules/history/page.tsx
```diff
@@ -0,0 +1,235 @@
...
+  id: number;
+  rule_id: number;
+  rule_name: string;
+  trigger_event: string;
+  status: string;
+  executed_at: string;
+  error_message?: string;
+}
+
+function LogsTable({ logs }: { logs: LogRow[] }) {
```
yihua-zhuo: **INFO** (suggestion)

**Issue:** LogsTable instantiates useRouter internally instead of receiving an onView callback as a prop, making it harder to test and coupling it to the router.

**Suggestion:** Pass an onView: (ruleId: number) => void prop to LogsTable and call it from the button onClick.

frontend/src/app/(app)/automation/rules/page.tsx
```diff
@@ -0,0 +1,300 @@
...
+            <DropdownMenuItem onClick={() => onEdit(rule.id)}>
+              <Pencil className="h-4 w-4 mr-2" />
+              Edit
+            </DropdownMenuItem>
+            <DropdownMenuItem onClick={() => onHistory(rule.id)}>
+              <History className="h-4 w-4 mr-2" />
+              View history
+            </DropdownMenuItem>
+            <DropdownMenuSeparator />
+            <DropdownMenuItem
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** window.confirm inside a JSX onClick handler mixes imperative DOM APIs with React; clicking Enter in the dropdown will also fire the button click, causing double-confirmation.

**Suggestion:** Extract to a named handler function and call e.stopPropagation() on the click event, or replace window.confirm with a dedicated confirmation Dialog component.

frontend/src/app/(app)/automation/rules/page.tsx
```diff
@@ -0,0 +1,300 @@
...
+} from "@/components/ui/dropdown-menu";
+import {
+  MoreHorizontal,
+  Plus,
+  History,
+  Pencil,
+  Trash2,
+} from "lucide-react";
+
+function formatDate(dateStr: string): string {
```
yihua-zhuo, line 26: **INFO** (suggestion)

**Issue:** formatDate is defined in this file but is not called within it; the template literal fallback in the JSX uses raw date strings directly without formatting.

**Suggestion:** Use formatDate on the last_run / updated_at / created_at values, e.g. {formatDate(rule.last_run ?? rule.updated_at ?? rule.created_at)}.

frontend/src/app/(app)/automation/rules/templates/page.tsx
```diff
@@ -0,0 +1,191 @@
+"use client";
+import { useState } from "react";
```
yihua-zhuo: **INFO** (suggestion)

**Issue:** useState is imported but never called; there is no local component state — activatingTemplate is tracked via a string variable.

**Suggestion:** Remove the unused useState import.

frontend/src/components/automation/action-row.tsx
```diff
@@ -0,0 +1,203 @@
...
+  { value: "ticket.assign", label: "Assign Ticket" },
+  { value: "ticket.update_priority", label: "Update Ticket Priority" },
+  { value: "opportunity.add_note", label: "Add Note to Opportunity" },
+  { value: "task.create", label: "Create Task" },
+  { value: "email.send", label: "Send Email" },
+  { value: "webhook.call", label: "Call Webhook" },
+  { value: "tag.add", label: "Add Tag" },
+];
+
+export interface ActionRowValue {
```
yihua-zhuo, line 33: **INFO** (suggestion)

**Issue:** ActionRowValue.type is typed as open string rather than a union of the known ACTION_TYPE values.

**Suggestion:** Define a literal union type (e.g. type ActionType = 'notification.send' | 'ticket.assign' | ...) and use it for ActionRowValue.type instead of string.

frontend/src/components/automation/condition-row.tsx
```diff
@@ -0,0 +1,94 @@
...
+  onChange: (value: ConditionRowValue) => void;
+  onRemove: () => void;
+  disabled?: boolean;
+}
+
+export function ConditionRow({ value, onChange, onRemove, disabled }: ConditionRowProps) {
+  function updateField(field: string) {
+    onChange({ ...value, field });
+  }
+  function updateOperator(operator: string) {
```
yihua-zhuo: **INFO** (suggestion)

**Issue:** updateOperator parameter is typed as open string rather than the narrowed OperatorValue type defined in the codebase.

**Suggestion:** Change the parameter type from string to OperatorValue to match the constrained OPERATORS list.

frontend/src/components/automation/rule-builder.tsx
```diff
@@ -0,0 +1,267 @@
...
+            {enabled ? "Activate Rule" : "Save as Draft"}
+          </Button>
+          {onCancel && (
+            <Button type="button" variant="outline" onClick={onCancel} disabled={disabled}>
```
yihua-zhuo, line 264: **WARNING** (suggestion)

**Issue:** The cancel Button also has type="button" — if this form is nested or has a default action, this prevents unintended submission, but it is correct as-is.

**Suggestion:** No change needed for type="button" on cancel; verify this is intentional if the form ever adds a default action.

frontend/src/components/automation/rule-builder.tsx
```diff
@@ -0,0 +1,267 @@
...
+
+  const removeAction = useCallback((id: string) => {
+    setActions((prev) => prev.filter((a) => a.id !== id));
+  }, []);
+
+  const updateAction = useCallback((id: string, val: ActionRowValue) => {
+    setActions((prev) => prev.map((a) => (a.id === id ? { ...val, id } : a)));
+  }, []);
+
+  async function doSubmit() {
```
yihua-zhuo, line 91: **WARNING** (suggestion)

**Issue:** doSubmit catches no exceptions — if onSubmit(payload) rejects, the error is unhandled and the UI shows no feedback to the user.

**Suggestion:** Wrap await onSubmit(payload) in a try/catch and set a user-facing error message (e.g. setValidationError or a separate error state) on failure.

frontend/src/components/automation/rule-builder.tsx
```diff
@@ -0,0 +1,267 @@
...
+    if (!trigger) {
+      setValidationError("A trigger event is required.");
+      return;
+    }
+
+    const payload: RuleBuilderValues = {
+      name: name.trim(),
+      description: description.trim(),
+      trigger_event: trigger,
+      conditions: conditions.filter((c) => c.field.trim() !== ""),
```
yihua-zhuo, line 112: **INFO** (suggestion)

**Issue:** Condition rows with empty field names are silently filtered out on submit, but no UI feedback is given to the user when they add a condition and leave it blank.

**Suggestion:** Show a per-row validation hint or disable the Add condition button while an empty condition exists, so users are aware their blank row will be discarded.

frontend/src/components/ui/badge.tsx
```diff
@@ -11,6 +11,12 @@ const badgeVariants = cva(
         secondary: "border-transparent bg-secondary text-secondary-foreground hover:bg-secondary/80",
         destructive: "border-transparent bg-destructive text-destructive-foreground hover:bg-destructive/80",
         outline: "text-foreground",
+        blue: "border-transparent bg-blue-100 text-blue-800",
```
yihua-zhuo, line 14: **WARNING** (suggestion)

**Issue:** Variants 'blue', 'green', and 'gray' are semantically ambiguous (blue/green/gray vs success/failed/skipped) and create confusion about intended use.

**Suggestion:** Rename 'blue' to 'info', 'green' to a name matching its semantic purpose, and remove 'gray' if redundant with 'secondary'.

frontend/src/components/ui/badge.tsx
```diff
@@ -11,6 +11,12 @@ const badgeVariants = cva(
...
+        green: "border-transparent bg-green-100 text-green-800",
+        gray: "border-transparent bg-gray-100 text-gray-600",
+        success: "border-transparent bg-green-100 text-green-800",
```
yihua-zhuo, line 17: **WARNING** (suggestion)

**Issue:** 'success' and 'green' variants produce identical visual output (bg-green-100 text-green-800), violating DRY and risking misuse.

**Suggestion:** Remove the 'green' variant or redefine it with a distinct semantic color.

frontend/src/lib/api/queries.ts
```diff
@@ -362,3 +365,87 @@ export function useDeleteReminder() {
...
+  const token = useAuthStore((s) => s.token);
+  return useMutation({
+    mutationFn: (id: number) =>
+      apiClient.post(`/api/v1/automation/rules/${id}/toggle`, {}, token ?? undefined),
+    onSuccess: () => {
+      qc.invalidateQueries({ queryKey: ["automation_rules"] });
+      qc.invalidateQueries({ queryKey: ["automation_logs"] });
+    },
+  });
+}
```
yihua-zhuo, line 538: **WARNING** (suggestion)

**Issue:** useReminders return type is not generic — apiClient.get returns untyped response unlike all other hooks which return ApiEnvelope<T>.

**Suggestion:** Add generic typing: apiClient.get<ApiEnvelope<Record<string, unknown>>>("...", token ?? undefined)

frontend/src/lib/api/queries.ts
```diff
@@ -16,6 +16,9 @@ export const qk = {
   reminders: (upcomingOnly = false) => ["reminders", upcomingOnly] as const,
   activities: (page = 1, type = "") => ["activities", page, type] as const,
   slaBreaches: () => ["sla", "breaches"] as const,
+  automationRules: (page = 1, pageSize = 20) => ["automation_rules", page, pageSize] as const,
+  automationRule: (id: number) => ["automation_rules", "detail", id] as const,
+  automationLogs: (page = 1, pageSize = 20, ruleId?: number, status?: string) => ["automation_logs", page, pageSize, ruleId ?? null, status ?? ""] as const,
 } as const;
 
 interface PaginatedResponse<T> {
```
yihua-zhuo, line 27: **INFO** (suggestion)

**Issue:** qk.automationLogs uses `ruleId ?? null` which creates distinct query keys for ruleId=undefined vs ruleId=null, potentially causing unnecessary cache fragmentation.

**Suggestion:** Use a sentinel value or filter undefined before building the key array.

tests/integration/test_automation_rules_ui_integration.py
```diff
@@ -0,0 +1,262 @@
...
+                "event_type": "customer.created",
+                "context": {"customer_id": 123, "name": "Test Customer"},
+            },
+        )
+        assert trigger_resp.status_code == 200
```
yihua-zhuo, line 178: **INFO** (suggestion)

**Issue:** test_trigger_event_executes_matching_rules only asserts response success; it does not verify that an AutomationLogModel row was actually persisted to the DB, so the trigger-to-log path is unverified.

**Suggestion:** After the POST /trigger call, add a follow-up GET /api/v1/automation/logs?rule_id=<rule_id> and assert that a log entry with status='success' and actions_executed appears.

tests/unit/conftest.py
```diff
@@ -627,6 +634,79 @@ def handler(sql_text, params):
     return handler
 
 
+def make_automation_handler(state: MockState):
```
yihua-zhuo, line 964: **INFO** (suggestion)

**Issue:** make_automation_handler lacks an UPDATE handler for tenant_id mismatch (it silently returns rowcount=0 when the rule exists under a different tenant), letting a cross-tenant update succeed as a no-op rather than raising NotFoundException.

**Suggestion:** In the UPDATE branch, if rid not in state.automation_rules, check whether the id exists under a different tenant_id and return rowcount=0 only then; consider raising so cross-tenant mutations fail loudly.

tests/unit/test_automation_rules_ui.py
```diff
@@ -0,0 +1,474 @@
...
+
+        session.execute = AsyncMock(side_effect=execute_side_effect)
+
+        svc = AutomationService(session)
+        updated = await svc.update_rule(rule_id=5, tenant_id=1, name="Updated Name")
+        assert hasattr(updated, "id")
+        assert hasattr(updated, "to_dict")
+
+    async def test_delete_rule_returns_deleted_id(self, mock_db_session):
+        """AutomationService.delete_rule returns the deleted rule id."""
```
yihua-zhuo, line 639: **WARNING** (suggestion)

**Issue:** test_delete_rule_returns_deleted_id only checks the returned id equals 5 (a hardcoded value), not that the mock's rowcount was consulted or that the service actually validated the tenant before deleting.

**Suggestion:** Assert that session.execute was called and that the returned id is the rule's id from the DB state, not a literal injected from the test.

tests/unit/test_automation_rules_ui.py
```diff
@@ -0,0 +1,474 @@
...
+        items, total = await svc.list_rules(
+            tenant_id=1,
+            page=1,
+            page_size=20,
+            enabled=True,
+        )
+        assert isinstance(items, list)
+        assert isinstance(total, int)
+
+    async def test_update_rule_preserves_unmodified_fields(self, mock_db_session):
```
yihua-zhuo, line 896: **WARNING** (suggestion)

**Issue:** test_update_rule_preserves_unmodified_fields checks only updated.name == 'New Name' but never asserts that trigger_event or description retain their original values — the name of the test is not validated by its assertions.

**Suggestion:** Add assertions for the unmodified fields: assert updated.trigger_event == 'ticket.created' and assert updated.description == 'Original description'.

tests/unit/test_automation_rules_ui.py
```diff
@@ -0,0 +1,474 @@
...
+        # Wire session.execute to return a saved rule on INSERT flush
+        saved_rule = MagicMock()
+        saved_rule.id = 42
+        saved_rule.name = "Test Rule"
+        saved_rule.trigger_event = "ticket.created"
+        saved_rule.enabled = True
+        saved_rule.tenant_id = 1
+        saved_rule.to_dict = MagicMock(return_value={"id": 42, "name": "Test Rule"})
+
+        def execute_side_effect(sql, **kwargs):
```
yihua-zhuo, line 502: **INFO** (suggestion)

**Issue:** test_create_rule_returns_orm_object checks that rule.id==42 and rule.name=='Test Rule' but does not assert that the conditions and actions stored in the ORM object match what was passed in (list equality vs JSONB round-trip).

**Suggestion:** Assert rule.conditions == [{'field': 'priority', 'operator': 'eq', 'value': 'high'}] and rule.actions == [{'type': 'notification.send', ...}] to verify the JSONB fields survive the ORM round-trip.

frontend/src/app/(app)/automation/rules/[id]/edit/page.tsx
```diff
@@ -0,0 +1,150 @@
...
+    conditions: safeConditions,
+    actions: safeActions,
+    enabled: Boolean(raw.enabled),
+  };
+}
+
+function EditRulePageInner() {
+  const router = useRouter();
+  const params = useParams<{ id: string }>();
+  const ruleId = Number(params.id);
```
Comment (yihua-zhuo) [inline frontend/src/app/(app)/automation/rules/[id]/edit/page.tsx:57]: **CRITICAL** (critical)

**Issue:** Number() silently returns NaN for non-numeric path segments; NaN is then passed as ruleId to the query, producing undefined behavior.

**Suggestion:** Validate before using: if (!Number.isFinite(ruleId)) { router.push('/automation/rules'); return; } after the Number() call.

frontend/src/app/(app)/automation/rules/history/[rule_id]/page.tsx
```diff
@@ -0,0 +1,168 @@
...
+  trigger_event: string;
+  status: string;
+  executed_at: string;
+  error_message?: string;
+}
+
+function PerRuleHistoryContent() {
+  const params = useParams<{ rule_id: string }>();
+  const parsedRuleId = Number(params.rule_id);
+  const ruleId = Number.isFinite(parsedRuleId) ? parsedRuleId : undefined;
```
Comment (yihua-zhuo) [inline frontend/src/app/(app)/automation/rules/history/[rule_id]/page.tsx:43]: **CRITICAL** (critical)

**Issue:** Number.isFinite check produces undefined ruleId but the component continues rendering with an undefined parameter to useAutomationLogs, sending rule_id=undefined to the API instead of a 404.

**Suggestion:** If ruleId is undefined after parsing, redirect: if (ruleId === undefined) { router.push('/automation/rules'); return; }

frontend/src/app/(app)/automation/rules/history/page.tsx
```diff
@@ -0,0 +1,232 @@
...
+    executed_at: String(l.executed_at ?? ""),
+    error_message: l.error_message ? String(l.error_message) : undefined,
+  }));
+
+  const total = info?.total ?? 0;
+  const pageSize = info?.page_size ?? 20;
+  const totalPages = Math.ceil(total / pageSize);
+
+  function handleView(ruleId: number) {
+    router.push(`/automation/rules/history/${ruleId}`);
```
yihua-zhuo, line 138: **CRITICAL** (critical)

**Issue:** handleView navigates to /automation/rules/history/${id} but that dynamic route does not exist.

**Suggestion:** Create frontend/src/app/(app)/automation/rules/history/[id]/page.tsx or change the router.push target to an existing route.

frontend/src/app/(app)/automation/rules/page.tsx
```diff
@@ -0,0 +1,300 @@
...
+  }));
+
+  function handleToggle(id: number) {
+    toggleRule.mutate(id);
+  }
+  function handleEdit(id: number) {
+    router.push(`/automation/rules/${id}/edit`);
+  }
+  function handleHistory(id: number) {
+    router.push(`/automation/rules/history/${id}`);
```
yihua-zhuo, line 179: **CRITICAL** (critical)

**Issue:** handleEdit navigates to /automation/rules/${id}/edit but no edit page exists at that path.

**Suggestion:** Create frontend/src/app/(app)/automation/rules/[id]/edit/page.tsx or change router.push to an existing route.

frontend/src/app/(app)/automation/rules/page.tsx
```diff
@@ -0,0 +1,300 @@
...
+  }
```
yihua-zhuo, line 180: **CRITICAL** (critical)

**Issue:** handleHistory navigates to /automation/rules/history/${id} but no detail history page exists at that path.

**Suggestion:** Create frontend/src/app/(app)/automation/rules/history/[id]/page.tsx or change router.push to an existing route.

frontend/src/app/(app)/automation/rules/templates/page.tsx
```diff
@@ -0,0 +1,190 @@
...
+          className="rounded-md border border-red-200 bg-red-50 px-4 py-3 text-sm text-red-800"
+        >
+          {String(createRule.error?.message ?? "Failed to activate template. Please try again.")}
+        </div>
+      )}
+
+      <div className="grid grid-cols-1 sm:grid-cols-2 lg:grid-cols-3 gap-4">
+        {TEMPLATES.map((template) => (
+          <TemplateCard
+            key={template.name}
```
yihua-zhuo, line 180: **CRITICAL** (critical)

**Issue:** handleHistory is defined and passed to TemplateCard but never used — history is not rendered on template cards, so dead code.

**Suggestion:** If history per-template is intended, add a View history button to TemplateCard; otherwise remove handleHistory entirely.

yihua-zhuo, line 180: **CRITICAL** (critical)

**Issue:** TemplateCard renders only an Activate button; no way to view execution history for a given template's created rule.

**Suggestion:** Add a 'View history' link/button to TemplateCard or add a dedicated history column to the grid.

frontend/src/components/automation/condition-row.tsx
```diff
@@ -0,0 +1,94 @@
...
+  function updateValue(newValue: string) {
+    onChange({ ...value, value: newValue });
+  }
+
+  return (
+    <div className="flex items-center gap-2">
+      <Input
+        placeholder="field_name"
+        value={value.field}
+        onChange={(e) => updateField(e.target.value.replace(/[^a-zA-Z0-9._-]/g, ""))}
```
yihua-zhuo, line 58: **CRITICAL** (critical)

**Issue:** Silently stripping characters from the field name without telling the user causes confusion and lost input.

**Suggestion:** Remove the `replace` call and instead validate after input or show an error message; the stripping alone provides no injection protection since the value is still used as-is downstream.

frontend/src/components/automation/rule-builder.tsx
```diff
@@ -0,0 +1,272 @@
...
+            <p className="text-sm font-medium">
+              {enabled ? "Active" : "Disabled"}
+            </p>
+            <p className="text-xs text-muted-foreground">
+              {enabled ? "Rule will run automatically when triggered" : "Save as draft without activating"}
+            </p>
+          </div>
+        </div>
+        <div className="flex gap-3">
+          <Button type="submit" onClick={doSubmit} disabled={disabled}>
```
yihua-zhuo, line 260: **CRITICAL** (critical)

**Issue:** Button has both `type="submit"` and `onClick={doSubmit}`, so clicking it calls doSubmit twice — once from the form's onSubmit handler and once directly.

**Suggestion:** Remove `onClick={doSubmit}` from the submit Button; let the form's onSubmit handle it. Keep type="submit" for enter-key support.

tests/unit/test_automation_rules_ui.py
```diff
@@ -0,0 +1,920 @@
...
+
+        svc = AutomationService(session)
+        updated = await svc.update_rule(rule_id=12, tenant_id=1, name="New Name")
+        assert updated.name == "New Name"
+        assert updated.trigger_event == "ticket.created"
+        assert updated.description == "Original description"
+        assert hasattr(updated, "to_dict")
+
+
+class TestAutomationServiceRuleBuilder:
```
yihua-zhuo, line 486: **CRITICAL** (critical)

**Issue:** Duplicate class TestAutomationServiceRuleBuilder starting at line 486 repeats all 13 test methods from the first class (lines 28–483), doubling test runtime and creating 2x the expected test count.

**Suggestion:** Remove the second TestAutomationServiceRuleBuilder class (lines 486–921). Keep only the first class.

frontend/src/app/(app)/automation/rules/history/[rule_id]/page.tsx
```diff
@@ -0,0 +1,168 @@
...
+              </tr>
+            </thead>
+            <tbody>
+              {logs.map((log) => (
+                <tr key={log.id} className="border-b hover:bg-muted/40 transition-colors">
+                  <td className="px-3 py-3">
+                    <span className="text-xs font-mono text-muted-foreground">{log.trigger_event}</span>
+                  </td>
+                  <td className="px-3 py-3">
+                    <Badge variant={(log.status as "success" | "failed" | "skipped") ?? "gray"}>
```
Comment (yihua-zhuo) [inline frontend/src/app/(app)/automation/rules/history/[rule_id]/page.tsx:118]: **WARNING** (suggestion)

**Issue:** Badge variant prop type does not include 'failed' or 'skipped' (badge.tsx only defines default/secondary/destructive/outline/blue/green/gray); this is a TypeScript type mismatch.

**Suggestion:** Add 'failed' and 'skipped' variants to badgeVariants in badge.tsx, or cast the status to 'success' | 'gray' with a fallback.

Comment (yihua-zhuo) [inline frontend/src/app/(app)/automation/rules/history/[rule_id]/page.tsx:118]: **INFO** (suggestion)

**Issue:** STATUS_COLORS (lines 11–15) is defined in module scope but unused — styles are inlined directly rather than referencing it.

**Suggestion:** Remove STATUS_COLORS constant or replace the inline className with STATUS_COLORS[log.status].

frontend/src/app/(app)/automation/rules/templates/page.tsx
```diff
@@ -0,0 +1,190 @@
...
+        </Link>
+      </div>
+      <div>
+        <h1 className="text-2xl font-bold">Rule Templates</h1>
+        <p className="text-sm text-muted-foreground mt-1">
+          Start with a pre-built template and customize it to your needs.
+        </p>
+      </div>
+
+      {createRule.isError && (
```
yihua-zhuo, line 167: **INFO** (suggestion)

**Issue:** createRule.isError is checked but createRule.isSuccess is not used to show a success state after activation.

**Suggestion:** Add a success message or redirect confirmation after a successful template activation.

frontend/src/components/automation/action-row.tsx
```diff
@@ -0,0 +1,232 @@
...
+  type: 'notification.send' | 'ticket.assign' | 'ticket.update_priority' | 'opportunity.add_note' | 'task.create' | 'email.send' | 'webhook.call' | 'tag.add' | string;
```
yihua-zhuo, line 34: **WARNING** (suggestion)

**Issue:** ActionRowValue.type is a union of literals plus a兜底 `string`, making it easy to pass invalid types without runtime errors.

**Suggestion:** Use a strict discriminated-union type or an explicit string enum instead of `| string`.

frontend/src/components/automation/action-row.tsx
```diff
@@ -0,0 +1,232 @@
...
+            />
+            <Input
+              placeholder="subject"
+              value={value.params.subject ?? ""}
+              onChange={(e) => updateParam("subject", e.target.value)}
+              disabled={disabled}
+              className="flex-1"
+            />
+          </div>
+          {value.params.to && !/^[^\s@]+@[^\s@]+\.[^\s@]+$/.test(value.params.to) && (
```
yihua-zhuo, line 141: **INFO** (suggestion)

**Issue:** Email input has no aria-label, making screen-reader announcements unclear when focus lands on it.

**Suggestion:** Add aria-label="recipient email" to the email Input (line 125).

frontend/src/components/automation/action-row.tsx
```diff
@@ -0,0 +1,232 @@
...
+            <Input
+              placeholder="url"
+              type="url"
+              value={value.params.url ?? ""}
+              onChange={(e) => updateParam("url", e.target.value)}
+              disabled={disabled}
+              className="flex-1"
+            />
+          </div>
+          {value.params.url && !isValidUrl(value.params.url) && (
```
yihua-zhuo, line 159: **INFO** (suggestion)

**Issue:** URL input has no aria-label, making screen-reader announcements unclear when focus lands on it.

**Suggestion:** Add aria-label="webhook URL" to the url Input (line 151).

frontend/src/components/automation/rule-builder.tsx
```diff
@@ -0,0 +1,272 @@
...
+      {/* Footer */}
+      <div className="flex flex-col gap-4 pt-2 border-t">
+        <div className="flex items-center gap-3">
+          <button
+            type="button"
+            role="switch"
+            aria-checked={enabled}
+            aria-label="Enable rule"
+            onClick={() => setEnabled((v) => !v)}
+            className={`relative inline-flex h-6 w-11 items-center rounded-full transition-colors focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-ring cursor-pointer ${
```
yihua-zhuo, line 240: **WARNING** (suggestion)

**Issue:** The enabled toggle re-implements a Switch using role="switch" without keyboard handling (Enter/Space), violating keyboard accessibility (WCAG 2.1 §2.1.1).

**Suggestion:** Use @radix-ui/react-switch or @radix-ui/react-toggle instead of a raw button, or add onKeyDown handling for Space/Enter.

frontend/src/components/automation/rule-builder.tsx
```diff
@@ -0,0 +1,272 @@
...
+
+interface RuleBuilderProps {
+  initialValues?: Partial<RuleBuilderValues>;
+  onSubmit: (values: RuleBuilderValues) => Promise<void>;
+  onCancel?: () => void;
+  submitLabel?: string;
+  disabled?: boolean;
+}
+
+function makeCondition(): ConditionRowValue & { id: string } {
```
yihua-zhuo, line 36: **WARNING** (suggestion)

**Issue:** `crypto.randomUUID()` during SSR (Next.js server render) throws in some configurations, risking hydration mismatches.

**Suggestion:** Use React's `useId()` hook for stable IDs, or lazy-initialize via `useState` with a function: `useState(() => crypto.randomUUID())`.

frontend/src/components/automation/rule-builder.tsx
```diff
@@ -0,0 +1,272 @@
...
+  initialValues,
+  onSubmit,
+  onCancel,
+  submitLabel = "Save Rule",
+  disabled = false,
+}: RuleBuilderProps) {
+  const [name, setName] = useState(initialValues?.name ?? "");
+  const [description, setDescription] = useState(initialValues?.description ?? "");
+  const [trigger, setTrigger] = useState(initialValues?.trigger_event ?? "");
+  const [conditions, setConditions] = useState<RuleBuilderCondition[]>(
```
yihua-zhuo, line 54: **WARNING** (suggestion)

**Issue:** No initial condition is added by default (unlike actions which get `makeAction()` on init), so a new rule starts with zero conditions — potentially confusing UX since the conditions section shows 'No conditions'.

**Suggestion:** Either auto-add one condition on mount like actions do, or rename the empty-state message to make the optional nature clearer.

frontend/src/lib/api/queries.ts
```diff
@@ -16,6 +16,9 @@ export const qk = {
   reminders: (upcomingOnly = false) => ["reminders", upcomingOnly] as const,
   activities: (page = 1, type = "") => ["activities", page, type] as const,
   slaBreaches: () => ["sla", "breaches"] as const,
+  automationRules: (page = 1, pageSize = 20) => ["automation_rules", page, pageSize] as const,
```
yihua-zhuo, line 22: **INFO** (suggestion)

**Issue:** qk.reminders uses a boolean parameter (upcomingOnly) as a query key element; booleans can cause TanStack Query cache fragmentation if the default changes, though this is minor.

**Suggestion:** Consider normalizing to string ('all' | 'upcoming') for the query key to avoid false-positive cache misses on type coercion.

tests/unit/test_automation_rules_ui.py
```diff
@@ -0,0 +1,920 @@
...
+            mock_result.scalars = MagicMock(return_value=MagicMock(all=MagicMock(return_value=[])))
+            return mock_result
+
+        session.execute = AsyncMock(side_effect=execute_side_effect)
+
+        svc = AutomationService(session)
+        deleted_id = await svc.delete_rule(rule_id=5, tenant_id=1)
+        assert deleted_id == 5
+        # Verify execute was called (rowcount was consulted via the mock)
+        session.execute.assert_called_once()
```
yihua-zhuo, line 199: **WARNING** (suggestion)

**Issue:** test_delete_rule calls session.execute.assert_called_once() but the test's execute_side_effect always returns a mock result, so execute is always called — the assertion adds no value and is inconsistent with other tests that omit it.

**Suggestion:** Remove 'session.execute.assert_called_once()' from line 199, consistent with the rest of the test class.

tests/unit/test_automation_rules_ui.py
```diff
@@ -0,0 +1,920 @@
...
+        session.execute = AsyncMock(side_effect=execute_side_effect)
+
+        svc = AutomationService(session)
+        items, total = await svc.list_logs(tenant_id=1, page=1, page_size=20, rule_id=7)
+        assert isinstance(items, list)
+        assert len(items) == 1
+        assert items[0].rule_id == 7
+        assert total == 1
+
+    async def test_list_logs_with_status_filter(self, mock_db_session):
```
yihua-zhuo, line 291: **WARNING** (suggestion)

**Issue:** test_list_logs_with_status_filter has a side-effect assertion inside execute_side_effect that causes the test to raise AssertionError if the SQL text does not contain 'status' or 'failed' — this is a test-design bug (assertions belong in test bodies, not side-effects) and the assertion is redundant since the real service query will always contain the filter text.

**Suggestion:** Move the assertion out of execute_side_effect into the test body; or remove it since the service is the source of truth for SQL generation.

tests/unit/test_automation_rules_ui.py
```diff
@@ -0,0 +1,920 @@
...
+            actions=[{"type": "notification.send", "params": {"message": "Alert!"}}],
+            enabled=True,
+            created_by=1,
+        )
+        # verify rule returned correctly
+        assert hasattr(rule, "id")
+        assert hasattr(rule, "to_dict")
+        assert rule.conditions == [{"field": "priority", "operator": "eq", "value": "high"}]
+
+    async def test_list_rules_respects_trigger_event_filter(self, mock_db_session):
```
yihua-zhuo, line 405: **WARNING** (suggestion)

**Issue:** test_list_rules_respects_trigger_event_filter does not verify the filter was actually applied to the query — it only checks that any call captured the trigger_event value, which does not confirm the SQL WHERE clause contains the filter.

**Suggestion:** Strengthen the assertion to parse the SQL string and confirm 'ticket.created' or the filter parameter is present in a WHERE clause, not just in any logged call.

tests/unit/test_automation_rules_ui.py
```diff
@@ -0,0 +1,920 @@
...
+            page=1,
+            page_size=20,
+            trigger_event="ticket.created",
+        )
+        assert isinstance(items, list)
+        assert isinstance(total, int)
+        # At least one call should have included the trigger_event filter
+        assert any(called_with_trigger)
+
+    async def test_list_rules_respects_enabled_filter(self, mock_db_session):
```
yihua-zhuo, line 432: **WARNING** (suggestion)

**Issue:** test_list_rules_respects_enabled_filter has a weak assertion — it only checks that enabled=True was passed as a kwarg, but never confirms the SQL query actually contained the enabled filter, making the test pass trivially regardless of whether the service applies the filter.

**Suggestion:** Add a SQL string parse assertion to verify the SQL contains the enabled filter, not just that the kwarg was passed.

tests/unit/test_automation_rules_ui.py
```diff
@@ -0,0 +1,920 @@
...
+        session.execute = AsyncMock(side_effect=execute_side_effect)
+
+        svc = AutomationService(session)
+        items, total = await svc.list_logs(tenant_id=1, page=1, page_size=20)
+        assert isinstance(items, list)
+        assert len(items) == 2
+        assert isinstance(total, int)
+        assert total == 2
+
+    async def test_list_logs_with_rule_id_filter(self, mock_db_session):
```
yihua-zhuo, line 263: **INFO** (suggestion)

**Issue:** test_list_logs_with_rule_id_filter appends string conversions to logged_rule_ids but never asserts on the accumulated list — the test passes even if the filter is never passed to the SQL.

**Suggestion:** Add an assertion at the end of the test (e.g., 'assert any(logged_rule_ids)') or remove the logging logic.

frontend/src/app/(app)/automation/rules/new/page.tsx
```diff
@@ -0,0 +1,53 @@
...
+import type { RuleBuilderValues } from "@/components/automation/rule-builder";
+import { useCreateAutomationRule } from "@/lib/api/queries";
+import { ArrowLeft } from "lucide-react";
+import Link from "next/link";
+
+export default function NewRulePage() {
+  const router = useRouter();
+  const createRule = useCreateAutomationRule();
+
+  async function handleSubmit(values: RuleBuilderValues) {
```
yihua-zhuo, line 13: **CRITICAL** (critical)

**Issue:** handleSubmit silently swallows all errors — if createRule.mutateAsync throws (network failure, 422, 500), the user sees no feedback and the form appears to do nothing.

**Suggestion:** Wrap in try/catch, call createRule.error reset on start, and display createRule.error.message or a toast notification on failure.

frontend/src/app/(app)/automation/rules/page.tsx
```diff
@@ -0,0 +1,300 @@
...
+  onEdit,
+  onHistory,
+  onDelete,
+}: {
+  rule: RuleRow;
+  onToggle: (id: number) => void;
+  onEdit: (id: number) => void;
+  onHistory: (id: number) => void;
+  onDelete: (id: number) => Promise<void>;
+}) {
```
yihua-zhuo, line 80: **CRITICAL** (critical)

**Issue:** handleDeleteClick uses window.confirm for a destructive action — this is a browser native dialog with no customization, and the Promise<void> return type of onDelete is ignored.

**Suggestion:** Replace window.confirm with a proper confirmation dialog (e.g., a custom modal or AlertDialog component) and handle the async result safely.

frontend/src/app/(app)/automation/rules/templates/page.tsx
```diff
@@ -0,0 +1,190 @@
...
+        </div>
+      </CardContent>
+    </Card>
+  );
+}
+
+export default function TemplatesPage() {
+  const router = useRouter();
+  const createRule = useCreateAutomationRule();
+  const [activatingTemplate, setActivatingTemplate] = useState<string | null>(null);
```
yihua-zhuo, line 132: **CRITICAL** (critical)

**Issue:** `useState` is used on line 132 but is not imported — runtime will throw a ReferenceError.

**Suggestion:** Add `import { useState } from "react";` to the imports at the top of the file.

tests/unit/test_automation_rules_ui.py
```diff
@@ -0,0 +1,920 @@
...
+        assert updated.name == "New Name"
+        assert updated.trigger_event == "ticket.created"
+        assert updated.description == "Original description"
+        assert hasattr(updated, "to_dict")
+
+
+class TestAutomationServiceRuleBuilder:
+    """Service-layer rule CRUD — tested via mock DB session."""
+
+    async def test_create_rule_returns_orm_object(self, mock_db_session):
```
yihua-zhuo, line 489: **CRITICAL** (critical)

**Issue:** Tests in the second `TestAutomationServiceRuleBuilder` class silently replace identically-named tests from the first class. `test_list_logs_with_rule_id_filter` (line 719) and `test_list_logs_with_status_filter` (line 738) lose all meaningful assertions — they only assert `isinstance(items, list)` and `isinstance(total, int)` with no behavioral verification, unlike the first-class versions which checked `len(items)==1`, `items[0].rule_id==7`, `items[0].status=='failed'`.

**Suggestion:** Remove the second class entirely and keep only the first class's more thorough test versions. Alternatively, merge only the truly unique tests (e.g., any new scenarios) into the first class.

frontend/src/app/(app)/automation/rules/[id]/edit/page.tsx
```diff
@@ -0,0 +1,150 @@
+"use client";
+import { useState } from "react";
```
Comment (yihua-zhuo) [inline frontend/src/app/(app)/automation/rules/[id]/edit/page.tsx:2]: **INFO** (suggestion)

**Issue:** useRouter is imported but never used — all navigation is done via Link components and router.push in the submit handler.

**Suggestion:** Remove the unused useRouter import from next/navigation.

frontend/src/app/(app)/automation/rules/[id]/page.tsx
Comment (yihua-zhuo) [inline frontend/src/app/(app)/automation/rules/[id]/page.tsx:2]: **INFO** (suggestion)

**Issue:** useRouter is imported but never used — all navigation is done via Link components.

**Suggestion:** Remove the unused useRouter import from next/navigation.

frontend/src/app/(app)/automation/rules/[id]/page.tsx
```diff
@@ -0,0 +1,161 @@
...
+        {actions.length === 0 ? (
+          <p className="text-sm text-muted-foreground">No actions configured.</p>
+        ) : (
+          <div className="space-y-2">
+            {actions.map((action, i) => (
+              <div key={i} className="rounded-md border bg-muted/20 p-3">
+                <span className="font-medium text-sm">{action.type}</span>
+                {Object.keys(action.params ?? {}).length > 0 && (
+                  <div className="mt-1 text-xs text-muted-foreground font-mono">
+                    {Object.entries(action.params).map(([k, v]) => `${k}: ${v}`).join(", ")}
```
Comment (yihua-zhuo) [inline frontend/src/app/(app)/automation/rules/[id]/page.tsx:151]: **WARNING** (suggestion)

**Issue:** Object.entries(action.params) is called without confirming params is an object; if params is a non-object primitive the call silently produces an empty result rather than surfacing the mismatch.

**Suggestion:** Add a guard: if (typeof action.params !== 'object' || action.params === null) return null; before Object.entries, or assert the type via a type guard.

frontend/src/app/(app)/automation/rules/[id]/page.tsx
```diff
@@ -0,0 +1,161 @@
...
+
+      {/* Status & trigger metadata */}
+      <div className="flex flex-wrap gap-4 items-center">
+        <div className="flex items-center gap-2">
+          <span className="text-sm text-muted-foreground">Trigger:</span>
+          <Badge variant="blue">{triggerLabel}</Badge>
+        </div>
+        <div className="flex items-center gap-2">
+          <span className="text-sm text-muted-foreground">Status:</span>
+          <Badge variant={rule.enabled ? "green" : "gray"}>
```
Comment (yihua-zhuo) [inline frontend/src/app/(app)/automation/rules/[id]/page.tsx:111]: **WARNING** (suggestion)

**Issue:** Badge variant 'blue' may not be a valid variant in the project's Badge component — shadcn/ui's default Badge only supports 'default' and 'secondary', and invalid variants are silently ignored.

**Suggestion:** Use a variant the Badge component actually supports (e.g., variant='default' or a custom 'blue' class), or add 'blue' to the Badge's variant union if intended.

frontend/src/app/(app)/automation/rules/history/[rule_id]/page.tsx
```diff
@@ -0,0 +1,168 @@
...
+import { useState, Suspense } from "react";
+import { useParams } from "next/navigation";
+import { useAutomationLogs } from "@/lib/api/queries";
+import { Button } from "@/components/ui/button";
+import { Badge } from "@/components/ui/badge";
+import { AuthGuard } from "@/lib/components/auth-guard";
+import { ArrowLeft } from "lucide-react";
+import Link from "next/link";
+
+const STATUS_COLORS: Record<string, string> = {
```
Comment (yihua-zhuo) [inline frontend/src/app/(app)/automation/rules/history/[rule_id]/page.tsx:11]: **WARNING** (suggestion)

**Issue:** STATUS_COLORS is defined but never used — dead code.

**Suggestion:** Remove lines 11-15 or remove the unused constant.

frontend/src/app/(app)/automation/rules/history/[rule_id]/page.tsx
```diff
@@ -0,0 +1,168 @@
...
+}
+
+function PerRuleHistoryContent() {
+  const params = useParams<{ rule_id: string }>();
+  const parsedRuleId = Number(params.rule_id);
+  const ruleId = Number.isFinite(parsedRuleId) ? parsedRuleId : undefined;
+
+  const [page, setPage] = useState(1);
+
+  const { data, isLoading, isError } = useAutomationLogs(page, 20, ruleId);
```
Comment (yihua-zhuo) [inline frontend/src/app/(app)/automation/rules/history/[rule_id]/page.tsx:47]: **INFO** (suggestion)

**Issue:** Invalid rule IDs in the URL (e.g. /automation/rules/history/xyz) result in silent fallback to all-logs view with no user feedback.

**Suggestion:** Add a check: if ruleId is undefined due to a non-numeric segment, show a 'Invalid rule ID' message or redirect.

frontend/src/app/(app)/automation/rules/history/page.tsx
```diff
@@ -0,0 +1,232 @@
...
+  id: number;
+  rule_id: number;
+  rule_name: string;
+  trigger_event: string;
+  status: string;
+  executed_at: string;
+  error_message?: string;
+}
+
+function LogsTable({ logs, onView }: { logs: LogRow[]; onView: (ruleId: number) => void }) {
```
yihua-zhuo, line 46: **WARNING** (suggestion)

**Issue:** LogsTable component accepts onView that calls router.push, but the table is not testable in isolation without mocking router — tight coupling to global router.

**Suggestion:** Pass a navigation callback or href instead of a callback that embeds router logic, reducing coupling.

frontend/src/app/(app)/automation/rules/page.tsx
```diff
@@ -0,0 +1,300 @@
...
+  function handleToggle(id: number) {
+    toggleRule.mutate(id);
+  }
+  function handleEdit(id: number) {
+    router.push(`/automation/rules/${id}/edit`);
+  }
+  function handleHistory(id: number) {
+    router.push(`/automation/rules/history/${id}`);
+  }
+  async function handleDelete(id: number) {
```
yihua-zhuo, line 181: **INFO** (suggestion)

**Issue:** handleDelete does not handle errors — if deleteRule.mutateAsync throws, the exception propagates silently with no user feedback.

**Suggestion:** Wrap in try/catch and show an error toast or inline message on failure.

frontend/src/components/automation/condition-row.tsx
```diff
@@ -0,0 +1,94 @@
...
+import {
+  Select,
+  SelectContent,
+  SelectItem,
+  SelectTrigger,
+  SelectValue,
+} from "@/components/ui/select";
+import { X } from "lucide-react";
+
+const OPERATORS = [
```
yihua-zhuo, line 13: **INFO** (suggestion)

**Issue:** Empty conditions (field = "") pass the `field` key through to the API with no server-side enforcement that prevents empty-field conditions from being saved.

**Suggestion:** RuleBuilder filters empty conditions on submit (field.trim() !== ""), but ConditionRow itself accepts empty field values without any visual or client-side warning. Consider adding a subtle validation indicator when field is blank.

frontend/src/components/automation/rule-builder.tsx
```diff
@@ -0,0 +1,272 @@
...
+    }
+
+    const payload: RuleBuilderValues = {
+      name: name.trim(),
+      description: description.trim(),
+      trigger_event: trigger,
+      conditions: conditions.filter((c) => c.field.trim() !== ""),
+      actions: filledActions,
+      enabled,
+    };
```
yihua-zhuo, line 115: **INFO** (suggestion)

**Issue:** The form's `onSubmit` fires even when `trigger` is empty — the validation error is set but the form still submits (browser fires `submit` again).

**Suggestion:** Call `e.preventDefault()` in `handleSubmit` before `await doSubmit()` so a failed validation does not re-trigger the form lifecycle.

frontend/src/lib/api/queries.ts
```diff
@@ -19,6 +19,9 @@ export const qk = {
...
+  automationRule: (id: number) => ["automation_rules", "detail", id] as const,
+  automationLogs: (page = 1, pageSize = 20, ruleId?: number, status?: string) => ["automation_logs", page, pageSize, ruleId ?? null, status ?? ""] as const,
 } as const;
 
 interface PaginatedResponse<T> {
```
yihua-zhuo, line 27: **WARNING** (suggestion)

**Issue:** ApiEnvelope and PaginatedResponse are defined but unused — all queryFn return types are manually typed as ApiEnvelope<Record<string, unknown>>, losing generic type safety.

**Suggestion:** Use a generic wrapper or define per-domain typed response interfaces to avoid casting to Record<string, unknown> for every query.

tests/integration/test_automation_rules_ui_integration.py
```diff
@@ -0,0 +1,332 @@
...
+        get_resp = await api_client.get(f"/api/v1/automation/rules/{rule_id}")
+        assert get_resp.status_code == 200
+        get_data = get_resp.json()
+        assert get_data["success"] is True
+        assert get_data["data"]["id"] == rule_id
+        assert get_data["data"]["name"] == "Test Rule"
+        assert len(get_data["data"]["actions"]) == 1
+        assert get_data["data"]["actions"][0]["type"] == "notification.send"
+
+    async def test_list_rules(self, api_client, tenant_id_web):
```
yihua-zhuo, line 50: **INFO** (suggestion)

**Issue:** test_list_rules creates one rule and checks it appears in the list, but does not verify other tenants' rules are excluded — cross-tenant isolation is not asserted.

**Suggestion:** Add assertion checking that no rules from a different tenant appear in the response, or seed rules for a second tenant and confirm only tenant_id_web's rules are returned.

tests/integration/test_automation_rules_ui_integration.py
```diff
@@ -0,0 +1,332 @@
...
+        assert logs_resp.status_code == 200
+        logs_data = logs_resp.json()
+        assert logs_data["success"] is True
+        assert len(logs_data["data"]["items"]) >= 1
+        log_entry = logs_data["data"]["items"][0]
+        assert log_entry["rule_id"] == rule_id
+        assert log_entry["status"] == "success"
+        assert len(log_entry["actions_executed"]) == 1
+
+    async def test_list_logs(self, api_client, tenant_id_web):
```
yihua-zhuo, line 193: **INFO** (suggestion)

**Issue:** test_list_logs reads logs without verifying cross-tenant isolation — a rule created by one test in a prior test could leak into this test's assertion.

**Suggestion:** Assert that returned logs belong to tenant_id_web or check that total count is consistent.

tests/integration/test_automation_rules_ui_integration.py
```diff
@@ -0,0 +1,332 @@
...
+                "name": "No Actions Rule",
+                "trigger_event": "ticket.created",
+                "conditions": [],
+                "actions": [],
+                "enabled": True,
+            },
+        )
+        assert response.status_code == 422
+
+    async def test_pagination_params(self, api_client, tenant_id_web):
```
yihua-zhuo, line 326: **INFO** (suggestion)

**Issue:** test_pagination_params only checks that page and page_size are echoed back — it does not verify that pagination actually affects the items returned (e.g. page_size=5 with 10 rules returns exactly 5 items).

**Suggestion:** Seed multiple rules and assert that `len(items) == 5` or `len(items) <= 5` for the first page.

tests/integration/test_automation_rules_ui_integration.py
```diff
@@ -0,0 +1,332 @@
...
+        assert data["success"] is True
+
+        # Verify an AutomationLogModel row was persisted
+        logs_resp = await api_client.get(f"/api/v1/automation/logs?rule_id={rule_id}")
+        assert logs_resp.status_code == 200
+        logs_data = logs_resp.json()
+        assert logs_data["success"] is True
+        assert len(logs_data["data"]["items"]) >= 1
+        log_entry = logs_data["data"]["items"][0]
+        assert log_entry["rule_id"] == rule_id
```
yihua-zhuo, line 189: **INFO** (suggestion)

**Issue:** Assertion `len(log_entry['actions_executed']) == 1` checks the action count but not the actual content or type of the executed action.

**Suggestion:** Assert `log_entry['actions_executed'][0]['type'] == 'notification.send'` to verify the right action fired.

tests/unit/conftest.py
```diff
@@ -954,6 +961,81 @@ def handler(sql_text, params):
...
+        if ("select" in sql_text and "from automation_rules" in sql_text
+                and "where id" in sql_text and "count" not in sql_text):
+            rid = params.get("id")
+            if rid in state.automation_rules:
+                row = state.automation_rules[rid]
+                return MockResult([MockRow(row.copy())])
+
+        # SELECT automation_rules list (no id filter) — count first, then data
+        if ("select" in sql_text and "from automation_rules" in sql_text
+                and "count" not in sql_text and "order_by" not in sql_text):
```
yihua-zhuo, line 1000: **INFO** (suggestion)

**Issue:** In make_automation_handler, list-query detection checks `'order_by' not in sql_text`, but SQLAlchemy 2.x typically generates `ORDER BY` (all-caps with a space) — the check may not fire when ORDER BY is present.

**Suggestion:** Check for `'order by' in sql_text` (lowercase, as the SQL is lowercased on line 1096) rather than 'order_by' to match the actual string in generated SQL.

tests/unit/test_automation_rules_ui.py
```diff
@@ -0,0 +1,920 @@
...
+        session = mock_db_session
+
+        mock_failed = MagicMock(id=2, rule_id=7, status="failed")
+
+        def execute_side_effect(sql, **kwargs):
+            mock_result = MagicMock()
+            mock_result.scalar_one = MagicMock(return_value=1)
+            sql_str = str(sql).lower() if hasattr(sql, "__str__") else ""
+            # Verify status filter is applied
+            assert "status" in sql_str or "failed" in sql_str
```
yihua-zhuo, line 302: **WARNING** (suggestion)

**Issue:** Assert statement is inside the `execute_side_effect` function (a mock side effect), not in test flow. If `sql_str` does not contain expected text, the assertion fires during query execution rather than as a clear test failure, and the test still passes if no assertion fires.

**Suggestion:** Remove the in-side-effect assertion. Store SQL text in a list and assert on it after the call: `assert 'status' in sql_calls[0] or 'failed' in sql_calls[0]`.

PR #413, opened 2026-05-14
Auto: implement issue #113

### LINKED ISSUE #113: [Backend] SLA Summary Counts API — missing endpoint for dashboard totals
## Feature Description

The frontend SLA page (#66) shows 3 stat cards (Breached / At Risk / On Track) that compute counts from the current page of tickets only. Since `useTickets` is paginated, the totals change when paging and don't reflect the full dataset.

### `GET /api/v1/sla/summary`

Return aggregated SLA counts from the full dataset (not paginated).

Response shape:
```json
{
  "success": true,
  "data": {
    "breached": 3,
    "at_risk": 7,
    "on_track": 24,
    "total_tickets": 34
  }
}
```

- `breached` = tickets with `response_deadline < now`
- `at_risk` = open tickets with `response_deadline` within 4 hours from now
- `on_track` = open tickets with `response_deadline` > 4 hours from now (or resolved/closed)

Counts should be computed server-side from the full dataset, not from a single page.

## Acceptance Criteria

1. Returns accurate counts from full ticket dataset
2. All tickets are tenant-isolated
3. OpenAPI spec updated with `/api/v1/sla/summary`

## Frontend Reference

`frontend/src/app/(app)/sla/page.tsx` — currently computes `atRisk` and `onTrack` from `useTickets(page, "open")` which only returns one page. Replace with `useSlaSummary()` hook once this endpoint exists.

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

yihua-zhuo:
Closes #113

Plan: `.plans/issue-113.md`

1. **Plan** — Claude wrote a detailed implementation plan to `.plans/issue-113.md`.

Please review before merging.

## Summary by CodeRabbit

* **New Features**
  * Added GET /api/v1/sla/summary returning tenant-scoped SLA stats: breached, at-risk, on-track, and total tickets.

* **Documentation**
  * Added an implementation plan detailing expected behavior, acceptance criteria, and test scenarios for SLA summary.

* **Tests**
  * Added comprehensive unit and integration tests covering classification, tenant isolation, boundary behavior, zero/defaults, and endpoint responses.
  * Added test helpers and mocks to simulate SLA query results and seed integration fixtures.

[Review Change Stack](https://app.coderabbit.ai/change-stack/yihua-zhuo/agent-job/pull/413?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


.plans/issue-113.md
```diff
@@ -0,0 +1,69 @@
...
+   - Add `GET /sla/summary` to `tickets_router` (in the existing `# SLA endpoints` section alongside `check_sla_status` and `get_sla_breach_tickets`):
+     ```
+     @tickets_router.get("/sla/summary")
+     async def get_sla_summary(
+         ctx: AuthContext = Depends(require_auth),
+         session: AsyncSession = Depends(get_db),
+     ):
+         """Return aggregated SLA counts across the full ticket dataset for the current tenant."""
+         sla_svc = SLAService(session)
+         counts = await sla_svc.get_sla_summary(tenant_id=ctx.tenant_id or 0)
```
yihua-zhuo, line 39: **WARNING** (suggestion)

**Issue:** Plan shows `tenant_id=ctx.tenant_id or 0` (line 39) — passing a fallback tenant_id of 0 when auth token is missing will silently return data for tenant 0 instead of rejecting the unauthenticated request.

**Suggestion:** Use `ctx.tenant_id` directly without fallback; if auth middleware guarantees tenant_id is set, remove `or 0` from all endpoints in tickets.py.

src/api/routers/tickets.py
```diff
@@ -85,6 +85,15 @@ class TicketBulkUpdate(BaseModel):
     status: str | None = Field(None, pattern="^(open|in_progress|pending|resolved|closed)$")
 
 
+class SLAStatCard(BaseModel):
```
yihua-zhuo, line 88: **INFO** (suggestion)

**Issue:** SLAStatCard is defined but never imported or used as a dependency — the return type annotation on get_sla_summary is missing.

**Suggestion:** Add `-> SLAStatCard` return type hint to `get_sla_summary` and consider using it as FastAPI's response_model for automatic OpenAPI documentation.

src/services/sla_service.py
```diff
@@ -39,6 +41,57 @@ def check_sla_status(self, ticket: Ticket) -> Literal["normal", "warning", "brea
...
+        on_track_count = (total_count or 0) - (breached_count or 0) - (at_risk_count or 0)
+
+        return {
+            "breached": breached_count or 0,
+            "at_risk": at_risk_count or 0,
+            "on_track": on_track_count,
+            "total_tickets": total_count or 0,
+        }
+
     async def get_breach_tickets(self, tenant_id: int = 0, tickets: list[Ticket] = None) -> list[Ticket]:
```
yihua-zhuo, line 103: **WARNING** (suggestion)

**Issue:** Rule 45 / 46: `get_breach_tickets` silently returns [] when both `tickets` is None and `self._ticket_service` is None, with no log message or detectable signal — a caller mistakenly omitting both arguments gets zero tickets with no indication of the cause.

**Suggestion:** Add a guard: if tickets is None and self._ticket_service is None: raise ValueError('ticket_service must be provided or tickets passed directly').

tests/integration/test_sla_integration.py
```diff
@@ -0,0 +1,183 @@
...
+
+    assert result["breached"] == 0
+    assert result["at_risk"] == 0
+    assert result["on_track"] == 0
+    assert result["total_tickets"] == 0
+
+
+async def test_sla_summary_tenant_isolation(db_schema, tenant_id, tenant_id_2, async_session):
+    """Seed tickets for two different tenant_ids; each query only sees its own."""
+    from datetime import UTC
```
yihua-zhuo: **INFO** (suggestion)

**Issue:** Rule 17: `from datetime import UTC` is imported at line 105 but `datetime` is already imported at line 2, making line 105 a dead import that adds noise.

**Suggestion:** Remove line 105: `from datetime import UTC` — use `datetime` already imported at line 2.

tests/unit/test_sla_service.py
```diff
@@ -0,0 +1,198 @@
...
+
+import pytest
+
+# Ensure src/ on path
+_project_root = Path(__file__).resolve().parents[2]
+_src_root = _project_root / "src"
+if str(_src_root) not in sys.path:
+    sys.path.insert(0, str(_src_root))
+
+from tests.unit.conftest import MockResult
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** MockResult is imported but never used in the file.

**Suggestion:** Remove the unused import: `from tests.unit.conftest import MockResult`

tests/unit/test_sla_service.py
```diff
@@ -0,0 +1,198 @@
+"""Unit tests for SLAService.get_sla_summary."""
+from __future__ import annotations
+
+import sys
+from pathlib import Path
+
+import pytest
+
+# Ensure src/ on path
```
yihua-zhuo, line 11: **INFO** (suggestion)

**Issue:** Lines 9–13 duplicate the `sys.path` manipulation that conftest.py already performs at lines 32–35.

**Suggestion:** Remove lines 9–13; rely on conftest.py to set PYTHONPATH once, or move path setup entirely to conftest.py and remove it from this file.

tests/unit/test_sla_service.py
```diff
@@ -0,0 +1,198 @@
...
+    return _SlaMockSession([handler])
+
+
+# ---------------------------------------------------------------------------
+# Tests
+# ---------------------------------------------------------------------------
+
+
+class TestGetSlaSummary:
+    async def test_all_categories_populated(self, mock_db_session_all_populated):
```
yihua-zhuo, line 84: **INFO** (suggestion)

**Issue:** Module-level imports are repeated inside each of the 6 test methods instead of being at module level.

**Suggestion:** Move `from services.sla_service import SLAService` to the top-level imports section (alongside the existing `from tests.unit.conftest import MockResult` line).

yihua-zhuo:

**Status:** `fail`
**Inline comments posted:** 16

**Summary:** Reviewed 9 file(s) between origin/master and origin/auto/issue-113.

- `tests/unit/domain_handlers/sla.py`: Module lacks a get_handlers(state) function, so the conftest auto-loader (which calls get_handlers on every discovered domain_handlers module) will never register sla.py's handlers automatically.
- `tests/integration/conftest.py`: No issues. Integration tests correctly use real DB (no mocks), tenant fixtures are well-designed, and per-file cleanup via _cleanup_after_module is properly scoped.
- `tests/integration/test_sla_integration.py`: No integration tests exist for get_breach_tickets, leaving a critical service method unverified against real Postgres.
- `tests/unit/test_tickets_router.py`: No test exercises ticket assignment to a different tenant (cross-tenant negative test) — rule 126 flags missing tenant isolation coverage.
- `tests/unit/test_tickets_router.py`: No test verifies that a service call was made with the correct tenant_id for any endpoint except /sla/summary (rule 126 tenant verification).

`tests/unit/conftest.py` L269: 
**INFO** (suggestion)

**Issue:** Bare 'except Exception' with pass body (BLE001 suppressed) in _execute_side_effect; this silently swallows compilation errors that might indicate SQL construction bugs.

**Suggestion:** Consider logging the exception or raising for non-expected failures so broken SQL fails loudly in tests rather than silently falling back to raw string matching. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`tests/unit/test_tickets_router.py` L138: 
**INFO** (suggestion)

**Issue:** get_db override returns a plain MagicMock() instead of an AsyncSession, but the mock service bypasses the DB entirely — this is misleading about the fixture's intent.

**Suggestion:** Use AsyncMock(spec=AsyncSession) or remove the override since the mocked service never touches the session. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`tests/unit/test_tickets_router.py` L120: 
**INFO** (suggestion)

**Issue:** mock_user_service is set up in the fixture but never asserted, verified, or used by any test — unused dead setup.

**Suggestion:** Remove mock_user_service from the fixture, or add a test that exercises UserService (e.g. an endpoint that resolves user identity). — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`tests/unit/test_tickets_router.py` L348: 
**WARNING** (suggestion)

**Issue:** test_empty_content_rejected only asserts status_code==422; it does not verify the error message, error code, or response shape — weak assertion per rule 144.

**Suggestion:** Add: assert resp.json()['message'] == '...'; assert 'content' in resp.json().get('errors', {}) — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`tests/unit/test_tickets_router.py` L157: 
**INFO** (suggestion)

**Issue:** test names in TestCreateTicketEndpoint do not use pytest convention (e.g., test_success_returns_201 could be more descriptive about the case). Rule 67 requires test names to clearly explain intent; current names are generic.

**Suggestion:** Rename to test_create_ticket_success_201_subject_and_description_returned, test_create_ticket_validation_error_4xx_customer_not_found, etc. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_

tests/unit/test_sla_service.py
```diff
@@ -0,0 +1,198 @@
...
+    async def __aexit__(self, *args):
+        pass
+
+
+# ---------------------------------------------------------------------------
+# SQL handler for SLA summary COUNT queries
+# ---------------------------------------------------------------------------
+
+
+def make_sla_summary_handler(breached: int, at_risk: int, total: int):
```
yihua-zhuo: **CRITICAL** (critical)

**Issue:** make_sla_summary_handler uses fragile string containment ('>' in sql_text, '<' in sql_text) instead of inspecting bind parameters, so it cannot validate that tenant_id is actually passed (Rule 133).

**Suggestion:** Inspect params['tenant_id'] explicitly, and match on structural SQLAlchemy elements (e.g., inspect clause element types or column names) rather than substring checks.

tests/unit/test_sla_service.py
```diff
@@ -0,0 +1,198 @@
...
+    """Return a handler that responds to select count(...) from tickets queries.
+
+    The handler detects which filter is active based on the SQL text and returns
+    the appropriate count.
+    """
+
+    def handler(sql_text: str, params: dict):
```
yihua-zhuo: **CRITICAL** (critical)

**Issue:** Mock handler returns the same counts for any tenant_id because it never reads params['tenant_id'], so tests test_tenant_1_isolation and test_tenant_2_isolation do not actually verify tenant isolation in the SQL layer (Rules 126, 134).

**Suggestion:** Add a params dict inspection: if params.get('tenant_id') != expected_tenant, return None so that a missing or wrong tenant_id bind causes the query to return None (unhandled). This forces the mock to actually validate tenant scope.

.plans/issue-113.md
```diff
@@ -0,0 +1,69 @@
...
+         return {"success": True, "data": counts}
+     ```
```
yihua-zhuo, line 41: **INFO** (suggestion)

**Issue:** Plan shows `tenant_id=ctx.tenant_id or 0` but implementation uses `tenant_id=ctx.tenant_id` without the `or 0` fallback.

**Suggestion:** Align plan snippet with actual router code, or vice versa. The `or 0` pattern is used consistently across all other endpoints in tickets.py.

tests/unit/test_sla_service.py
```diff
@@ -0,0 +1,198 @@
...
+        if not ("select" in sql_text and "count" in sql_text and "from tickets" in sql_text):
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** Handler uses '>' in sql_text as the at_risk signal. This is imprecise — '>' appears in many contexts (operators, table aliases, column names). Any count query whose SQL text happens to contain '>' (e.g. from a JOIN, CTE, or comment) would be misrouted.

**Suggestion:** Match on the actual SLA column name (e.g. 'response_deadline') and comparison operator near it, or inspect the SQLAlchemy clause tree structure rather than doing a raw substring scan.

tests/unit/test_sla_service.py
```diff
@@ -0,0 +1,198 @@
...
+            return None
+
+        # Check for ">" first — at_risk SQL has both "> now" and "<= now+4h" but no plain "<"
+        # Breached SQL has only "< now" with no ">" comparison
+        if ">" in sql_text:
+            return at_risk
+
+        # breached: resolved_at IS NULL AND response_deadline IS NOT NULL AND response_deadline < now
+        if "<" in sql_text:
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** Handler uses '<' in sql_text as the breached signal, but '< now' could appear in comparisons on other columns (e.g., resolved_at < updated_at). If any count query in get_sla_summary includes a '<' on a non-SLA column, it will be misclassified.

**Suggestion:** Narrow the match to the specific SLA deadline column name and timestamp comparison pattern.

tests/unit/test_sla_service.py
```diff
@@ -0,0 +1,198 @@
...
+    the appropriate count.
+    """
+
+    def handler(sql_text: str, params: dict):
+        if not ("select" in sql_text and "count" in sql_text and "from tickets" in sql_text):
+            return None
+
+        # Check for ">" first — at_risk SQL has both "> now" and "<= now+4h" but no plain "<"
+        # Breached SQL has only "< now" with no ">" comparison
+        if ">" in sql_text:
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** The at_risk / breached / total branching order is fragile: at_risk SQL contains both '>' and '<', so it is caught by the '>' branch — but this depends entirely on the SQL text never containing '>' in a non-at_risk context. The ordering is an implicit assumption not enforced anywhere.

**Suggestion:** Extract a canonical column name or SQLAlchemy type from the query and route on that, so the branching logic is explicit and reviewable.

tests/unit/test_sla_service.py
```diff
@@ -0,0 +1,288 @@
...
+
+        svc = SLAService(mock_db_session_only_on_track)
+        result = await svc.get_sla_summary(tenant_id=1)
+
+        assert result["breached"] == 0
+        assert result["at_risk"] == 0
+        assert result["on_track"] == 4  # all tickets are on-track
+        assert result["total_tickets"] == 4
+
+    async def test_tenant_1_isolation(self, mock_db_session_tenant_1):
```
yihua-zhuo, line 128: **INFO** (suggestion)

**Issue:** test_tenant_1_isolation and test_tenant_2_isolation only assert the return values match the seeded counts — they do not assert that the handler received and validated the correct tenant_id in the bound params, so a bug where the service ignores tenant_id entirely would pass these tests.

**Suggestion:** Add handler-side assertions to verify params['tenant_id'] equals the expected value for each test case, ensuring the SQL bind actually carries the tenant scope.

tests/unit/test_sla_service.py
```diff
@@ -0,0 +1,288 @@
...
+
+    get_sla_summary calls session.scalar(select(func.count(...))) directly,
+    not session.execute(), so we need to support that path.
+    """
+
+    def __init__(self, handlers):
+        self._handlers = handlers
+
+    async def scalar(self, sql, params=None):
+        sql_text = str(sql).lower().strip()
```
yihua-zhuo: **INFO** (suggestion)

**Issue:** CountingSession.scalar method is declared `async def` but calls `await counting_scalar(...)` where counting_scalar is itself async — the outer await is unnecessary since scalar is not awaited via session.scalar() in Python's async protocol and adds noise.

**Suggestion:** Remove `await` inside CountingSession.scalar (make counting_scalar synchronous) or remove the outer async/await wrapper if not needed.

tests/unit/test_sla_service.py
```diff
@@ -0,0 +1,292 @@
...
+    Detects which bucket a query targets by inspecting the SQL clause structure:
+
+      - total    : no "response_deadline" column at all
+      - at_risk  : "response_deadline" column AND ">=" (the now-to-threshold window)
+      - breached : "response_deadline" column present but no ">=" (past-deadline only)
+
+    Uses column-name + operator detection rather than raw ">" / "<" substring scanning.
+    """
+
+    def handler(sql_text: str, params: dict):
```
yihua-zhuo: **CRITICAL** (critical)

**Issue:** Mock handler never inspects params['tenant_id'] and returns the same counts for all tenants, so a broken query with no tenant filter would silently pass all tests.

**Suggestion:** Have the handler validate that params contains a tenant_id key matching the expected tenant, or inspect the SQL text for 'tenant_id = :tenant_id' before returning counts. For example: if tenant_id == 2, return only on a query that includes tenant filtering.

tests/unit/test_sla_service.py
```diff
@@ -0,0 +1,292 @@
...
+    """Minimal mock AsyncSession that routes session.scalar() to SQL handlers.
+
+    get_sla_summary calls session.scalar(select(func.count(...))) directly,
+    not session.execute(), so we need to support that path.
+    """
+
+    def __init__(self, handlers):
+        self._handlers = handlers
+
+    async def scalar(self, sql, params=None):
```
yihua-zhuo: **CRITICAL** (critical)

**Issue:** Scalar handler ignores params argument entirely — it receives bind parameters but never uses them to verify the real query's tenant filter, making it trivially possible to pass the wrong placeholder name.

**Suggestion:** Read and assert on params['tenant_id'] inside the scalar() method; raise AssertionError if the query the service sends omits the tenant bind.

yihua-zhuo: **CRITICAL** (critical)

**Issue:** Rule 133 violation: handler never validates that the SQL under test actually binds tenant_id. A query with WHERE tenant_id = 1 hardcoded instead of :tenant_id would pass the mock.

**Suggestion:** Inspect params in scalar() to confirm the service passes tenant_id as a bind parameter, not a literal.

src/api/routers/tickets.py
```diff
@@ -420,3 +429,18 @@ async def get_sla_breach_tickets(
...
+async def get_sla_summary(
+    ctx: AuthContext = Depends(require_auth),
+    session: AsyncSession = Depends(get_db),
+):
+    """Return aggregated SLA counts across the full ticket dataset for the current tenant."""
+    sla_svc = SLAService(session)
+    counts = await sla_svc.get_sla_summary(tenant_id=ctx.tenant_id or 0)
+    return {
+        "success": True,
+        "data": SLAStatCard(**counts).model_dump(),
```
yihua-zhuo, line 444: **INFO** (suggestion)

**Issue:** SLAStatCard is used once and could be inlined; keeping it is acceptable as a named schema.

**Suggestion:** No action needed — schema class is a documented public API contract.

src/services/sla_service.py
```diff
@@ -1,15 +1,20 @@
...
 
+if TYPE_CHECKING:
+    from services.ticket_service import TicketService
+
 
 class SLAService:
     """SLA管理"""
 
-    def __init__(self, session: AsyncSession, ticket_service=None):
+    def __init__(self, session: AsyncSession, ticket_service: "TicketService | None" = None):
```
yihua-zhuo, line 17: **INFO** (suggestion)

**Issue:** Rule 122: get_sla_summary issues three scalar COUNT queries — one SELECT per bucket. Three separate round-trips is acceptable for this scale but worth documenting if the dataset grows.

**Suggestion:** Consider a single query with COUNT OVER for large tenants: `select response_deadline IS NULL ... as bucket, count(*) over ()`.

src/services/sla_service.py
```diff
@@ -39,6 +44,57 @@ def check_sla_status(self, ticket: Ticket) -> Literal["normal", "warning", "brea
...
+    async def get_sla_summary(self, tenant_id: int) -> dict[str, int]:
+        """Return aggregated SLA counts across the full ticket dataset for a tenant.
+
+        Counts are bucketed as:
+          - breached: open, deadline set, deadline < now
+          - at_risk : open, deadline set, now < deadline <= now + 4h
+          - on_track: resolved OR no deadline OR deadline > now + 4h
+          - total_tickets: all tickets for the tenant
+        """
+        now = datetime.now(UTC)
```
yihua-zhuo, line 61: **INFO** (suggestion)

**Issue:** Rule 16: Docstring is in English; inline Chinese comments on lines 22-23 and 105 use mixed language. Not a bug, but inconsistent.

**Suggestion:** Standardize all docstrings to English for consistency.

src/services/sla_service.py
```diff
@@ -39,6 +44,57 @@ def check_sla_status(self, ticket: Ticket) -> Literal["normal", "warning", "brea
 
         return "normal"
 
+    async def get_sla_summary(self, tenant_id: int) -> dict[str, int]:
```
yihua-zhuo, line 52: **INFO** (suggestion)

**Issue:** Rule 5: Two dead comments in get_sla_summary (lines 49-54 describe bucket semantics verbatim) are unnecessary — the method name and return dict keys already convey this.

**Suggestion:** Remove lines 49-54 block comment; keep only the single-line description above.

tests/integration/test_sla_integration.py
```diff
@@ -0,0 +1,348 @@
...
+    """Seed 3 breached, 2 at_risk, 5 on_track tickets; assert counts match."""
+    from db.models.ticket import TicketModel
+    from services.sla_service import SLAService
+
+    now = datetime.now(UTC)
+    four_hours = timedelta(hours=4)
+
+    # 3 breached: resolved_at=None, deadline in past
+    for _ in range(3):
+        async_session.add(
```
yihua-zhuo, line 35: **WARNING** (suggestion)

**Issue:** Rule 136: TicketModel is constructed with status='open' (string) instead of TicketStatus.OPEN (enum). If the model field is typed as Enum this may cause silent coercion issues or ignore type checking.

**Suggestion:** Use TicketStatus.OPEN instead of 'open' string throughout: `status=TicketStatus.OPEN`.

tests/integration/test_sla_integration.py
```diff
@@ -0,0 +1,348 @@
...
+                subject="AtRisk",
+                description="desc",
+                status="open",
+                priority="medium",
+                channel="email",
+                customer_id=1,
+                resolved_at=None,
+                response_deadline=now + timedelta(hours=2),
+            )
+        )
```
yihua-zhuo, line 63: **WARNING** (suggestion)

**Issue:** Rule 136: Same string-vs-enum issue as line 19 — status='open' used in at_risk ticket seeds.

**Suggestion:** Use TicketStatus.OPEN instead of 'open' string.

tests/integration/test_sla_integration.py
```diff
@@ -0,0 +1,348 @@
...
+
+    assert result["breached"] == 0
+    assert result["at_risk"] == 0
+    assert result["on_track"] == 0
+    assert result["total_tickets"] == 0
+
+
+async def test_sla_summary_tenant_isolation(db_schema, tenant_id, tenant_id_2, async_session):
+    """Seed tickets for two different tenant_ids; each query only sees its own."""
+    from datetime import UTC
```
yihua-zhuo: **INFO** (suggestion)

**Issue:** Duplicate import: `from datetime import UTC` is re-imported locally (line 106) even though it was already imported at module level (line 2).

**Suggestion:** Remove the local `from datetime import UTC` on line 106 — module-level import on line 2 is sufficient.

tests/unit/test_sla_service.py
```diff
@@ -0,0 +1,292 @@
...
+
+        handler = make_sla_summary_handler(breached=1, at_risk=1, total=3)
+        session = _SlaMockSession([handler])
+        svc = SLAService(session)
+        result = await svc.get_sla_summary(tenant_id=0)
+
+        assert result["total_tickets"] == 3
+        assert result["on_track"] == 1  # 3 - 1 - 1
+
+    async def test_on_track_is_derived_not_queried(self):
```
yihua-zhuo, line 218: **INFO** (suggestion)

**Issue:** TrackingSession test uses inline MagicMock instead of the shared handler framework (Rule 135); the pattern is sound but diverges from the project convention.

**Suggestion:** Consider using make_sla_summary_handler for the tracking test as well, or formalize the inline CountingSession as a reusable helper.

tests/unit/test_tickets_router.py
```diff
@@ -540,3 +540,101 @@ def test_not_found_returns_404(self, client_with_service):
...
+    def test_service_receives_tenant_id_from_auth_context(self, client_with_service):
+        """Endpoint must pass the authenticated tenant_id to the service."""
+        client, _, sla_svc = client_with_service
+        sla_svc.get_sla_summary = AsyncMock(
+            return_value={"breached": 1, "at_risk": 0, "on_track": 2, "total_tickets": 3}
+        )
+        resp = client.get("/api/v1/sla/summary")
+        assert resp.status_code == 200
+        # _make_auth_ctx() uses tenant_id=1; get_sla_summary must be called with tenant_id=1
+        sla_svc.get_sla_summary.assert_awaited_once_with(tenant_id=1)
```
yihua-zhuo, line 601: **WARNING** (suggestion)

**Issue:** Only one test (TestGetSlaSummaryEndpoint.test_service_receives_tenant_id_from_auth_context) asserts that tenant_id is forwarded to the service; all other service methods receive no tenant_id assertion.

**Suggestion:** Add tenant_id verification to representative tests for list_tickets, get_ticket, get_sla_breaches, and get_breach_tickets — at least one test per method family should assert the tenant_id bind.

tests/unit/conftest.py
```diff
@@ -860,6 +860,137 @@ def campaign_handler(sql_text, params):
...
+
+        try:
+            # Compile with a specific dialect to get the rendered SQL string
+            compiled = sql.compile(dialect=self._get_dialect())
+            # Normalize to lower-case so handlers can match patterns case-insensitively
+            sql_text = str(compiled).lower()
+            # compiled.params is a dict (attribute, not method in SQLAlchemy 2.x)
+            for k, v in compiled.params.items():
+                if k not in params:
+                    params[k] = v
```
yihua-zhuo: **CRITICAL** (critical)

**Issue:** Bare `except Exception` swallows all errors silently, masking SQL compilation failures in tests.

**Suggestion:** Log or re-raise compilation errors so broken SQL is surfaced rather than silently falling through to raw-string matching.

tests/unit/test_sla_service.py
```diff
@@ -0,0 +1,239 @@
...
+# ---------------------------------------------------------------------------
+# Tests
+# ---------------------------------------------------------------------------
+
+
+class TestGetSlaSummary:
+    async def test_all_categories_populated(self, mock_db_session_all_populated):
+        from services.sla_service import SLAService
+
+        svc = SLAService(mock_db_session_all_populated)
```
yihua-zhuo, line 87: **CRITICAL** (critical)

**Issue:** SLAService.__init__ declares session as AsyncSession with no default; SlaMockSession only provides scalar/__aenter__/__aexit__ and does not inherit from AsyncSession, violating the type contract.

**Suggestion:** Make SlaMockSession inherit from AsyncSession (even if via a test-specific stub) so type-checkers and readers see the contract is intentionally satisfied.

.plans/issue-113.md
```diff
@@ -0,0 +1,69 @@
...
+   - Add `GET /sla/summary` to `tickets_router` (in the existing `# SLA endpoints` section alongside `check_sla_status` and `get_sla_breach_tickets`):
+     ```python
+     @tickets_router.get("/sla/summary")
+     async def get_sla_summary(
+         ctx: AuthContext = Depends(require_auth),
+         session: AsyncSession = Depends(get_db),
+     ):
+         """Return aggregated SLA counts across the full ticket dataset for the current tenant."""
+         sla_svc = SLAService(session)
+         counts = await sla_svc.get_sla_summary(tenant_id=ctx.tenant_id or 0)
```
yihua-zhuo, line 39: **INFO** (suggestion)

**Issue:** Plan uses `ctx.tenant_id or 0` which silently coerces None to 0 — same pattern exists in other endpoints so is consistent, but masks a missing tenant.

**Suggestion:** Consider documenting that `or 0` is intentional fallback and not a bug, or align with any stricter auth middleware that would reject unauthenticated requests before the endpoint runs.

.plans/issue-113.md
```diff
@@ -0,0 +1,69 @@
...
+         return {"success": True, "data": counts}
+     ```
+   - The endpoint requires auth (same as other SLA endpoints) and uses the standard dependency injection pattern
+
+3. **Write unit tests** (`tests/unit/test_sla_service.py` — new file):
```
yihua-zhuo, line 44: **INFO** (suggestion)

**Issue:** Plan specifies creating `tests/unit/test_sla_service.py` but no unit test file exists — only the integration test file was created.

**Suggestion:** Create `tests/unit/test_sla_service.py` with mocked-DB tests per the plan, or update the plan to remove the unit-test step if integration-only coverage was the final decision.

tests/integration/test_sla_integration.py
```diff
@@ -0,0 +1,348 @@
...
+    for _ in range(3):
+        async_session.add(
+            TicketModel(
+                tenant_id=tenant_id,
+                subject="Breached",
+                description="desc",
+                status="open",
+                priority="medium",
+                channel="email",
+                customer_id=1,
```
yihua-zhuo: **INFO** (suggestion)

**Issue:** TicketModel inserts use hardcoded customer_id=1 throughout the file. No customer is seeded in these tests. If the tickets table has a foreign key to customers, inserts may fail on a clean db_schema.

**Suggestion:** Seed a customer record for tenant_id first, or use a customer_id that is guaranteed to exist in the db_schema fixture (e.g. via a _seed_customer helper if one exists in tests/integration/conftest.py).

tests/unit/conftest.py
```diff
@@ -860,6 +860,137 @@ def campaign_handler(sql_text, params):
...
+
+class SlaMockSession:
+    """Minimal mock AsyncSession that routes session.scalar() to SQL handlers.
+
+    get_sla_summary calls session.scalar(select(func.count(...))) directly,
+    not session.execute(), so we need to support that path.
+    """
+
+    def __init__(self, handlers):
+        self._handlers = handlers
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** SlaMockSession only implements `scalar()` but not `execute()` — tests that call `session.execute()` will receive a bare MagicMock with no SQL routing, and will silently pass against the wrong handler.

**Suggestion:** Add an `execute()` method to SlaMockSession that mirrors the scalar routing logic, so any test that calls execute() also gets routed to the handlers.

tests/unit/conftest.py
```diff
@@ -860,6 +860,137 @@ def campaign_handler(sql_text, params):
...
+
+    def handler(sql_text: str, params: dict):
+        if not ("select" in sql_text and "count" in sql_text and "from tickets" in sql_text):
+            return None
+
+        # Tenant-scope assertion: catch silent tenant_id omissions.
+        # SQLAlchemy may label the bind as "tenant_id_1" when multiple tenant_id
+        # columns are present, so we check that at least one param key contains
+        # "tenant_id" as a prefix.
+        if validate_tenant_id:
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** Tenant validation only reads `tenant_id` or `tenant_id_1`; SQLAlchemy may generate other keys (tenant_id_2, etc.) when multiple tenant_id conditions exist, allowing a mismatched tenant query to pass undetected.

**Suggestion:** Inspect all params keys starting with 'tenant_id' and validate each against expected_tenant_id, or assert that only one such key exists.

tests/unit/conftest.py
```diff
@@ -873,11 +1004,6 @@ def handler(sql_text, params):
...
 
 
-# ---------------------------------------------------------------------------
-# Session builder
-# ---------------------------------------------------------------------------
-
-
 def _load_domain_handler_modules():
     """Import domain-owned handler modules and re-export their named helpers."""
     package_name = "tests.unit.domain_handlers"
```
yihua-zhuo: **INFO** (suggestion)

**Issue:** _load_domain_handler_modules() is called at module import time with no try/except — a broken domain handler module will crash the entire test suite at collection time with an opaque import error.

**Suggestion:** Wrap the import loop in try/except ImportError and either skip the module or fail with a descriptive message listing the unavailable module.

tests/unit/test_sla_service.py
```diff
@@ -0,0 +1,239 @@
+"""Unit tests for SLAService.get_sla_summary."""
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** No cross-tenant negative test: all tests use tenant_id=1 or 2 with matching mock responses; missing a test where the service is called with tenant_id=1 but the query returns only tenant_id=2 data to confirm tenant isolation.

**Suggestion:** Add a test case where the mock returns empty/foreign data for a mismatched tenant_id and assert the result reflects zero counts.

yihua-zhuo: **INFO** (suggestion)

**Issue:** Tests directly import and call SLAService but there is no corresponding unit test for get_breach_tickets() — the method is only exercised via the router endpoint test, leaving the service contract untested in isolation.

**Suggestion:** Add a direct unit test for get_breach_tickets() similar to the get_sla_summary tests, asserting the filtering logic and optional tickets parameter handling.

tests/integration/test_sla_integration.py
```diff
@@ -0,0 +1,347 @@
...
+pytestmark = pytest.mark.integration
+
+
+async def test_sla_summary_all_categories(db_schema, tenant_id, async_session, _seed_customer):
+    """Seed 3 breached, 2 at_risk, 5 on_track tickets; assert counts match."""
+    from db.models.ticket import TicketModel
+    from services.sla_service import SLAService
+
+    now = datetime.now(UTC)
+    cid = _seed_customer
```
yihua-zhuo: **CRITICAL** (critical)

**Issue:** Line 16 calls `cid = _seed_customer` as a synchronous callable, but `_seed_customer` is an async fixture defined in `tests/integration/conftest.py:253` that must be awaited (returns `customer.id` integer after `await async_session.flush()`). Omitting `await` assigns the coroutine object itself, not the customer ID integer.

**Suggestion:** Change `cid = _seed_customer` to `cid = await _seed_customer`. All other integration tests in the project (e.g. `test_tickets_integration.py:49`, `test_intelligence_integration.py:41`) use `cid = await _seed_customer(async_session, tenant_id)` correctly — this file is the outlier.

src/api/routers/tickets.py
```diff
@@ -420,3 +429,18 @@ async def get_sla_breach_tickets(
...
+
+
+@tickets_router.get("/sla/summary")
+async def get_sla_summary(
+    ctx: AuthContext = Depends(require_auth),
+    session: AsyncSession = Depends(get_db),
+):
+    """Return aggregated SLA counts across the full ticket dataset for the current tenant."""
+    sla_svc = SLAService(session)
+    counts = await sla_svc.get_sla_summary(tenant_id=ctx.tenant_id or 0)
```
yihua-zhuo, line 441: **WARNING** (suggestion)

**Issue:** Endpoint falls back to tenant_id=0 when ctx.tenant_id is None, silently accepting an unauthenticated request instead of rejecting it. Rule 126.

**Suggestion:** Remove the `or 0` fallback and let the request fail with 401 if no tenant context is available, or assert tenant_id is non-zero at the dependency level.

tests/unit/conftest.py
```diff
@@ -860,6 +866,170 @@ def campaign_handler(sql_text, params):
...
+
+            warnings.warn(
+                f"SlaMockSession.execute: SQL compilation failed ({exc}), falling back to raw string", UserWarning
+            )
+            sql_text = str(sql).lower().strip()
+
+        for h in self._handlers:
+            result = h(sql_text, params)
+            if result is not None:
+                return MockResult([[result]])
```
yihua-zhuo: **INFO** (suggestion)

**Issue:** SlaMockSession.execute() wraps handler results in an extra list layer: `MockResult([[result]])`. When `result` is a plain value (int, None), `fetchone()` returns `[value]` — a list, not the scalar value. By contrast, SlaMockSession.scalar() returns handler results directly. The SLA service uses `session.scalar()` exclusively, so this does not affect current tests, but any caller using `session.execute()` would get malformed results.

**Suggestion:** Align SlaMockSession.execute() with scalar(): return `MockResult([result])` (single wrapping) or return the handler's MockResult directly without re-wrapping.

tests/unit/test_sla_service.py
```diff
@@ -0,0 +1,333 @@
...
+                status=TicketStatus.RESOLVED,
+                priority=TicketPriority.MEDIUM,
+                channel=TicketChannel.EMAIL,
+                sla_level=SLALevel.STANDARD,
+                resolved_at=now - timedelta(hours=1),
+                response_deadline=breached_deadline,
+            ),  # resolved → not breached
+        ]
+
+        svc = SLAService(session=None)
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** Rule 5 violated: SLAService is instantiated with session=None (`SLAService(session=None)`), but the service contract requires `AsyncSession` with no default. The tests happen to work because `get_breach_tickets` with an in-memory tickets list does not call self._session. However, instantiating the service in an invalid state masks whether the __init__ contract is respected.

**Suggestion:** Use the sla_mock_session fixture consistently: `SLAService(sla_mock_session([...]))` — consistent with `test_all_at_risk`, `test_on_track_is_derived_not_queried`, etc. Pass an empty handler list when the session is not used.

tests/unit/test_sla_service.py
```diff
@@ -0,0 +1,333 @@
...
+        result = await svc.get_sla_summary(tenant_id=1)
+
+        # Service must forward tenant_id=1 → handler must see tenant_id=1 → counts
+        # returned must be from the correct bucket, not from cross-tenant data.
+        assert result["breached"] == 99
+        assert result["at_risk"] == 99
+        assert result["total_tickets"] == 99
+
+
+class TestGetBreachTickets:
```
yihua-zhuo, line 252: **INFO** (suggestion)

**Issue:** test_cross_tenant_data_isolation is named for isolation but only tests that tenant_id=1 is forwarded to the mock handler; it does not test that data from tenant 2 is invisible to tenant 1. The conftest.make_sla_summary_handler validates the bind parameter name but the stateless mock returns fixed counts regardless of which tenant_id value is passed.

**Suggestion:** Rename to `test_tenant_id_bind_is_forwarded` or extend the handler to return different counts per expected_tenant_id and assert the isolation property directly.

src/api/routers/tickets.py
```diff
@@ -420,3 +429,18 @@ async def get_sla_breach_tickets(
...
+
+
+@tickets_router.get("/sla/summary")
+async def get_sla_summary(
+    ctx: AuthContext = Depends(require_auth),
+    session: AsyncSession = Depends(get_db),
+):
+    """Return aggregated SLA counts across the full ticket dataset for the current tenant."""
+    sla_svc = SLAService(session)
+    counts = await sla_svc.get_sla_summary(tenant_id=ctx.tenant_id)
```
yihua-zhuo: **CRITICAL** (critical)

**Issue:** AuthContext.tenant_id is typed int | None; passing it directly to SLAService.get_sla_summary(tenant_id: int) violates the int contract and differs from every other endpoint in this router which uses 'ctx.tenant_id or 0' (rules 27, 126).

**Suggestion:** Change line 441 to: counts = await sla_svc.get_sla_summary(tenant_id=ctx.tenant_id or 0)

tests/unit/conftest.py
```diff
@@ -282,6 +293,165 @@ async def _mock_aenter():
     return session
 
 
+# ---------------------------------------------------------------------------
+# SLA mock session (scalar-based, for SLAService tests)
+# ---------------------------------------------------------------------------
```
yihua-zhuo: **CRITICAL** (critical)

**Issue:** Rule 149 violation: SLA mock primitives (SlaMockSession, sla_mock_session, make_sla_summary_handler) were added to tests/unit/conftest.py for a domain feature; they should live in tests/unit/domain_handlers/sla.py or tests/unit/sla_conftest.py.

**Suggestion:** Extract SlaMockSession, sla_mock_session, and make_sla_summary_handler into tests/unit/domain_handlers/sla.py (or tests/unit/sla_conftest.py) and re-export from conftest for backward compat.

.plans/issue-113.md
yihua-zhuo, line 39: **WARNING** (suggestion)

**Issue:** Plan step 2 documents 'ctx.tenant_id or 0' as the fallback but the actual endpoint at tickets.py:441 uses bare 'ctx.tenant_id' — plan is stale (rule 140).

**Suggestion:** Update line 39 to: counts = await sla_svc.get_sla_summary(tenant_id=ctx.tenant_id) — or reconcile plan with implementation.

tests/integration/test_sla_integration.py
```diff
@@ -0,0 +1,362 @@
...
+    for _ in range(2):
+        async_session.add(
+            TicketModel(
+                tenant_id=tenant_id,
+                subject="AtRisk",
+                description="desc",
+                status="open",
+                priority="medium",
+                channel="email",
+                customer_id=0,
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** Hardcoded customer_id=0 creates invalid FK references in at_risk tickets; cid is already seeded but not used.

**Suggestion:** Use cid instead of 0: customer_id=cid, for all at_risk, resolved, and on_track ticket seeds where customer_id=0 appears.

tests/integration/test_sla_integration.py
```diff
@@ -0,0 +1,362 @@
...
+    for _ in range(3):
+        async_session.add(
+            TicketModel(
+                tenant_id=tenant_id,
+                subject="Resolved",
+                description="desc",
+                status="open",
+                priority="medium",
+                channel="email",
+                customer_id=0,
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** Hardcoded customer_id=0 creates invalid FK references in resolved tickets.

**Suggestion:** Use cid instead of 0 for these resolved ticket seeds.

tests/integration/test_sla_integration.py
```diff
@@ -0,0 +1,362 @@
...
+        )
+    for _ in range(2):
+        async_session.add(
+            TicketModel(
+                tenant_id=tenant_id,
+                subject="OnTrack",
+                description="desc",
+                status="open",
+                priority="medium",
+                channel="email",
```
yihua-zhuo, line 88: **WARNING** (suggestion)

**Issue:** Hardcoded customer_id=0 creates invalid FK references in on_track tickets.

**Suggestion:** Use cid instead of 0 for these on_track ticket seeds.

tests/integration/test_sla_integration.py
```diff
@@ -0,0 +1,362 @@
...
+        async_session.add(
+            TicketModel(
+                tenant_id=tenant_id,
+                subject="AtRisk",
+                description="desc",
+                status="open",
+                priority="medium",
+                channel="email",
+                customer_id=0,
+                resolved_at=None,
```
yihua-zhuo, line 60: **INFO** (suggestion)

**Issue:** Tickets created with status='open' then used to test 'resolved' SLA behavior may not reflect real-world resolved-state semantics.

**Suggestion:** Use status='resolved' for resolved ticket fixtures to match real SLA model behavior.

tests/unit/test_sla_service.py
```diff
@@ -0,0 +1,334 @@
...
+    async def test_filters_breached_from_list(self):
+        """get_breach_tickets filters the provided list to only breached tickets."""
+        from services.sla_service import SLAService
+        from tests.unit.conftest import sla_mock_session
+
+        # Build a mixed list: 2 breached, 1 on-track, 1 at-risk
+        now = datetime.utcnow()
```
yihua-zhuo: **INFO** (suggestion)

**Issue:** datetime.utcnow() is deprecated since Python 3.12; the rest of the test suite uses datetime.now(UTC).

**Suggestion:** Replace datetime.utcnow() with datetime.now(UTC) to stay consistent with the rest of the codebase.

.plans/issue-113.md
```diff
@@ -0,0 +1,69 @@
...
+   - Create `MockSlaTicket` class mimicking `TicketModel` with `id`, `tenant_id`, `status`, `resolved_at`, `response_deadline` fields and a `check_sla_breach()` method
+   - Define a `mock_sla_db_session(state, tickets)` fixture that returns pre-seeded tickets matching given conditions (breached, at_risk, on_track) on the appropriate SQL `select count... from tickets` queries
+   - Test cases: all categories populated, zero tickets, all breached, only on_track, tenant isolation (two tenants return different counts)
+   - Use `MockRow`/`MockResult` from `tests/unit/conftest.py` for SQL response mocking
+
```
yihua-zhuo, line 49: **CRITICAL** (critical)

**Issue:** Plan step 3 references a non-existent `MockSlaTicket` class that is not implemented anywhere in the codebase.

**Suggestion:** Remove the `MockSlaTicket` class reference. The actual implementation uses `SlaMockSession`, `sla_mock_session`, and `make_sla_summary_handler` from `tests/unit/domain_handlers/sla.py`.

yihua-zhuo, line 49: **CRITICAL** (critical)

**Issue:** Plan step 3 references `mock_sla_db_session` fixture which does not exist. The actual fixture is named `sla_mock_session` in `tests/unit/domain_handlers/sla.py`.

**Suggestion:** Update plan step 3 to reference `sla_mock_session` (from `tests/unit/domain_handlers/sla.py`) instead of `mock_sla_db_session`.

yihua-zhuo, line 49: **CRITICAL** (critical)

**Issue:** Plan step 3 references `MockRow`/`MockResult` from `tests/unit/conftest.py` for SLA test mocking, but the actual implementation uses `SlaMockSession` + `make_sla_summary_handler` in `tests/unit/domain_handlers/sla.py`.

**Suggestion:** Update plan step 3 to reflect the actual SLA mock implementation pattern using `SlaMockSession`.

src/services/sla_service.py
```diff
@@ -39,6 +49,57 @@ def check_sla_status(self, ticket: Ticket) -> Literal["normal", "warning", "brea
...
+        on_track_count = (total_count or 0) - (breached_count or 0) - (at_risk_count or 0)
+
+        return {
+            "breached": breached_count or 0,
+            "at_risk": at_risk_count or 0,
+            "on_track": on_track_count,
+            "total_tickets": total_count or 0,
+        }
+
     async def get_breach_tickets(self, tenant_id: int = 0, tickets: list[Ticket] = None) -> list[Ticket]:
```
yihua-zhuo, line 103: **CRITICAL** (critical)

**Issue:** Mutable default argument `tickets: list[Ticket] = None` violates Python best practice (Rule 22). Although callers always pass the argument explicitly so no bug occurs, the default is unnecessary and misleading.

**Suggestion:** Change signature to `async def get_breach_tickets(self, tenant_id: int = 0, tickets: list[Ticket] | None = None)` and update the internal `tickets or []` logic accordingly.

tests/unit/conftest.py
```diff
@@ -34,6 +34,12 @@
 if str(_src_root) not in sys.path:
     sys.path.insert(0, str(_src_root))
 
+# Re-export SLA domain handlers for backward compat — primary location is tests/unit/domain_handlers/sla.py
+from tests.unit.domain_handlers.sla import (  # noqa: F401, E402
```
yihua-zhuo, line 38: **CRITICAL** (critical)

**Issue:** conftest.py re-exports SLA domain symbols (SlaMockSession, make_sla_summary_handler, sla_mock_session), making the shared conftest a domain-owned registry — violating rule 125 which protects conftest.py from domain edits.

**Suggestion:** Remove lines 38-42. Domain handlers in domain_handlers/sla.py should be imported directly by consumer test files; backward compat is already served by test_sla_service.py importing from conftest, which in turn imports from sla.py — but the direction of dependency is inverted.

src/api/routers/tickets.py
```diff
@@ -420,3 +429,18 @@ async def get_sla_breach_tickets(
         "data": [t.to_dict() for t in tickets],
         "message": "查询成功",
     }
+
+
+@tickets_router.get("/sla/summary")
```
yihua-zhuo, line 434: **INFO** (suggestion)

**Issue:** Plan item 2 shows `GET /sla/summary` added under the existing SLA section, but the plan's acceptance criteria don't list unit/integration tests as deliverables, yet the implementation includes comprehensive test files. This is an improvement over the plan, not a violation.

**Suggestion:** No change needed; implementation exceeds the plan, which is acceptable.

tests/integration/test_sla_integration.py
```diff
@@ -0,0 +1,362 @@
...
+    for _ in range(2):
+        async_session.add(
+            TicketModel(
+                tenant_id=tenant_id,
+                subject="Breached",
+                description="desc",
+                status="open",
+                priority="medium",
+                channel="email",
+                customer_id=0,
```
yihua-zhuo, line 136: **WARNING** (suggestion)

**Issue:** Multiple tests use customer_id=0 without seeding a matching customer; this is fragile and will break silently if a FK constraint is added to tickets.customer_id.

**Suggestion:** Seed a real customer via _seed_customer and use its returned id instead of the bare 0 placeholder.

tests/integration/test_sla_integration.py
```diff
@@ -0,0 +1,362 @@
...
+        async_session.add(
+            TicketModel(
+                tenant_id=tenant_id,
+                subject="AtRisk",
+                description="desc",
+                status="open",
+                priority="medium",
+                channel="email",
+                customer_id=0,
+                resolved_at=None,
```
yihua-zhuo, line 151: **WARNING** (suggestion)

**Issue:** Same customer_id=0 bare-placeholder issue.

**Suggestion:** Seed a real customer and use its id.

tests/integration/test_sla_integration.py
```diff
@@ -0,0 +1,362 @@
...
+    for _ in range(5):
+        async_session.add(
+            TicketModel(
+                tenant_id=tenant_id_2,
+                subject="Breached",
+                description="desc",
+                status="open",
+                priority="medium",
+                channel="email",
+                customer_id=0,
```
yihua-zhuo, line 166: **WARNING** (suggestion)

**Issue:** Same customer_id=0 bare-placeholder issue.

**Suggestion:** Seed a real customer and use its id.

tests/integration/test_sla_integration.py
```diff
@@ -0,0 +1,362 @@
...
+        )
+    async_session.add(
+        TicketModel(
+            tenant_id=tenant_id_2,
+            subject="AtRisk",
+            description="desc",
+            status="open",
+            priority="medium",
+            channel="email",
+            customer_id=0,
```
yihua-zhuo, line 179: **WARNING** (suggestion)

**Issue:** Same customer_id=0 bare-placeholder issue.

**Suggestion:** Seed a real customer and use its id.

tests/integration/test_sla_integration.py
```diff
@@ -0,0 +1,362 @@
...
+    assert r2["breached"] == 5
+    assert r2["at_risk"] == 1
+    assert r2["total_tickets"] == 6
+
+
+async def test_sla_summary_no_deadline_tickets_counted_as_on_track(db_schema, tenant_id, async_session):
+    """Tickets with no response_deadline should be counted as on_track, not breached."""
+    from db.models.ticket import TicketModel
+    from services.sla_service import SLAService
+
```
yihua-zhuo, line 204: **WARNING** (suggestion)

**Issue:** customer_id=0 bare-placeholder issue in at_risk boundary test.

**Suggestion:** Seed a real customer and use its id.

tests/integration/test_sla_integration.py
```diff
@@ -0,0 +1,362 @@
...
+    # 1 breached ticket to ensure non-deadline tickets aren't misclassified
+    async_session.add(
+        TicketModel(
+            tenant_id=tenant_id,
+            subject="Breached",
+            description="desc",
+            status="open",
+            priority="medium",
+            channel="email",
+            customer_id=0,
```
yihua-zhuo, line 232: **WARNING** (suggestion)

**Issue:** customer_id=0 bare-placeholder issue in on_track boundary test.

**Suggestion:** Seed a real customer and use its id.

tests/unit/domain_handlers/sla.py
```diff
@@ -0,0 +1,135 @@
...
+
+        for h in self._handlers:
+            result = h(sql_text, params)
+            if result is not None:
+                return result
+        return None
+
+    async def execute(self, sql, params=None):
+        """Route session.execute() to the same handlers as scalar()."""
+        from tests.unit.conftest import MockResult
```
yihua-zhuo, line 45: **WARNING** (suggestion)

**Issue:** Late import of MockResult from tests.unit.conftest inside a method body creates implicit coupling; if conftest's internals change, sla.py breaks silently.

**Suggestion:** Move 'from tests.unit.conftest import MockResult' to the module-level TYPE_CHECKING block at the top.

tests/unit/test_sla_service.py
```diff
@@ -0,0 +1,334 @@
...
+        svc = SLAService(session)
+        result = await svc.get_sla_summary(tenant_id=1)
+
+        # Handler confirmed bind parameter tenant_id=1 was used
+        assert result["breached"] == 99
+        assert result["at_risk"] == 99
+        assert result["total_tickets"] == 99
+
+
+class TestGetBreachTickets:
```
yihua-zhuo, line 252: **WARNING** (suggestion)

**Issue:** TestGetBreachTickets covers only in-memory list filtering; get_breach_tickets also has a branch that calls self._ticket_service.get_sla_breaches(tenant_id) when tickets=None and a ticket_service is configured, but no test exercises this branch.

**Suggestion:** Add a test that constructs SLAService(mock_db_session, ticket_service=fake_ticket_service) and calls get_breach_tickets(tenant_id=1, tickets=None) to verify the service-query fallback path.

tests/unit/test_sla_service.py
```diff
@@ -0,0 +1,334 @@
...
+    async def test_return_value_has_all_required_keys(self, mock_db_session_all_populated):
+        """Result dict must contain exactly the four documented keys."""
+        from services.sla_service import SLAService
+
+        svc = SLAService(mock_db_session_all_populated)
+        result = await svc.get_sla_summary(tenant_id=1)
+
+        assert set(result.keys()) == {"breached", "at_risk", "on_track", "total_tickets"}
+
+    async def test_null_scalar_responses_default_to_zero(self):
```
yihua-zhuo, line 174: **INFO** (suggestion)

**Issue:** test_null_scalar_responses_default_to_zero passes a handler returning None for every query, which exercises the service's None → 0 fallback; this is a good defensive test but the handler also matches non-SLA queries that the real service would never issue, giving a false sense of coverage.

**Suggestion:** Narrow the null handler to only match SLA queries: lambda sql, params: None if 'from tickets' in sql and 'count' in sql else MockResult([[None]])

tests/unit/test_sla_service.py
```diff
@@ -0,0 +1,334 @@
...
+        """Passing an empty list returns an empty result."""
+        from services.sla_service import SLAService
+        from tests.unit.conftest import sla_mock_session
+
+        svc = SLAService(sla_mock_session([]))
+        result = await svc.get_breach_tickets(tickets=[])
+
+        assert result == []
+
+    async def test_handles_none_tickets_without_ticket_service(self):
```
yihua-zhuo, line 326: **INFO** (suggestion)

**Issue:** Test name says 'without_ticket_service' but the assert checks tickets=[], not the tickets=None path specifically; the test exercises the early-return for tickets=None, but the name does not reflect what is actually being tested.

**Suggestion:** Rename to test_returns_empty_when_tickets_is_none_without_ticket_service, or merge with test_returns_empty_when_no_tickets and cover both None and [] in one parameterized test.

PR #440, opened 2026-05-14
Auto: implement issue #143

### LINKED ISSUE #143: [Test] Add unit tests for ActivityService — CRUD + pagination
## Task
Create `tests/unit/test_activity_service.py` covering core CRUD operations in `ActivityService`.

## Implementation
1. Fixture `activity_service` returning `ActivityService()` instance
2. Tests:
   - `test_create_activity_returns_api_response` — create activity and assert ApiResponse with correct fields
   - `test_get_activity_found` — create then get by id, assert content matches
   - `test_get_activity_not_found` — get non-existent id, assert bool is False
   - `test_update_activity` — update content, assert updated value
   - `test_delete_activity` — delete then confirm get returns not found
   - `test_list_activities_pagination` — create 3, list with page_size=2, assert len==2 and total==3

## Verification
- `pytest tests/unit/test_activity_service.py --collect-only` — 6 items, 0 errors
- `pytest tests/unit/test_activity_service.py -v` — all pass
- `pytest tests/unit/ -q` — suite still green

yihua-zhuo:
Closes #143

Plan: `.plans/issue-143.md`

1. **Plan** — Claude wrote a detailed implementation plan to `.plans/issue-143.md`.

Please review before merging.

## Summary by CodeRabbit

* **Tests**
  * Added a comprehensive unit test suite for the Activity service covering create/read/update/delete, validation, tenant isolation, and pagination.
  * Expanded test fixtures and in-memory mock handlers to simulate related domain data and DB behaviors for detailed assertions.
* **Documentation**
  * Added an implementation plan describing test scope, acceptance criteria, and open questions.
* **Chores**
  * CI improved to detect and block unintended edits to protected repository paths.

[Review Change Stack](https://app.coderabbit.ai/change-stack/yihua-zhuo/agent-job/pull/440?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


tests/unit/test_activity_service.py
```diff
@@ -0,0 +1,234 @@
...
+from unittest.mock import AsyncMock, MagicMock
+
+import pytest
+
+from models.activity import Activity, ActivityType
+from pkg.errors.app_exceptions import NotFoundException
+from services.activity_service import ActivityService
+
+
+@pytest.fixture
```
yihua-zhuo, line 13: **CRITICAL** (critical)

**Issue:** MockRow._fields is never accessed by _to_activity (which uses dot-access on the ORM model), but it is defined on the class. It is dead code — it shadows a private attribute convention and serves no role in these tests.

**Suggestion:** Remove the _fields property from MockRow in conftest.py; it is unused in test_activity_service.py.

tests/unit/test_activity_service.py
```diff
@@ -0,0 +1,234 @@
...
+            scalar_one_or_none=MagicMock(return_value=None),
+        ))
+
+        service = ActivityService(mock_db_session)
+
+        with pytest.raises(NotFoundException):
+            await service.get_activity(activity_id=9999, tenant_id=1)
+
+
+class TestUpdateActivity:
```
yihua-zhuo, line 125: **CRITICAL** (critical)

**Issue:** update_activity with kwargs={} (no fields to update) calls _fetch then exits without any execute/flush call — the test mocks 3 execute calls but the real service only calls _fetch once in this path, so the side_effect list is misaligned with the actual call count.

**Suggestion:** Test the no-op branch explicitly: call update_activity(activity_id=1, tenant_id=1) with no kwargs; mock execute to return scalar_one_or_none=original; assert result.content=='Original content' and execute was called once.

.plans/issue-143.md
```diff
@@ -0,0 +1,59 @@
...
+   - Raise on invalid type for `create_activity` via `ValidationException` — the handler does not need to handle this; the service raises it before any SQL runs
+
+3. **Register `make_activity_handler` in `all_handlers()`** (`tests/unit/conftest.py`, `all_handlers` function):
+   - Append `make_activity_handler(state)` to the returned list
+
+4. **Create `tests/unit/test_activity_service.py`** with:
+   - Import `ActivityService`, `Activity`, `ActivityType` from `src`, plus `NotFoundException` and `ValidationException`
+   - `mock_db_session` fixture using `MagicMock` pattern (same as `test_customer_service.py`) so `session.execute` uses `AsyncMock(side_effect=...)`; `session.flush` returns `AsyncMock()`; `session.refresh` is a no-op
+   - `activity_service` fixture: `return ActivityService(mock_db_session())`
+   - `TestCreateActivity` class with `test_create_activity_returns_api_response` — calls `create_activity(customer_id=1, activity_type="call", content="Test call", created_by=1, tenant_id=1)`, asserts returned `Activity` has `.content == "Test call"` and `.type == ActivityType.CALL`
```
yihua-zhuo, line 36: **WARNING** (suggestion)

**Issue:** Test case name 'test_create_activity_returns_api_response' is misleading — the plan itself correctly notes the service returns an Activity domain object, not an ApiResponse envelope.

**Suggestion:** Rename to 'test_create_activity_returns_activity' to match the actual return type and align with the test class name TestCreateActivity.

.plans/issue-143.md
```diff
@@ -0,0 +1,59 @@
...
+   - `TestGetActivity` class with `test_get_activity_found` — first creates an activity via the service (use `session.flush` side-effect to populate a row the subsequent GET can find), then calls `get_activity(id=1, tenant_id=1)`, asserts `.content` matches what was created
+   - `test_get_activity_not_found` — calls `get_activity(9999, tenant_id=1)`, expects `NotFoundException`
+   - `TestUpdateActivity` class with `test_update_activity` — pre-populate mock result for `session.execute` so that `_fetch` returns a row; call `update_activity(id=1, tenant_id=1, content="Updated content")`; assert returned activity `.content == "Updated content"`
```
yihua-zhuo, line 39: **WARNING** (suggestion)

**Issue:** Test case 'test_update_activity' describes using a pre-populated mock result for '_fetch', but '_fetch' is a private method — the test should document which public call triggers the _fetch internally.

**Suggestion:** No code change needed; add a clarifying comment in the plan noting that _fetch is called internally by update_activity.

tests/unit/conftest.py
```diff
@@ -158,6 +158,8 @@ def __init__(self):
         self.users: dict[int, dict] = {}
         self.users_next_id: int = 1
         self.deleted_user_ids: set[int] = set()
+        self.activities: dict[int, dict] = {}
```
yihua-zhuo, line 163: **INFO** (suggestion)

**Issue:** MockState docstring still says 'Per-test mutable state consumed by customer & user handlers' but now also covers activities — it is out of date.

**Suggestion:** Update docstring to: 'Per-test mutable state consumed by customer, user, and activity handlers.'

tests/unit/conftest.py
```diff
@@ -614,6 +616,82 @@ def campaign_handler(sql_text, params):
...
+            return MockResult([[count_val]])
+
+        # SELECT by id
+        if "where id" in sql_text:
+            activity_id = params.get("id")
+            if activity_id in state.activities:
+                return MockResult([MockRow(state.activities[activity_id].copy())])
+            return MockResult([])
+
+        # SELECT list
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** SELECT list branch is guarded only by 'if state.activities' — this matches customer_handler but means any unmatched query on the activities table (e.g. a future JOIN or filtered list) will return all tenant rows instead of None, silently swallowing the query.

**Suggestion:** Narrow the guard to require that the query is specifically a list query, e.g.: 'if "select" in sql_text and "where id" not in sql_text:' before the state.activities check.

tests/unit/test_activity_service.py
```diff
@@ -0,0 +1,234 @@
...
+    """Tests for ActivityService.update_activity."""
+
+    @pytest.mark.asyncio
+    async def test_update_activity(self, mock_db_session):
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** update_activity flushes at line 91 of activity_service.py after the UPDATE query — the DELETE path (line 105) also flushes. These flushes are local to the router-owned transaction (rule 122), but no test verifies flush() call count for update or delete operations.

**Suggestion:** Add assertions in test_update_activity and test_delete_activity: mock_db_session.flush.assert_called() (or assert_called_once() if flush is called once per operation).

tests/unit/test_activity_service.py
```diff
@@ -0,0 +1,234 @@
...
+
+        with pytest.raises(NotFoundException):
+            await service.get_activity(activity_id=1, tenant_id=1)
+
+
+class TestListActivitiesPagination:
+    """Tests for ActivityService.list_activities pagination."""
+
+    @pytest.mark.asyncio
+    async def test_list_activities_pagination(self, mock_db_session):
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** test_list_activities_pagination mocks COUNT returning 3 total but SELECT returning only 2 items — the ActivityModel row returned by SELECT uses the full Activity(...) domain model, not a dict. Passing Activity domain objects where the service expects to call _to_activity(row) is correct, but the mock's scalars().all() returns a list directly rather than a mock iterator, inconsistent with other tests that wrap in MagicMock.

**Suggestion:** Wrap the list return in MagicMock for consistency: MagicMock(all=MagicMock(return_value=[activity1, activity2])) already does this correctly — no change needed here, the existing code is consistent.

tests/unit/test_activity_service.py
```diff
@@ -0,0 +1,234 @@
...
+@pytest.fixture
+def activity_service(mock_db_session):
+    return ActivityService(mock_db_session)
+
+
+class TestCreateActivity:
+    """Tests for ActivityService.create_activity."""
+
+    @pytest.mark.asyncio
+    async def test_create_activity_returns_activity(self, mock_db_session):
```
yihua-zhuo, line 64: **INFO** (suggestion)

**Issue:** TestCreateActivity sets mock_db_session.execute to a new AsyncMock after fixture creation — this overrides the session-level execute mock set up in the fixture, which is the correct pattern but slightly harder to follow than re-initializing session per test as other test files do.

**Suggestion:** No code change needed — the pattern is valid and mirrors test_customer_service.py correctly.

yihua-zhuo:

**Status:** `fail`
**Inline comments posted:** 26

**Summary:** Reviewed 13 file(s) between origin/master and origin/auto/issue-143.

- `tests/unit/test_activity_service.py`: UPDATE handler in activities.py (lines 59-68) validates tenant_id separately from the SQL WHERE clause — the UPDATE's WHERE is only `id = :id`. The service guards via `_fetch` (which includes `tenant_id` in its SELECT), so the mock's separate check is never the active enforcement path. A dedicated cross-tenant UPDATE test does not exist.

`tests/unit/test_activity_service.py` L255: 
**INFO** (suggestion)

**Issue:** test_list_activities_filters_customer_and_type is the last test and does not cover the combination of customer_id + activity_type + pagination together.

**Suggestion:** Add a test exercising list_activities with both filters AND pagination to verify the count and offset logic are correct under constrained result sets. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`tests/unit/test_activity_service.py` L255: 
**INFO** (suggestion)

**Issue:** The test file has no coverage for ActivityService.get_customer_activities, get_opportunity_activities, search_activities, or get_activity_summary — these are public methods with tenant-isolated SQL.

**Suggestion:** Add unit tests for the remaining service methods following the same pattern: happy path, not found, and cross-tenant rejection for each. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_

tests/unit/conftest.py
```diff
@@ -860,6 +856,116 @@ def campaign_handler(sql_text, params):
...
+        # DELETE
+        if sql_text.startswith("delete") and "activities" in sql_text:
+            activity_id = param(params, "id")
+            tenant_id = param(params, "tenant_id")
+            if activity_id in state.activities and matches_filters(state.activities[activity_id], tenant_id=tenant_id):
+                state.activities.pop(activity_id)
+                return MockResult([MockRow({"id": activity_id})])
+            return MockResult([])
+
+        # COUNT
```
yihua-zhuo: **CRITICAL** (critical)

**Issue:** The COUNT handler (line 930-942) reads tenant_id/customer_id/activity_type via param() but then counts all records in state.activities without applying any filters. Only tenant_id is used to probe for 'id_' prefix (line 932), not to filter records. Result: count=len(state.activities) instead of count matching the WHERE conditions. Both pagination tests (test_list_activities_pagination, test_list_activities_filters_customer_and_type) assert total==1/2 but the handler returns 3, causing test failures.

**Suggestion:** In the COUNT branch (line 931-942), replace the count with: len([r for r in state.activities.values() if matches_filters(r, tenant_id=param(params,'tenant_id'), customer_id=param(params,'customer_id'), activity_type=param(params,'type'))])

.plans/issue-143.md
```diff
@@ -0,0 +1,59 @@
...
+   - Handle `select count` for activities — return `MockResult([[len(state.activities)]])` or fall back to count from `state.activities`
+   - Handle `select ... from activities` (list) — return all `MockRow` records from `state.activities` filtered by `tenant_id`
+   - Raise on invalid type for `create_activity` via `ValidationException` — the handler does not need to handle this; the service raises it before any SQL runs
+
+3. **Register `make_activity_handler` in `all_handlers()`** (`tests/unit/conftest.py`, `all_handlers` function):
+   - Append `make_activity_handler(state)` to the returned list
+
+4. **Create `tests/unit/test_activity_service.py`** with:
+   - Import `ActivityService`, `Activity`, `ActivityType` from `src`, plus `NotFoundException` and `ValidationException`
+   - `mock_db_session` fixture using `MagicMock` pattern (same as `test_customer_service.py`) so `session.execute` uses `AsyncMock(side_effect=...)`; `session.flush` returns `AsyncMock()`; `session.refresh` is a no-op
```
yihua-zhuo, line 34: **INFO** (suggestion)

**Issue:** Plan step 4 says mock_db_session uses 'MagicMock pattern (same as test_customer_service.py)', but test_activity_service.py actually uses make_mock_session([make_activity_handler(state)]) — a different pattern. The plan description does not match the implementation.

**Suggestion:** Update plan step 4 to describe the make_mock_session + make_activity_handler pattern used in the implementation.

.plans/issue-143.md
```diff
@@ -0,0 +1,59 @@
...
+   - `test_get_activity_not_found` — calls `get_activity(9999, tenant_id=1)`, expects `NotFoundException`
+   - `TestUpdateActivity` class with `test_update_activity` — pre-populate mock result for `session.execute` so that `_fetch` returns a row; call `update_activity(id=1, tenant_id=1, content="Updated content")`; assert returned activity `.content == "Updated content"`
+   - `TestDeleteActivity` class with `test_delete_activity` — pre-populate mock result for `_fetch` to find the row; call `delete_activity(id=1, tenant_id=1)`; assert returned dict has `{"id": 1}`; subsequent `get_activity(1, tenant_id=1)` raises `NotFoundException`
+   - `TestListActivitiesPagination` class with `test_list_activities_pagination` — mock `session.execute` to return two `MockRow` instances for the SELECT and `[[3]]` for the COUNT; call `list_activities(tenant_id=1, page=1, page_size=2)`; assert `items` has `len == 2` and `total == 3`
+
```
yihua-zhuo, line 42: **INFO** (suggestion)

**Issue:** Plan claims '6 new test cases' but the implementation has 10 test methods across 5 classes. The plan under-reports scope by ~4 tests (test_update_activity_with_no_fields_only_refetches, test_update_activity_rejects_invalid_type, test_update_activity_rejects_wrong_tenant, test_list_activities_filters_customer_and_type).

**Suggestion:** Update the test count and enumerate all test methods by name in the plan.

.plans/issue-143.md
```diff
@@ -0,0 +1,59 @@
+Now I have everything I need. Let me write the implementation plan.
```
yihua-zhuo, line 1: **INFO** (suggestion)

**Issue:** First-person planning narrative ('Now I have everything I need. Let me write...') is stale content that should be removed from the final plan document.

**Suggestion:** Remove the opening narrative line; start directly with the # Implementation Plan heading.

tests/unit/conftest.py
```diff
@@ -860,6 +856,116 @@ def campaign_handler(sql_text, params):
...
+        if "select" in sql_text and "count" in sql_text and "from activities" in sql_text:
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** The count 'probes' tenant_id via startswith('tenant_id_') on line 932 — this finds no SQLAlchemy-generated bind param since WHERE clause params use bare names, not underscore-suffixed names. tenant_id stays as the raw value which may be an int, not None. But param() also tries startswith('tenant_id_') which returns None (no key matches), then falls back to params['tenant_id']. This indirect resolution is fragile and the COUNT handler ignores these filters entirely when computing the count.

**Suggestion:** After fixing the COUNT filter (see critical issue), also replace param() with direct params.get() for WHERE clause filters since SQLAlchemy does not generate underscore-suffixed names for simple comparison columns.

tests/unit/test_activity_service.py
```diff
@@ -0,0 +1,224 @@
+"""Unit tests for src/services/activity_service.py."""
+
+from datetime import UTC, datetime
+
+import pytest
+
+from models.activity import ActivityType
+from pkg.errors.app_exceptions import NotFoundException, ValidationException
+from services.activity_service import ActivityService
+from tests.unit.conftest import MockState, make_activity_handler, make_mock_session
```
yihua-zhuo, line 10: **WARNING** (suggestion)

**Issue:** make_mock_session is imported from conftest but never used in this file — mock_db_session is built by calling make_mock_session at line 20, not by referencing the imported symbol.

**Suggestion:** Remove 'make_mock_session' from the import line.

tests/unit/test_activity_service.py
```diff
@@ -0,0 +1,224 @@
+"""Unit tests for src/services/activity_service.py."""
+
+from datetime import UTC, datetime
```
yihua-zhuo, line 3: **INFO** (suggestion)

**Issue:** UTC is imported from datetime but not used anywhere in the file (the add_activity helper uses datetime.now(UTC) internally, not exposed as a fixture parameter).

**Suggestion:** Remove 'UTC' from the import.

tests/unit/test_activity_service.py
```diff
@@ -0,0 +1,224 @@
...
+
+    session.refresh.side_effect = refresh_side_effect
+    return session
+
+
+@pytest.fixture
+def activity_service(mock_db_session):
+    return ActivityService(mock_db_session)
+
+
```
yihua-zhuo, line 34: **INFO** (suggestion)

**Issue:** Plan step 4 says 'mock_db_session fixture using MagicMock pattern (same as test_customer_service.py)' but the actual implementation uses make_mock_session + make_activity_handler. The plan description is misleading for future readers.

**Suggestion:** Update the plan to match the actual pattern, or align test_activity_service.py with the described MagicMock pattern if that is the preferred convention.

tests/unit/conftest.py
```diff
@@ -51,7 +51,7 @@ def __init__(self, mapping):
             if "tags" in self._mapping and isinstance(self._mapping["tags"], str):
                 try:
                     self._mapping["tags"] = _json.loads(self._mapping["tags"])
-                except Exception:
+                except Exception:  # noqa: S110 - invalid JSON should leave the original value intact.
```
yihua-zhuo, line 53: **WARNING** (suggestion)

**Issue:** Bare `except:` clause catches all exceptions including `KeyboardInterrupt` and `SystemExit`; catch only `json.JSONDecodeError` or at minimum `Exception`.

**Suggestion:** Change `except Exception:` to `except json.JSONDecodeError:` since that's the only exception raised in the try block.

tests/unit/conftest.py
```diff
@@ -860,6 +856,121 @@ def campaign_handler(sql_text, params):
...
+        activity_id = param(params, "id")
+        if "where" in sql_text and "activities.id" in sql_text and activity_id is not None:
+            tenant_id = param(params, "tenant_id")
+            if activity_id in state.activities and matches_filters(state.activities[activity_id], tenant_id=tenant_id):
+                return MockResult([MockRow(state.activities[activity_id].copy())])
+            return MockResult([])
+
+        # SELECT list
+        if state.activities:
+            tenant_id = param(params, "tenant_id", 0)
```
yihua-zhuo: **INFO** (suggestion)

**Issue:** Inconsistent `tenant_id` access within `make_activity_handler`: the COUNT query uses `param(params, "tenant_id")` (line 932) which handles prefixed binds, but the SELECT-list query uses `params.get("tenant_id", 0)` (line 955) which does not. If SQLAlchemy generates a prefixed bind like `tenant_id_1`, the COUNT path finds it but the list path misses it.

**Suggestion:** Use `param(params, "tenant_id", 0)` consistently in the SELECT-list block (line 955) to match the COUNT block's behavior.

tests/unit/test_activity_service.py
```diff
@@ -0,0 +1,230 @@
...
+    @pytest.mark.asyncio
+    async def test_update_activity_with_no_fields_only_refetches(self, activity_service, activity_state, mock_db_session):
+        """update_activity with no fields skips UPDATE and returns current activity."""
+        add_activity(activity_state, activity_id=1, tenant_id=1, content="Unchanged")
+
+        result = await activity_service.update_activity(activity_id=1, tenant_id=1)
+
+        assert result.content == "Unchanged"
+        assert mock_db_session.execute.await_count == 2
```
yihua-zhuo, line 146: **INFO** (suggestion)

**Issue:** `mock_db_session.execute.await_count == 2` asserts only call count, not which SQL was executed. If the service accidentally calls two different queries instead of a fetch-by-id + a no-op, the assertion still passes.

**Suggestion:** Add a spec check or tag to the execute side-effect so individual calls can be distinguished, or verify the returned row's `id` matches the expected activity.

scripts/ci/fix_pr_comments.py
```diff
@@ -45,6 +52,48 @@ def run_gh(args: list[str]) -> subprocess.CompletedProcess[str]:
     return subprocess.run(["gh", *args], capture_output=True, text=True, check=False)
```
yihua-zhuo, line 53: **CRITICAL** (critical)

**Issue:** Bare 'except:' in MockRow.__init__ swallows any JSON decode error when parsing 'tags', hiding data corruption silently.

**Suggestion:** Use 'except (json.JSONDecodeError, ValueError, TypeError):' so malformed tags are explicitly handled rather than silently ignored.

tests/unit/conftest.py
```diff
@@ -956,9 +246,15 @@ def make_mock_session(handlers=None):
 
     def _execute_side_effect(sql, params=None):
         sql_text = str(sql).lower().strip()
-        params = params or {}
+        bound_params = {}
+        try:
+            bound_params.update(getattr(sql.compile(), "params", {}) or {})
+        except Exception as exc:  # noqa: BLE001 - some SQLAlchemy test doubles are not compilable.
```
yihua-zhuo, line 252: **CRITICAL** (critical)

**Issue:** Bare 'except:' in _execute_side_effect (inside make_mock_session) silently swallows all SQL compilation errors, populating bound_params with a '_compile_error' key and proceeding with incorrect params — a query binding mismatch would silently pass the test while failing in production.

**Suggestion:** Use 'except Exception as exc:' and log or propagate it — Rule 25 and Rule 46 prohibit silent exception swallowing on data-layer operations. A mis-bound test that passes silently is worse than a noisy failure.

tests/unit/domain_handlers/activities.py
```diff
@@ -0,0 +1,124 @@
...
+
+from __future__ import annotations
+
+from tests.unit.conftest import MockResult, MockRow, MockState
+
+ORDER = 70
+
+
+def param(params, name, default=None):
+    if name in params:
```
yihua-zhuo, line 11: **CRITICAL** (critical)

**Issue:** param() function's logic is inverted: it checks the direct key first (returning it if present) before trying the prefix search, so if SQL bind keys and prefix-search candidates both exist in params (e.g., both 'id' and 'id_activities' are present), the wrong key wins. More critically, if the exact key 'id' exists but SQLAlchemy names the compiled bind 'activities.id' for a qualified-column WHERE clause, param() would return 'activities.id' as the ID value and the subsequent 'if activity_id in state.activities' (integer-keyed dict) would always be False, breaking get/update/delete handlers silently.

**Suggestion:** Rewrite param() to only use prefix search, or rename it to reflect its actual semantics and add a direct-lookup variant. For the handlers, prefer explicit params.get('id') over param() for the primary entity ID since that is the most critical binding contract per Rule 133.

tests/unit/domain_handlers/campaigns.py
```diff
@@ -0,0 +1,25 @@
...
+ORDER = 60
+
+
+def campaign_handler(sql_text, params):
+    """Handle campaign-related SQL."""
+    if "insert into campaigns" in sql_text:
+        return MockResult([MockRow({"id": 1, "tenant_id": params.get("tenant_id", 0), "name": params.get("name", "Campaign"), "campaign_type": params.get("campaign_type", "email"), "status": "draft", "created_by": params.get("created_by"), "created_at": params.get("created_at"), "updated_at": params.get("updated_at")})])
+
+    if "from campaigns" in sql_text:
+        return MockResult([MockRow({"id": 1, "tenant_id": 1, "name": "Campaign A", "campaign_type": "email", "status": "draft", "created_by": 1, "created_at": None, "updated_at": None})])
```
yihua-zhuo, line 16: **CRITICAL** (critical)

**Issue:** The FROM campaigns SELECT returns hardcoded tenant_id: 1 regardless of params['tenant_id'] — violates Rule 134 (tenant isolation semantics).

**Suggestion:** Return tenant_id from params.get("tenant_id", 0) instead of hardcoding 1.

tests/unit/domain_handlers/customers.py
```diff
@@ -0,0 +1,158 @@
...
+                                "tags": _json.dumps(params.get("tags", [])),
+                                "created_at": None,
+                                "updated_at": None,
+                            }
+                        )
+                    ]
+                )
+            return MockResult([])
+
+        if sql_text.startswith("delete") and "customers" in sql_text:
```
yihua-zhuo, line 67: **CRITICAL** (critical)

**Issue:** DELETE handler does not check tenant_id before returning — any tenant can delete any customer id, violating Rule 134 (tenant isolation).

**Suggestion:** Add tenant_id check: if params.get('tenant_id') not in (rec.get('tenant_id'), ...): return MockResult([]) before the delete.

tests/unit/domain_handlers/sales.py
```diff
@@ -0,0 +1,36 @@
...
+
+ORDER = 40
+
+
+def pipeline_handler(sql_text, params):
+    """Handle pipeline-related SQL."""
+    if "insert into pipelines" in sql_text:
+        return MockResult([MockRow({"id": 1, "tenant_id": params.get("tenant_id", 0), "name": params.get("name", "Pipeline"), "description": params.get("description"), "is_active": True, "created_at": params.get("created_at"), "updated_at": params.get("updated_at")})])
+
+    if "from pipelines" in sql_text:
```
yihua-zhuo, line 15: **CRITICAL** (critical)

**Issue:** FROM pipelines SELECT returns hardcoded tenant_id: 1 regardless of params['tenant_id'] — violates Rule 134 (tenant isolation semantics).

**Suggestion:** Return tenant_id from params.get("tenant_id", 0) instead of hardcoding 1.

tests/unit/domain_handlers/sales.py
```diff
@@ -0,0 +1,36 @@
...
+
+    return None
+
+
+def opportunity_handler(sql_text, params):
+    """Handle opportunity-related SQL."""
+    if "insert into opportunities" in sql_text:
+        return MockResult([MockRow({"id": 1, "tenant_id": params.get("tenant_id", 0), "customer_id": 1, "title": params.get("title", "Opportunity"), "amount": params.get("amount", 1000), "stage": "qualification", "probability": 20, "owner_id": params.get("owner_id", 0), "expected_close_date": params.get("expected_close_date"), "created_at": params.get("created_at"), "updated_at": params.get("updated_at")})])
+
+    if "from opportunities" in sql_text:
```
yihua-zhuo, line 26: **CRITICAL** (critical)

**Issue:** FROM opportunities SELECT returns hardcoded tenant_id: 1 regardless of params['tenant_id'] — violates Rule 134 (tenant isolation semantics).

**Suggestion:** Return tenant_id from params.get("tenant_id", 0) instead of hardcoding 1.

tests/unit/domain_handlers/tenants.py
```diff
@@ -0,0 +1,56 @@
...
+    if "update" in sql_text and "tenants" in sql_text:
+        tenant_id = params.get("tenant_id") or params.get("id")
+        if tenant_id == 1:
+            return MockResult([[1, "Updated Name", "pro", "active", "{}", None, None]])
+        return MockResult([])
+
+    if "select" in sql_text and "count" in sql_text and "from tenants" in sql_text:
+        return MockResult([[2]])
+
+    if "select" in sql_text and "from tenants" in sql_text and "count" not in sql_text:
```
yihua-zhuo, line 30: **CRITICAL** (critical)

**Issue:** Bulk SELECT handler returns all tenants regardless of tenant_id filter, enabling cross-tenant data exposure if a query omits the WHERE tenant_id clause.

**Suggestion:** Filter returned rows by tenant_id: add `if rec.get('tenant_id') != tenant_filter: continue` inside the loop at line 40, and remove the unconditional fixture appends at lines 41-43.

tests/unit/domain_handlers/tenants.py
```diff
@@ -0,0 +1,56 @@
...
+    if "insert into tenants" in sql_text:
+        return MockResult([MockRow({"id": 1, "name": params.get("name"), "plan": params.get("plan"), "status": "active", "settings": params.get("settings", "{}"), "created_at": params.get("now"), "updated_at": params.get("now")})])
+
+    if "delete" in sql_text and "tenants" in sql_text:
+        tenant_id = params.get("tenant_id")
+        if tenant_id and tenant_id != 1:
+            return MockResult([])
+        return MockResult([[1, "Deleted Tenant", "pro", "deleted", "{}", None, None]])
+
+    if "update" in sql_text and "tenants" in sql_text:
```
yihua-zhuo, line 21: **CRITICAL** (critical)

**Issue:** UPDATE handler checks tenant_id but returns a fake row for tenant 1 without verifying the original row's tenant_id matches the request, breaking cross-tenant isolation.

**Suggestion:** After line 23, when tenant_id == 1, fetch the actual record and verify its tenant_id field before returning; do not silently return a hardcoded fake row.

tests/unit/domain_handlers/tickets.py
```diff
@@ -0,0 +1,25 @@
...
+ORDER = 50
+
+
+def ticket_sql_handler(sql_text, params):
+    """Handle ticket-related SQL."""
+    if "insert into tickets" in sql_text:
+        return MockResult([MockRow({"id": 1, "tenant_id": params.get("tenant_id", 0), "subject": params.get("subject", "Ticket"), "description": params.get("description"), "status": "open", "priority": params.get("priority", "medium"), "customer_id": 1, "assignee_id": params.get("assignee_id"), "created_at": params.get("created_at"), "updated_at": params.get("updated_at")})])
+
+    if "from tickets" in sql_text:
+        return MockResult([MockRow({"id": 1, "tenant_id": 1, "subject": "Issue A", "description": "Desc", "status": "open", "priority": "medium", "customer_id": 1, "assignee_id": 1, "created_at": None, "updated_at": None})])
```
yihua-zhuo, line 16: **CRITICAL** (critical)

**Issue:** ticket_sql_handler ignores tenant_id on SELECT/UPDATE/DELETE and always returns a hardcoded fixture for tenant_id=1, creating cross-tenant data leakage if any production code drops the tenant filter.

**Suggestion:** Add `if params.get('tenant_id', 1) != 1: return MockResult([])` at the top of the 'from tickets' branch, or look up the record in a state dict keyed by id and verify tenant_id matches.

tests/unit/domain_handlers/tickets.py
```diff
@@ -0,0 +1,25 @@
...
+from __future__ import annotations
+
+from tests.unit.conftest import MockResult, MockRow, MockState
+
+ORDER = 50
+
+
+def ticket_sql_handler(sql_text, params):
+    """Handle ticket-related SQL."""
+    if "insert into tickets" in sql_text:
```
yihua-zhuo, line 12: **CRITICAL** (critical)

**Issue:** INSERT handler ignores tenant_id and always sets customer_id=1 in the returned row, masking bugs where callers pass wrong customer or tenant values.

**Suggestion:** Use `params.get('tenant_id')` and `params.get('customer_id')` for the returned record columns instead of hardcoded values.

tests/unit/domain_handlers/users.py
```diff
@@ -0,0 +1,130 @@
...
+            was_in_mock = user_id in state.users
+            if was_in_mock:
+                state.users[user_id]["status"] = "deleted"
+                del state.users[user_id]
+            state.deleted_user_ids.add(user_id)
+            if was_in_mock:
+                return MockResult([MockRow({"id": user_id})])
+            return MockResult([])
+
+        if ("set " in sql_text or sql_text.startswith("update")) and "where id" in sql_text:
```
yihua-zhuo, line 50: **CRITICAL** (critical)

**Issue:** UPDATE handler (line 50) updates and returns a user by id without checking tenant_id, breaking Rule 134 — a user belonging to tenant B could be mutated by a request scoped to tenant A if the id happens to collide.

**Suggestion:** Add tenant_id verification after line 52: `if state.users[user_id].get('tenant_id') != params.get('tenant_id'): return MockResult([])` before mutating the record.

tests/unit/domain_handlers/users.py
```diff
@@ -0,0 +1,130 @@
...
+            }
+            if user_id in fixtures:
+                row = fixtures[user_id].copy()
+                now = dt.utcnow()
+                row["created_at"] = now
+                row["updated_at"] = now
+                return MockResult([MockRow(row)])
+            return MockResult([])
+
+        if state.users:
```
yihua-zhuo, line 113: **CRITICAL** (critical)

**Issue:** Fallback SELECT returns all users in state regardless of tenant_id when the query has no filter clauses, violating Rule 134 — tests for tenant B will silently receive tenant A's users.

**Suggestion:** Filter rows by tenant_id: `tenant_filter = params.get('tenant_id'); rows = [MockRow(r.copy()) for r in state.users.values() if r.get('tenant_id') == tenant_filter]`.

tests/unit/test_activity_service.py
```diff
@@ -0,0 +1,254 @@
...
+            activity_type="call",
+            content="Other tenant",
+            created_at=datetime(2025, 1, 1, 13, 0),
+        )
+
+        items, total = await activity_service.list_activities(tenant_id=1, page=1, page_size=2)
+
+        assert len(items) == 2
+        assert total == 3
+        assert {item.id for item in items} == {2, 3}
```
yihua-zhuo, line 238: **CRITICAL** (critical)

**Issue:** test_list_activities_pagination asserts {item.id for item in items} == {2, 3}, which depends on the mock created_at ordering — id order and created_at order differ in the seed data, making the test fragile and incorrect if the handler sort is ever fixed.

**Suggestion:** Seed activities with non-sequential created_at values that produce the expected order explicitly, e.g. activity 1: 2025-01-01 12:00, activity 2: 2025-01-01 10:00, activity 3: 2025-01-01 08:00, then assert the chronological ordering by id or content to match the actual created_at sort.

.plans/issue-143.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Test Plan
+
+- **Unit tests in `tests/unit/`**: `tests/unit/test_activity_service.py` — 6 new test cases covering `create_activity`, `get_activity` (found + not found), `update_activity`, `delete_activity`, and `list_activities` pagination
+- **Integration tests in `tests/integration/`**: No integration tests required for this task
+
+## Acceptance Criteria
+
+- `pytest tests/unit/test_activity_service.py --collect-only` reports exactly 6 collected items with 0 errors
+- `pytest tests/unit/test_activity_service.py -v` — all 6 tests pass
```
yihua-zhuo, line 51: **INFO** (suggestion)

**Issue:** Plan acceptance criteria uses ruff command 'ruff check src/' but the project convention (per CLAUDE.md) is 'ruff check src/' — the difference is cosmetic here but acceptance criteria should match exact commands developers run.

**Suggestion:** Keep 'ruff check src/' in acceptance criteria to match CLAUDE.md and the pre-push hook.

.plans/issue-143.md
```diff
@@ -0,0 +1,59 @@
...
+
+# Implementation Plan — Issue #143
+
+## Goal
+
+Add a `tests/unit/test_activity_service.py` unit test suite for `ActivityService` covering all core CRUD operations and pagination, matching the patterns used by existing unit tests in the project.
+
+## Affected Files
+
+- `tests/unit/conftest.py` — add `activities` dict + `activities_next_id` to `MockState`, add `make_activity_handler(state)` factory, register it in `all_handlers()`
```
yihua-zhuo, line 13: **WARNING** (suggestion)

**Issue:** Plan directs edits to tests/unit/conftest.py, a central registry file — Rule 125 prohibits routine domain additions to central registry files.

**Suggestion:** Per Rule 125, MockState state fields and handler registrations should live in tests/unit/domain_handlers/activities.py via get_handlers(state), and the test file should compose only what it needs without editing conftest.py. The plan's approach (editing conftest.py) violates the project convention established after this plan was written.

yihua-zhuo, line 13: **WARNING** (suggestion)

**Issue:** Plan does not mention alembic/versions/ migration per Rule 129 — if issue #143 touches ORM models, a migration must be reviewed.

**Suggestion:** If the issue modifies any tenant-owned ORM model, the plan should include an alembic migration review step.

tests/unit/conftest.py
```diff
@@ -906,20 +200,10 @@ def _domain_handlers(state: MockState):
 
 def all_handlers(state: MockState):
```
yihua-zhuo, line 201: **INFO** (suggestion)

**Issue:** all_handlers(state) returns _domain_handlers(state) but the function accepts 'state' as a required param yet there is no pytest fixture that calls it with the state argument — tests that rely on the default all-handlers composition may not have their state properly scoped.

**Suggestion:** Add a module-level or pytest-accessible 'all_handlers_fixture' fixture that returns all_handlers(MockState()) so tests using the full mock can share typed state.

tests/unit/domain_handlers/activities.py
```diff
@@ -0,0 +1,124 @@
...
+        return params[name]
+    prefix = f"{name}_"
+    for key, value in params.items():
+        if key.startswith(prefix):
+            return value
+    return default
+
+
+def matches_filters(record, tenant_id=None, customer_id=None, activity_type=None):
```
yihua-zhuo, line 20: **INFO** (suggestion)

**Issue:** matches_filters receives tenant_id=None in the COUNT branch when param(params,'tenant_id') returns None, then does 'record.get("tenant_id") != None' which is a !=-vs-is-not comparison — this is a stylistic issue; the semantic intent (filtering when tenant_id is not None) is correct but the comparison should use 'is not' for None identity checks per Python conventions.

**Suggestion:** Use 'is not None' for None identity comparisons: 'if tenant_id is not None and record.get("tenant_id") != tenant_id'.

tests/unit/domain_handlers/activities.py
```diff
@@ -0,0 +1,124 @@
...
+        )
+        if not is_activity_query:
+            return None
+
+        if sql_text.startswith("update") and "activities" in sql_text:
+            activity_id = param(params, "id")
+            tenant_id = param(params, "tenant_id")
+            if activity_id in state.activities:
+                rec = state.activities[activity_id]
+                if tenant_id is not None and rec.get("tenant_id") != tenant_id:
```
yihua-zhuo, line 63: **INFO** (suggestion)

**Issue:** In the UPDATE handler, 'if tenant_id is not None and rec.get("tenant_id") != tenant_id' compares an int from params against an int from the record — this is fine for int values, but the 'is not None' style is inconsistently applied vs line 21 where '!= None' (identity) is used.

**Suggestion:** Normalize to 'is not None' throughout matches_filters and the handler for consistent Python idiom.

tests/unit/domain_handlers/campaigns.py
```diff
@@ -0,0 +1,25 @@
+"""Campaign SQL handlers for unit tests."""
+
+from __future__ import annotations
+
+from tests.unit.conftest import MockResult, MockRow, MockState
+
+ORDER = 60
+
+
+def campaign_handler(sql_text, params):
```
yihua-zhuo, line 10: **WARNING** (suggestion)

**Issue:** campaign_handler is stateless — it receives state: MockState but never uses it, preventing tests from seeding custom campaign data.

**Suggestion:** Capture state in the closure: handler = lambda sql, params, _state=state: ...

tests/unit/domain_handlers/counts.py
```diff
@@ -0,0 +1,27 @@
...
+ORDER = 1000
+
+
+def make_count_handler(state: MockState):
+    """Fallback COUNT handler for queries not caught by domain handlers."""
+
+    def handler(sql_text, params):
+        if "select" not in sql_text or "count" not in sql_text:
+            return None
+        tenant_id = params.get("tenant_id")
```
yihua-zhuo, line 16: **INFO** (suggestion)

**Issue:** Count values (7 for tenant_id=1, 3 otherwise) are hardcoded rather than derived from actual state, so count tests may be inaccurate if state changes.

**Suggestion:** Use len([r for r in state.customers.values() if r.get('tenant_id') == tenant_id]) when state is available.

tests/unit/domain_handlers/customers.py
```diff
@@ -0,0 +1,158 @@
...
+            return MockResult([MockRow({"id": params.get("id", 1)})])
+
+        if "select" in sql_text and "count" in sql_text and "from customers" in sql_text:
+            tenant_id = params.get("tenant_id")
```
yihua-zhuo, line 71: **INFO** (suggestion)

**Issue:** COUNT returns hardcoded 3/7 instead of reflecting actual state.customers, so tests for customer count will be inaccurate if state changes.

**Suggestion:** Return len([r for r in state.customers.values() if r.get('tenant_id') == tenant_id]) for accuracy.

tests/unit/domain_handlers/customers.py
```diff
@@ -0,0 +1,158 @@
...
+                            "status": "customer",
+                            "owner_id": 1,
+                            "tags": "[]",
+                            "created_at": None,
+                            "updated_at": None,
+                        }
+                    ),
+                ]
+            )
+            if tenant_filter == 2:
```
yihua-zhuo, line 145: **WARNING** (suggestion)

**Issue:** Special-case block for tenant_id==2 clears all rows before appending fixtures, hiding cross-tenant bugs — if a test accidentally queries tenant 1 data while in a 'tenant 2' context, it would silently pass.

**Suggestion:** Remove the tenant-filter bypass and ensure list SELECT always enforces tenant_id from params.

tests/unit/domain_handlers/sales.py
```diff
@@ -0,0 +1,36 @@
+"""Sales SQL handlers for unit tests."""
+
+from __future__ import annotations
+
+from tests.unit.conftest import MockResult, MockRow, MockState
+
+ORDER = 40
+
+
+def pipeline_handler(sql_text, params):
```
yihua-zhuo, line 10: **WARNING** (suggestion)

**Issue:** pipeline_handler is stateless — it receives state: MockState but never uses it, preventing tests from seeding custom pipeline data.

**Suggestion:** Capture state in the closure so tests can seed records via state.pipelines.

tests/unit/domain_handlers/sales.py
```diff
@@ -0,0 +1,36 @@
...
+    if "insert into pipelines" in sql_text:
+        return MockResult([MockRow({"id": 1, "tenant_id": params.get("tenant_id", 0), "name": params.get("name", "Pipeline"), "description": params.get("description"), "is_active": True, "created_at": params.get("created_at"), "updated_at": params.get("updated_at")})])
+
+    if "from pipelines" in sql_text:
+        return MockResult([MockRow({"id": 1, "tenant_id": 1, "name": "Pipeline A", "description": "Desc", "is_active": True, "created_at": None, "updated_at": None})])
+
+    return None
+
+
+def opportunity_handler(sql_text, params):
```
yihua-zhuo, line 21: **WARNING** (suggestion)

**Issue:** opportunity_handler is stateless — it receives state: MockState but never uses it, preventing tests from seeding custom opportunity data.

**Suggestion:** Capture state in the closure so tests can seed records via state.opportunities.

tests/unit/domain_handlers/tenants.py
```diff
@@ -0,0 +1,56 @@
...
+ORDER = 30
+
+
+def tenant_handler(sql_text, params):
+    """Handle all tenant-related SQL."""
+    if "insert into tenants" in sql_text:
+        return MockResult([MockRow({"id": 1, "name": params.get("name"), "plan": params.get("plan"), "status": "active", "settings": params.get("settings", "{}"), "created_at": params.get("now"), "updated_at": params.get("now")})])
+
+    if "delete" in sql_text and "tenants" in sql_text:
+        tenant_id = params.get("tenant_id")
```
yihua-zhuo, line 16: **WARNING** (suggestion)

**Issue:** DELETE handler only rejects deletes for tenant_id values other than 1; deletes for tenant 2 succeed silently, masking cross-tenant deletion bugs.

**Suggestion:** Look up the activity by id first, verify its tenant_id matches the request, and return empty on mismatch instead of delegating to the global permissive fallback.

tests/unit/domain_handlers/tenants.py
```diff
@@ -0,0 +1,56 @@
...
+
+from tests.unit.conftest import MockResult, MockRow, MockState
+
+ORDER = 30
+
+
+def tenant_handler(sql_text, params):
+    """Handle all tenant-related SQL."""
+    if "insert into tenants" in sql_text:
+        return MockResult([MockRow({"id": 1, "name": params.get("name"), "plan": params.get("plan"), "status": "active", "settings": params.get("settings", "{}"), "created_at": params.get("now"), "updated_at": params.get("now")})])
```
yihua-zhuo, line 13: **WARNING** (suggestion)

**Issue:** INSERT returns a hardcoded 'active' status regardless of the 'plan' value, hiding bugs in plan-based logic.

**Suggestion:** Return the actual status from params if provided, e.g. `params.get('status', 'active')`, to faithfully reflect what the ORM would produce.

tests/unit/domain_handlers/tenants.py
```diff
@@ -0,0 +1,56 @@
...
+            return MockResult([])
+        return MockResult([[1, "Deleted Tenant", "pro", "deleted", "{}", None, None]])
+
+    if "update" in sql_text and "tenants" in sql_text:
+        tenant_id = params.get("tenant_id") or params.get("id")
+        if tenant_id == 1:
+            return MockResult([[1, "Updated Name", "pro", "active", "{}", None, None]])
+        return MockResult([])
+
+    if "select" in sql_text and "count" in sql_text and "from tenants" in sql_text:
```
yihua-zhuo, line 27: **INFO** (suggestion)

**Issue:** COUNT handler returns a fixed 2 without filtering by tenant_id, making the count unreliable for tenant-specific assertions.

**Suggestion:** Add tenant_id filtering to the COUNT response, or at minimum return different fixed values for different tenant_id values (e.g. tenant_id == 1 → 2, tenant_id == 2 → 1).

tests/unit/domain_handlers/tickets.py
```diff
@@ -0,0 +1,25 @@
+"""Ticket SQL handlers for unit tests."""
+
+from __future__ import annotations
+
+from tests.unit.conftest import MockResult, MockRow, MockState
+
+ORDER = 50
+
+
+def ticket_sql_handler(sql_text, params):
```
yihua-zhuo, line 10: **WARNING** (suggestion)

**Issue:** handler is a plain function (not make_* factory), so it cannot use per-test MockState — all tickets converge on the same static fixture, hiding stateful test interactions.

**Suggestion:** Rename to make_ticket_handler(state) following the make_* factory pattern used by users.py, customers.py, and activities.py, so tests can seed records via state.

tests/unit/domain_handlers/tickets.py
yihua-zhuo, line 12: **INFO** (suggestion)

**Issue:** INSERT returns hardcoded 'open' status and 'medium' priority instead of reading from params, which could mask validation or enum coercion issues.

**Suggestion:** Read `status` and `priority` from params: `status=params.get('status', 'open')`, `priority=params.get('priority', 'medium')`.

tests/unit/domain_handlers/users.py
```diff
@@ -0,0 +1,130 @@
...
+                "status": "pending",
+                "full_name": params.get("full_name"),
+                "bio": None,
+                "created_at": params.get("created_at"),
+                "updated_at": params.get("created_at"),
+            }
+            state.users[uid] = record
+            return MockResult([MockRow(record.copy())])
+
+        is_user_query = "from users" in sql_text or (sql_text.startswith("update") and "users" in sql_text)
```
yihua-zhuo, line 35: **WARNING** (suggestion)

**Issue:** The UPDATE guard `sql_text.startswith('update') and 'users' in sql_text` may not match ORM-generated SQL which uses aliases or multi-table contexts, causing the handler to silently return None and the update to succeed on a fallback handler.

**Suggestion:** Use `'update users' in sql_text or ('update' in sql_text and 'users' in sql_text and 'set' in sql_text)` as the guard for robustness.

tests/unit/domain_handlers/users.py
```diff
@@ -0,0 +1,130 @@
...
+        if ("set " in sql_text or sql_text.startswith("update")) and "where id" in sql_text:
+            user_id = params.get("id")
+            if user_id in state.users:
+                for k, v in params.items():
+                    if k not in ("id", "tenant_id"):
+                        state.users[user_id][k] = v
+                return MockResult([MockRow(state.users[user_id].copy())])
+            return MockResult([])
+
+        if "select" in sql_text and "count" in sql_text and "from users" in sql_text:
```
yihua-zhuo, line 59: **INFO** (suggestion)

**Issue:** COUNT handler ignores tenant_id when state.users is non-empty, returning the total count across all tenants instead of per-tenant counts.

**Suggestion:** Filter by tenant_id: `count_val = sum(1 for r in state.users.values() if r.get('tenant_id') == params.get('tenant_id'))`.

tests/unit/test_activity_service.py
```diff
@@ -0,0 +1,254 @@
...
+        add_activity(activity_state, activity_id=3, tenant_id=1, customer_id=2, activity_type="call")
+
+        items, total = await activity_service.list_activities(
+            tenant_id=1,
+            customer_id=1,
+            activity_type="call",
+        )
+
+        assert total == 1
+        assert [item.id for item in items] == [1]
```
yihua-zhuo, line 254: **WARNING** (suggestion)

**Issue:** test_list_activities_filters_customer_and_type asserts total == 1 but the mock count handler returns a fixed value (7 or 3 by tenant_id), unrelated to the seeded filter state — the test passes by accident.

**Suggestion:** Replace the count handler in the fixture with a tenant-specific handler that counts only matching records, or remove the total assertion and assert only on item count.

tests/unit/test_activity_service.py
```diff
@@ -0,0 +1,254 @@
...
+from tests.unit.conftest import MockState, make_activity_handler, make_mock_session
+
+
+@pytest.fixture
+def activity_state():
+    return MockState()
+
+
+@pytest.fixture
+def mock_db_session(activity_state):
```
yihua-zhuo, line 19: **INFO** (suggestion)

**Issue:** mock_db_session is built with only make_activity_handler but the broader test suite loads all domain handlers via conftest._domain_handlers, making isolation dependent on handler ordering rather than explicit composition.

**Suggestion:** Pass only `[make_activity_handler(activity_state)]` to make_mock_session and confirm no other handlers run for the domain under test, to guarantee true isolation per Rule 135.

tests/unit/domain_handlers/campaigns.py
```diff
@@ -0,0 +1,25 @@
...
+    """Handle campaign-related SQL."""
+    if "insert into campaigns" in sql_text:
+        return MockResult([MockRow({"id": 1, "tenant_id": params.get("tenant_id", 0), "name": params.get("name", "Campaign"), "campaign_type": params.get("campaign_type", "email"), "status": "draft", "created_by": params.get("created_by"), "created_at": params.get("created_at"), "updated_at": params.get("updated_at")})])
+
+    if "from campaigns" in sql_text:
```
yihua-zhuo, line 15: **CRITICAL** (critical)

**Issue:** Rule 134: SELECT handler returns hardcoded tenant-1 fixture regardless of the requested tenant_id — cross-tenant data can bleed through.

**Suggestion:** Inspect params['tenant_id'] and return fixture data only when it matches the record's tenant_id, otherwise return MockResult([]).

tests/unit/domain_handlers/customers.py
```diff
@@ -0,0 +1,158 @@
...
+                "status": params.get("status", "lead"),
+                "owner_id": params.get("owner_id", 0),
+                "tags": tags_val,
+                "created_at": params.get("created_at"),
+                "updated_at": params.get("updated_at"),
+            }
+            state.customers[cid] = record
+            return MockResult([MockRow(record.copy())])
+
+        if sql_text.startswith("update") and "customers" in sql_text:
```
yihua-zhuo, line 37: **CRITICAL** (critical)

**Issue:** Rule 134: UPDATE handler modifies state.customers[customer_id] without checking that the record's tenant_id matches the requested tenant_id — a tenant can update another tenant's customer.

**Suggestion:** Add tenant check: retrieve rec = state.customers[customer_id], verify rec.get('tenant_id') == params.get('tenant_id') and return MockResult([]) on mismatch.

tests/unit/domain_handlers/customers.py
```diff
@@ -0,0 +1,158 @@
...
+            return MockResult([[count_val]])
+
+        if "select" in sql_text and "from customers" in sql_text and "tags" in sql_text and "where id" in sql_text:
+            customer_id = params.get("id")
+            if customer_id in state.customers:
+                c = state.customers[customer_id]
+                return MockResult([MockRow({"id": customer_id, "tags": c.get("tags", "[]")})])
+            return MockResult([])
+
+        if "from customers where id" in sql_text:
```
yihua-zhuo, line 82: **CRITICAL** (critical)

**Issue:** Rule 134: Single-customer SELECT 'from customers where id' path does not check tenant_id against the record's tenant_id — a tenant can read another tenant's customer by ID.

**Suggestion:** After fetching state.customers[customer_id], verify its tenant_id matches params['tenant_id'] and return MockResult([]) on mismatch.

tests/unit/domain_handlers/customers.py
```diff
@@ -0,0 +1,158 @@
...
+                    "tags": "[]",
+                    "created_at": None,
+                    "updated_at": None,
+                },
+            }
+            if customer_id in fixtures:
+                return MockResult([MockRow(fixtures[customer_id].copy())])
+            return MockResult([])
+
+        if "select" in sql_text and "from customers" in sql_text and "where id" not in sql_text:
```
yihua-zhuo, line 105: **CRITICAL** (critical)

**Issue:** Rule 134: List SELECT path appends hardcoded fixture customers (IDs 1, 2) with the requested tenant_id applied, polluting any tenant's dataset with fixture data regardless of what is actually seeded.

**Suggestion:** Remove the hardcoded fixture append. Return only records from state.customers filtered by tenant_id. If test fixtures are needed, seed them into state.customers directly.

tests/unit/domain_handlers/sales.py
```diff
@@ -0,0 +1,36 @@
...
+ORDER = 40
+
+
+def pipeline_handler(sql_text, params):
+    """Handle pipeline-related SQL."""
+    if "insert into pipelines" in sql_text:
+        return MockResult([MockRow({"id": 1, "tenant_id": params.get("tenant_id", 0), "name": params.get("name", "Pipeline"), "description": params.get("description"), "is_active": True, "created_at": params.get("created_at"), "updated_at": params.get("updated_at")})])
+
+    if "from pipelines" in sql_text:
+        return MockResult([MockRow({"id": 1, "tenant_id": 1, "name": "Pipeline A", "description": "Desc", "is_active": True, "created_at": None, "updated_at": None})])
```
yihua-zhuo, line 16: **CRITICAL** (critical)

**Issue:** pipeline_handler SELECT returns hardcoded tenant_id=1 without checking params['tenant_id'] — violates Rule 134 (tenant isolation on SELECT).

**Suggestion:** Read params['tenant_id'] and filter/assert the returned row matches it, or return MockResult([]) for a mismatched tenant.

tests/unit/domain_handlers/sales.py
```diff
@@ -0,0 +1,36 @@
...
+    return None
+
+
+def opportunity_handler(sql_text, params):
+    """Handle opportunity-related SQL."""
+    if "insert into opportunities" in sql_text:
+        return MockResult([MockRow({"id": 1, "tenant_id": params.get("tenant_id", 0), "customer_id": 1, "title": params.get("title", "Opportunity"), "amount": params.get("amount", 1000), "stage": "qualification", "probability": 20, "owner_id": params.get("owner_id", 0), "expected_close_date": params.get("expected_close_date"), "created_at": params.get("created_at"), "updated_at": params.get("updated_at")})])
+
+    if "from opportunities" in sql_text:
+        return MockResult([MockRow({"id": 1, "tenant_id": 1, "customer_id": 1, "title": "Opportunity A", "amount": 1000, "stage": "qualification", "probability": 20, "owner_id": 1, "expected_close_date": None, "created_at": None, "updated_at": None})])
```
yihua-zhuo, line 27: **CRITICAL** (critical)

**Issue:** opportunity_handler SELECT returns hardcoded tenant_id=1 without checking params['tenant_id'] — violates Rule 134 (tenant isolation on SELECT).

**Suggestion:** Read params['tenant_id'] and filter/assert the returned row matches it, or return MockResult([]) for a mismatched tenant.

tests/unit/domain_handlers/tickets.py
```diff
@@ -0,0 +1,25 @@
...
+    """Handle ticket-related SQL."""
+    if "insert into tickets" in sql_text:
+        return MockResult([MockRow({"id": 1, "tenant_id": params.get("tenant_id", 0), "subject": params.get("subject", "Ticket"), "description": params.get("description"), "status": "open", "priority": params.get("priority", "medium"), "customer_id": 1, "assignee_id": params.get("assignee_id"), "created_at": params.get("created_at"), "updated_at": params.get("updated_at")})])
+
+    if "from tickets" in sql_text:
```
yihua-zhuo, line 15: **CRITICAL** (critical)

**Issue:** SELECT returns a stale hardcoded row regardless of which tenant is making the request — tests for non-tenant-1 access will silently use tenant-1 data.

**Suggestion:** Add tenant_id check before returning the fixture row, or return MockResult([]) when the tenant doesn't match.

tests/unit/domain_handlers/users.py
```diff
@@ -0,0 +1,130 @@
...
+            return MockResult([[count_val]])
+
+        if "where username" in sql_text:
+            username = params.get("username")
+            for rec in state.users.values():
+                if rec.get("username") == username:
+                    return MockResult([MockRow(rec.copy())])
+            fixtures = {
+                "existing": {"id": 1, "tenant_id": 0, "username": "existing", "email": "existing@test.com", "password_hash": None, "role": "user", "status": "pending", "full_name": None, "bio": None},
+                "alice": {"id": 1, "tenant_id": 1, "username": "alice", "email": "alice@test.com", "password_hash": None, "role": "user", "status": "active", "full_name": "Alice", "bio": "bio"},
```
yihua-zhuo, line 73: **CRITICAL** (critical)

**Issue:** Fixture 'alice' maps username='alice' to tenant_id=0 while other alice-adjacent fixtures (e.g. customers) use tenant_id=1 — this mismatch can cause tests to silently use cross-tenant data when looking up 'alice' by username.

**Suggestion:** Set fixture 'alice' tenant_id to 1, matching the tenant_id used by the rest of the alice-related test fixtures (customers, pipelines, etc.).

tests/unit/domain_handlers/activities.py
```diff
@@ -0,0 +1,124 @@
+"""Activity SQL handlers for unit tests."""
+
+from __future__ import annotations
+
+from tests.unit.conftest import MockResult, MockRow, MockState
+
+ORDER = 70
```
yihua-zhuo, line 7: **INFO** (suggestion)

**Issue:** ORDER = 70 is defined but never used in the module.

**Suggestion:** Remove the unused ORDER constant.

tests/unit/domain_handlers/activities.py
```diff
@@ -0,0 +1,124 @@
...
+        if "select" in sql_text and "count" in sql_text and "from activities" in sql_text:
+            tenant_id = param(params, "tenant_id")
+            customer_id = param(params, "customer_id")
+            activity_type = param(params, "type")
+            count_val = len(
+                [
+                    r
+                    for r in state.activities.values()
+                    if r.get("tenant_id") == tenant_id
+                    and matches_filters(r, customer_id=customer_id, activity_type=activity_type)
```
yihua-zhuo, line 88: **WARNING** (suggestion)

**Issue:** Rule 134: count handler passes tenant_id=None to matches_filters when tenant_id not in params, making count tenant-aware only when a value is present, unlike the SELECT path which always enforces tenant filtering.

**Suggestion:** Consistently handle missing tenant_id: either set a default (like SELECT does with 0) or return 0 count for missing tenant_id rather than returning cross-tenant count.

tests/unit/domain_handlers/counts.py
```diff
@@ -0,0 +1,27 @@
+"""Fallback SQL handlers for unit tests."""
+
+from __future__ import annotations
+
+from tests.unit.conftest import MockResult, MockState
+
+ORDER = 1000
+
+
+def make_count_handler(state: MockState):
```
yihua-zhuo, line 10: **INFO** (suggestion)

**Issue:** The handler runs unconditionally for any COUNT query not caught by domain handlers, meaning it can silently override domain-specific count logic if handler ordering is misconfigured.

**Suggestion:** Add a guard that returns None when a more specific domain count should apply, or add a note in __all__ documenting that this is a last-resort handler.

tests/unit/domain_handlers/customers.py
```diff
@@ -0,0 +1,158 @@
...
+
+def make_customer_handler(state: MockState):
+    """Handle all customer-related SQL (INSERT, UPDATE, DELETE, SELECT, COUNT)."""
+
+    def handler(sql_text, params):
+        if "insert into customers" in sql_text:
+            cid = state.customers_next_id
+            state.customers_next_id += 1
+            tags_raw = params.get("tags", [])
+            tags_val = _json.dumps(tags_raw) if not isinstance(tags_raw, str) else tags_raw
```
yihua-zhuo, line 20: **WARNING** (suggestion)

**Issue:** Rule 136: tags_raw is extracted via params.get('tags') but SQLAlchemy bind parameters for JSON/array columns may arrive as a JSON string or a list depending on how the service layer serializes them — this is not validated.

**Suggestion:** Normalize tags_val: if isinstance(tags_raw, str) attempt _json.loads first, otherwise use str(_json.dumps(tags_raw)) as a fallback to handle both serialized and raw forms.

tests/unit/domain_handlers/tenants.py
```diff
@@ -0,0 +1,56 @@
...
+        return MockResult([[2]])
+
+    if "select" in sql_text and "from tenants" in sql_text and "count" not in sql_text:
+        tenant_id = params.get("tenant_id")
+        if "limit 1" in sql_text and "where id" in sql_text:
+            if tenant_id == 1:
+                return MockResult([MockRow({"id": 1})])
+            if tenant_id == 42:
```
yihua-zhuo, line 35: **WARNING** (suggestion)

**Issue:** tenant_handler returns full tenant fixture for id=42 without any SELECT guard — a query for a non-existent tenant_id returns a valid-looking row, which can hide 'not found' bugs.

**Suggestion:** Only return the id=42 fixture when 'where id = 42' or similar specific filter is present in sql_text, not as a catch-all fallback on any limit-1 select.

tests/unit/domain_handlers/tenants.py
```diff
@@ -0,0 +1,56 @@
...
+                return MockResult([MockRow({"id": 42, "name": "Beta Org", "plan": "enterprise", "status": "active", "settings": '{"sso": true}', "created_at": None, "updated_at": None})])
+            return MockResult([])
+
+        if "where id" not in sql_text:
+            return MockResult([
+                MockRow({"id": 1, "name": "Tenant A", "plan": "pro", "status": "active", "settings": "{}", "created_at": None, "updated_at": None}),
+                MockRow({"id": 2, "name": "Tenant B", "plan": "enterprise", "status": "active", "settings": "{}", "created_at": None, "updated_at": None}),
+            ])
+
+        if tenant_id == 42:
```
yihua-zhuo, line 45: **WARNING** (suggestion)

**Issue:** Duplicate fixture definition for tenant_id=42 appears at both lines 35-36 and 45-46 — if one is updated the other will silently diverge.

**Suggestion:** Extract the tenant 42 fixture to a module-level constant and reuse it.

tests/unit/domain_handlers/tenants.py
```diff
@@ -0,0 +1,56 @@
...
+from tests.unit.conftest import MockResult, MockRow, MockState
+
+ORDER = 30
+
+
+def tenant_handler(sql_text, params):
+    """Handle all tenant-related SQL."""
+    if "insert into tenants" in sql_text:
+        return MockResult([MockRow({"id": 1, "name": params.get("name"), "plan": params.get("plan"), "status": "active", "settings": params.get("settings", "{}"), "created_at": params.get("now"), "updated_at": params.get("now")})])
+
```
yihua-zhuo, line 14: **INFO** (suggestion)

**Issue:** INSERT handler returns id=1 unconditionally — if a test inserts multiple tenants in sequence, all will report id=1, which could mask ordering-dependent test failures.

**Suggestion:** Use a state counter (e.g., state.tenants_next_id) consistent with how users.py and customers.py handle auto-increment IDs.

tests/unit/domain_handlers/users.py
```diff
@@ -0,0 +1,130 @@
...
+            state.deleted_user_ids.add(user_id)
+            if was_in_mock:
+                return MockResult([MockRow({"id": user_id})])
+            return MockResult([])
+
+        if ("set " in sql_text or sql_text.startswith("update")) and "where id" in sql_text:
+            user_id = params.get("id")
+            if user_id in state.users:
+                for k, v in params.items():
+                    if k not in ("id", "tenant_id"):
```
yihua-zhuo, line 54: **INFO** (suggestion)

**Issue:** UPDATE branch explicitly skips 'id' and 'tenant_id' from the param map (correct for id) but silently discards any tenant_id update attempt — if a service ever intentionally changes tenant_id, the mock won't reflect it.

**Suggestion:** Add a comment noting that tenant_id is intentionally immutable in the mock, or assert that tenant_id is never in params to catch unexpected mutations.

Issue #127, opened 2026-05-03
Analytics: embedded charts with PDF export


yihua-zhuo:
## Requirement

Analytics view must support embedded charts (bar, pie, line) with export-to-PDF functionality.

**Implementation notes:**
- Use a charting library: `recharts` or `chart.js` with React wrappers
- Charts: customer acquisition trends (line), revenue by region (bar), deal stage distribution (pie)
- PDF export: `window.print()` or `html2canvas` + `jspdf` for custom layout
- Backend: may need aggregation endpoints for analytics data

**Priority:** Medium

Issue #41, opened 2026-05-02
[Core] Integrate AI Agent Framework into FastAPI App — project structure


yihua-zhuo:
## Problem

`docs/agents/` 下的 AI agent 代码是独立脚本，未集成到 FastAPI 应用。项目需要统一的 AI Agent 框架。

## Goal

将 AI Agent 框架移入 `src/agents/`，作为 FastAPI 依赖：

```python
# src/agents/base.py
class BaseAgent:
    def __init__(self, llm_service, db):
        self.llm = llm_service
        self.db = db
    async def run(self, task: str) -> Dict:
        raise NotImplementedError

# src/services/agent_service.py
class AgentService:
    async def dispatch(self, agent_type: str, task: str) -> Dict:
        ...
```

## Dependencies

**Blocked by:** None — this is the core foundation.

### Agent Interface
- [ ] `src/agents/base.py` — `BaseAgent` abstract class
  - `run(task: str) -> Dict` method
  - `llm` and `db` dependencies injected
  - Type hints for all methods
- [ ] `BaseAgent` registered with `@agent_registry.register("type")`

### Coordinator Agent
- [ ] `src/agents/coordinator.py` — `CoordinatorAgent`
  - Parse natural language task description
  - Decompose into subtasks
  - Assign subtasks to appropriate agents
  - Track overall progress

### Agent Registry
- [ ] `src/agents/registry.py`
  - `AgentRegistry` singleton
  - `register(name)` decorator
  - `get(name) -> BaseAgent`
  - `list_agents() -> List[str]`

### LLM Service Integration
- [ ] `src/services/llm_service.py` — unified LLM interface
  - Supports multiple providers (configurable)
  - `chat(messages, model?) -> str`
  - `embed(text) -> List[float]`
  - Retry logic with backoff
  - Cost tracking per tenant

### FastAPI Integration
- [ ] `src/api/deps.py`
  - `get_llm_service()` dependency
  - `get_agent_service()` dependency
- [ ] Health check endpoint for AI services

### Tests
- [ ] Unit tests for AgentRegistry
- [ ] Unit tests for BaseAgent
- [ ] Integration tests for CoordinatorAgent

PR #764, opened 2026-05-31
Auto: implement issue #595

### LINKED ISSUE #595: Add POST /notifications/smart endpoint
Subtask of #47
Depends on #594

Add src/api/routers/notifications.py router with POST /notifications/smart that accepts an event payload, persists a SmartNotification record, calls NotificationRoutingService, and returns delivery status. Requires AuthContext and tenant isolation. Add unit tests in tests/unit/test_notifications.py. This endpoint does not call an LLM — it accepts pre-classified priority/channel from the caller; the LLM integration is left as a TODO comment pointing to the #41 framework.

Dev-Plan: docs/dev-plan/40-campaigns/0595-add-post-notifications-smart-endpoint.md

yihua-zhuo:
Closes #595

Plan: `.plans/issue-595.md`

1. **Plan** — Claude wrote a detailed implementation plan to `.plans/issue-595.md`.

Please review before merging.

## Summary by CodeRabbit

* **New Features**
  * Added a smart notifications API to create and route pre-classified notifications with priority, channel, timing, optional recipient filtering, and delivery result reporting.

* **Tests**
  * Added extensive unit tests covering happy path, validation failures, tenant isolation, routing outcomes (including empty deliveries), service persistence, and content validation.

* **Documentation**
  * Added an implementation plan outlining behavior, validation rules, acceptance criteria, and testing approach.

### DIFF
--- a/.plans/issue-595.md
+++ b/.plans/issue-595.md
@@ -1,0 +1,221 @@
+I now have a complete picture of the codebase. Here is the implementation plan.
+
+---
+
+# Implementation Plan — Issue #595
+
+## Goal
+
+Add `POST /notifications/smart` to the existing `notifications_router` at `src/api/routers/notifications.py`. The endpoint accepts a pre-classified event payload, persists a `SmartNotificationModel` record, calls `NotificationRoutingService.route()`, and returns the delivery status. It requires `AuthContext`, enforces tenant isolation, and includes a `TODO` comment pointing to issue #41 for future LLM integration. Unit tests live in `tests/unit/test_notifications.py`.
+
+---
+
+## Source Contract
+
+Dev-plan target: `/home/runner/work/agent-job/agent-job/docs/dev-plan/40-campaigns/0595-add-post-notifications-smart-endpoint.md`
+Template depth: `deep`
+Reading order followed:
+1. `/home/runner/work/agent-job/agent-job/docs/dev-plan/README.md`
+2. `/home/runner/work/agent-job/agent-job/docs/dev-plan/_template-deep.md`
+3. `/home/runner/work/agent-job/agent-job/docs/dev-plan/40-campaigns/0595-add-post-notifications-smart-endpoint.md`
+
+---
+
+## Affected Files
+
+- `src/api/routers/notifications.py` — add `SmartNotificationCreate` Pydantic schema and `POST /notifications/smart` endpoint
+- `src/services/notification_service.py` — add `create_smart_notification` method
+- `tests/unit/test_notifications.py` — **create new file** with unit tests for the new endpoint
+
+---
+
+## Implementation Steps
+
+### Step 1: Add `create_smart_notification` to `NotificationService`
+
+Open `src/services/notification_service.py` and add an import for `SmartNotificationModel` alongside the existing model imports. Then add a new async method to `NotificationService`:
+
+```python
+async def create_smart_notification(
+    self,
+    summarized_content: str,
+    priority: int,
+    channel: int,
+    timing: int,
+    tenant_id: int,
+    recipient_filter: dict | None = None,
+) -> SmartNotificationModel:
+    """Persist a smart notification record and return it.
+
+    Actual channel routing is performed by NotificationRoutingService.route()
+    in the router — this method only persists.
+    """
+    notification = SmartNotificationModel(
+        tenant_id=tenant_id,
+        summarized_content=summarized_content,
+        priority=priority,
+        channel=channel,
+        timing=timing,
+        recipient_filter=recipient_filter,
+    )
+    self.session.add(notification)
+    await self.session.flush()
+    return notification
+```
+
+The import to add (after the existing `NotificationModel` import):
+```python
+from db.models.smart_notification import SmartNotificationModel
+```
+
+**Completion check**: `PYTHONPATH=src python -c "from services.notification_service import NotificationService; print('ok')"` exits 0.
+
+### Step 2: Add Pydantic request schema and `POST /notifications/smart` endpoint
+
+Open `src/api/routers/notifications.py`. Add the import for the enums and the routing service at the top of the existing imports block:
+
+```python
+from db.models.smart_notification import Priority, Channel, Timing
+from models.channel_delivery import ChannelDelivery
+from services.notification_routing_service import NotificationRoutingService
+```
+
+Add the `SmartNotificationCreate` Pydantic schema after the existing `ReminderCreate` class (before the `# Notification endpoints` comment):
+
+```python
+class SmartNotificationCreate(BaseModel):
+    summarized_content: str = Field(..., min_length=1, max_length=1024)
+    priority: int = Field(..., ge=0, le=2, description="0=urgent, 1=normal, 2=low")
+    channel: int = Field(..., ge=0, le=3, description="0=email, 1=sms, 2=push, 3=in_app")
+    timing: int = Field(..., ge=0, le=1, description="0=immediate, 1=batch")
+    recipient_filter: dict | None = Field(None, description="Filter criteria for routing")
+
+    @field_validator("priority")
+    @classmethod
+    def priority_must_be_valid(cls, v: int) -> int:
+        if v not in {p.value for p in Priority}:
+            raise ValueError(f"priority must be 0 (urgent), 1 (normal), or 2 (low), got {v}")
+        return v
+
+    @field_validator("channel")
+    @classmethod
+    def channel_must_be_valid(cls, v: int) -> int:
+        if v not in {c.value for c in Channel}:
+            raise ValueError(f"channel must be 0 (email), 1 (sms), 2 (push), or 3 (in_app), got {v}")
+        return v
+```
+
+Add the endpoint after the `send_notification` endpoint (before `PUT /notifications/{id}/read`):
+
+```python
+@notifications_router.post(
+    "/notifications/smart",
+    summary="Create a smart notification with routing",
+)
+async def create_smart_notification(
+    body: SmartNotificationCreate,
+    current_user: AuthContext = Depends(require_auth),
+    session: AsyncSession = Depends(get_db),
+):
+    """Accept a pre-classified event payload, persist a SmartNotification, and route it.
+
+    The LLM classification integration is tracked in issue #41.
+    """
+    if current_user.tenant_id is None or current_user.tenant_id == 0:
+        raise HTTPException(status_code=401, detail="无效的租户信息")
+
+    svc = NotificationService(session)
+    record = await svc.create_smart_notification(
+        summarized_content=body.summarized_content,
+        priority=body.priority,
+        channel=body.channel,
+        timing=body.timing,
+        tenant_id=current_user.tenant_id,
+        recipient_filter=body.recipient_filter,
+    )
+
+    # Route via NotificationRoutingService to determine delivery channels
+    routing_svc = NotificationRoutingService(session)
+    deliveries = await routing_svc.route(record, tenant_id=current_user.tenant_id)
+
+    return {
+        "success": True,
+        "data": {
+            "notification": record.to_dict(),
+            "deliveries": [d.model_dump() for d in deliveries],
+        },
+        "message": "Smart notification created and routed",
+    }
+```
+
+**Completion check**: `ruff check src/api/routers/notifications.py` exits 0.
+
+### Step 3: Add unit tests in `tests/unit/test_notifications.py`
+
+Create the file `tests/unit/test_notifications.py` (the issue specifically requests this path, distinct from `test_notifications_router.py`). Follow the established pattern from `test_notifications_router.py`:
+
+```python
+"""Unit tests for POST /notifications/smart endpoint."""
+
+from __future__ import annotations
+
+from datetime import UTC, datetime
+from unittest.mock import AsyncMock, MagicMock, patch
+
+from fastapi import FastAPI
+from fastapi.testclient import TestClient
+from starlette.responses import JSONResponse
+
+from api.routers.notifications import notifications_router
+from db.connection import get_db
+from internal.middleware.fastapi_auth import AuthContext, require_auth
+from models.channel_delivery import ChannelDelivery
+from pkg.errors.app_exceptions import AppException
+from tests.unit.conftest import make_mock_session
+```
+
+Add a `_MockSmartNotificationModel` class (mirrors the `_MockNotificationModel` pattern, implements `to_dict()` and `__iter__`). Add tests for:
+
+- **Happy path**: valid payload → 200, `notification` key and `deliveries` key present, `NotificationService.create_smart_notification` and `NotificationRoutingService.route` both called with correct args.
+- **Validation — missing `summarized_content`**: 422.
+- **Validation — invalid `priority`**: 422 (value out of `{0,1,2}` range).
+- **Invalid tenant (tenant_id=0)**: 401.
+- **Routing returns empty deliveries** (e.g. priority=normal with no user_id): still 200 with empty `deliveries` list.
+
+**Completion check**: `PYTHONPATH=src pytest tests/unit/test_notifications.py -v` → all passed.
+
+### Step 4: Run full lint and unit suite
+
+```bash
+ruff check src/ && ruff format --check src/
+PYTHONPATH=src pytest tests/unit/ -v
+```
+
+All must exit 0.
+
+---
+
+## Test Plan
+
+- **Unit tests in `tests/unit/`**: Create `tests/unit/test_notifications.py` covering the `POST /notifications/smart` endpoint (schema validation, tenant isolation, service call forwarding, routing integration). Uses `patch` on `NotificationService` and `NotificationRoutingService` at the class level, following the established `test_notifications_router.py` pattern.
+- **Integration tests in `tests/integration/`**: No new integration test file required — the endpoint uses existing ORM models and services already covered by integration test fixtures. An integration test can be added as a follow-up if desired.
+- **Dev-plan verification**: The dev-plan board (§6) has no machine-checkable command beyond the existing `ruff check` and `pytest` suite; the new tests in `test_notifications.py` will be picked up by `PYTHONPATH=src pytest tests/unit/ -v`.
+
+---
+
+## Acceptance Criteria
+
+- `POST /api/v1/notifications/smart` with a valid payload returns `200` with `{"success": true, "data": {"notification": {...}, "deliveries": [...]}, "message": "..."}`
+- Pydantic schema rejects missing `summarized_content`, out-of-range `priority`, and out-of-range `channel` with `422`
+- `tenant_id=0` returns `401`
+- `NotificationRoutingService.route()` is called with the persisted record and `tenant_id`
+- `ruff check src/` exits 0
+- `PYTHONPATH=src pytest tests/unit/test_notifications.py -v` → all passed
+- A `TODO` comment pointing to issue #41 is present in the router or service
+
+---
+
+## Risks / Open Questions
+
+- The `NotificationRoutingService.route()` method reads `getattr(notification, "priority", None)` and expects string values (`"urgent"`, `"normal"`, `"low"`), but `SmartNotificationModel.priority` is an `IntEnum` (values `0`, `1`, `2`). The service will raise `ValidationException` on the integer enum values. **Action needed**: the router or a helper must map the integer `priority` (from the request body) to its string name before passing to `route()`, or `route()` must be updated to handle both. The safest approach is to add a string conversion in the router before calling `route()` (e.g. map `0→"urgent"`, `1→"normal"`, `2→"low"`), keeping the service unchanged and avoiding a breaking change to `route()`.
+- `ChannelDelivery` uses `channel: str` but `SmartNotificationModel.channel` is an `IntEnum`. The `getattr` call in `route()` returns the int — the current service code never reads `channel` via `getattr`, only `priority`, so this is not a runtime blocker, but consistency should be reviewed when the LLM integration in #41 is implemented.
--- a/.plans/issue-595.md
+++ b/.plans/issue-595.md
@@ -148,7 +148,55 @@ async def create_smart_notification(
     }
 ```
 
-**Completion check**: `ruff check src/api/routers/notifications.py` exits 0.
+#### Priority IntEnum → string adapter (part of Step 2)
+
+`NotificationRoutingService.route()` reads `getattr(notification, "priority", None)` and expects string values (`"urgent"`, `"normal"`, `"low"`), but `SmartNotificationModel.priority` is a `Priority` IntEnum (values `0`, `1`, `2`). Directly passing the ORM object to `route()` would cause a `ValidationException`.
+
+Add the following constants and helper class before the `# Notification endpoints` comment (after `SmartNotificationCreate`):
+
+```python
+_PRIORITY_MAP = {Priority.urgent: "urgent", Priority.normal: "normal", Priority.low: "low"}
+
+
+def _priority_to_string(priority) -> str:
+    """Convert a priority value (Priority enum, int, or string) to routing string."""
+    if isinstance(priority, Priority):
+        return _PRIORITY_MAP[priority]
+    if isinstance(priority, int):
+        try:
+            return _PRIORITY_MAP[Priority(priority)]
+        except KeyError:
+            return str(priority)
+    return str(priority)
+
+
+class _MockRoutingRecord:
+    """Lightweight adapter that exposes priority as a string for NotificationRoutingService.
+
+    Rather than mutate the ORM object (which would corrupt in-memory state before
+    serialization), we project it into a plain adapter with priority converted to a string.
+    """
+
+    def __init__(self, record):
+        self.id = record.id
+        self.tenant_id = record.tenant_id
+        self.priority = _priority_to_string(record.priority)
+        self.channel = record.channel
+        self.timing = record.timing
+        self.summarized_content = record.summarized_content
+        self.recipient_filter = record.recipient_filter
+```
+
+In the endpoint, replace `deliveries = await routing_svc.route(record, ...)` with:
+
+```python
+    routing_record = _MockRoutingRecord(record)
+    deliveries = await routing_svc.route(routing_record, tenant_id=current_user.tenant_id)
+```
+
+The `isinstance` guard in `_priority_to_string` handles all three input shapes (enum, int, string), and `_MockRoutingRecord` projects the ORM object into the shape `route()` expects.
+
+**Completion check**: `ruff check src/` exits 0.
 
 ### Step 3: Add unit tests in `tests/unit/test_notifications.py`
 
--- a/src/api/routers/notifications.py
+++ b/src/api/routers/notifications.py
@@ -111,6 +111,25 @@ class SmartNotificationCreate(BaseModel):
 
 _PRIORITY_MAP = {Priority.urgent: "urgent", Priority.normal: "normal", Priority.low: "low"}
 
+
+class _MockRoutingRecord:
+    """Lightweight adapter that exposes priority as a string for NotificationRoutingService.
+
+    NotificationRoutingService.route() expects priority to be a string
+    ('urgent' | 'normal' | 'low'). SmartNotificationModel stores it as a Priority IntEnum.
+    Rather than mutate the ORM object (which would corrupt in-memory state before
+    serialization), we project it into a plain adapter.
+    """
+
+    def __init__(self, record):
+        self.id = record.id
+        self.tenant_id = record.tenant_id
+        self.priority = _PRIORITY_MAP[record.priority]
+        self.channel = record.channel
+        self.timing = record.timing
+        self.summarized_content = record.summarized_content
+        self.recipient_filter = record.recipient_filter
+
 # ---------------------------------------------------------------------------
 # Notification endpoints
 # ---------------------------------------------------------------------------
@@ -196,10 +215,10 @@ async def create_smart_notification(
 
     # Route via NotificationRoutingService to determine delivery channels
     routing_svc = NotificationRoutingService(session)
-    # Map IntEnum priority to string name before routing
-    routing_priority = _PRIORITY_MAP[record.priority] if isinstance(record.priority, Priority) else record.priority
-    record.priority = routing_priority
-    deliveries = await routing_svc.route(record, tenant_id=current_user.tenant_id)
+    routing_record = record
+    if isinstance(record.priority, Priority):
+        routing_record = _MockRoutingRecord(record)
+    deliveries = await routing_svc.route(routing_record, tenant_id=current_user.tenant_id)
 
     return {
         "success": True,
--- a/src/api/routers/notifications.py
+++ b/src/api/routers/notifications.py
@@ -112,6 +112,19 @@ class SmartNotificationCreate(BaseModel):
 _PRIORITY_MAP = {Priority.urgent: "urgent", Priority.normal: "normal", Priority.low: "low"}
 
 
+def _priority_to_string(priority) -> str:
+    """Convert a priority value (Priority enum, int, or string) to routing string."""
+    if isinstance(priority, Priority):
+        return _PRIORITY_MAP[priority]
+    if isinstance(priority, int):
+        # Plain int (e.g. 0, 1, 2) — convert via Priority IntEnum then map
+        try:
+            return _PRIORITY_MAP[Priority(priority)]
+        except KeyError:
+            return str(priority)
+    return str(priority)
+
+
 class _MockRoutingRecord:
     """Lightweight adapter that exposes priority as a string for NotificationRoutingService.
 
@@ -124,7 +137,7 @@ class _MockRoutingRecord:
     def __init__(self, record):
         self.id = record.id
         self.tenant_id = record.tenant_id
-        self.priority = _PRIORITY_MAP[record.priority]
+        self.priority = _priority_to_string(record.priority)
         self.channel = record.channel
         self.timing = record.timing
         self.summarized_content = record.summarized_content
@@ -215,9 +228,7 @@ async def create_smart_notification(
 
     # Route via NotificationRoutingService to determine delivery channels
     routing_svc = NotificationRoutingService(session)
-    routing_record = record
-    if isinstance(record.priority, Priority):
-        routing_record = _MockRoutingRecord(record)
+    routing_record = _MockRoutingRecord(record)
     deliveries = await routing_svc.route(routing_record, tenant_id=current_user.tenant_id)
 
     return {
--- a/src/api/routers/notifications.py
+++ b/src/api/routers/notifications.py
@@ -9,7 +9,7 @@ from pydantic import BaseModel, Field, field_validator
 from sqlalchemy.ext.asyncio import AsyncSession
 
 from db.connection import get_db
-from db.models.smart_notification import Channel, Priority
+from db.models.smart_notification import Channel, Priority, Timing
 from internal.middleware.fastapi_auth import AuthContext, require_auth
 from pkg.constants.notification_constants import VALID_NOTIFICATION_CHANNELS
 from services.notification_routing_service import NotificationRoutingService
@@ -101,6 +101,15 @@ class SmartNotificationCreate(BaseModel):
             raise ValueError(f"channel must be 0 (email), 1 (sms), 2 (push), or 3 (in_app), got {v}")
         return v
 
+    @field_validator("timing")
+    @classmethod
+    def timing_must_be_valid(cls, v: int) -> int:
+        if v not in {t.value for t in Timing}:
+            raise ValueError(f"timing must be 0 (immediate) or 1 (batch), got {v}")
+        return v
+
+
+_PRIORITY_MAP = {Priority.urgent: "urgent", Priority.normal: "normal", Priority.low: "low"}
 
 # ---------------------------------------------------------------------------
 # Notification endpoints
@@ -187,6 +196,9 @@ async def create_smart_notification(
 
     # Route via NotificationRoutingService to determine delivery channels
     routing_svc = NotificationRoutingService(session)
+    # Map IntEnum priority to string name before routing
+    routing_priority = _PRIORITY_MAP[record.priority] if isinstance(record.priority, Priority) else record.priority
+    record.priority = routing_priority
     deliveries = await routing_svc.route(record, tenant_id=current_user.tenant_id)
 
     return {
--- a/src/api/routers/notifications.py
+++ b/src/api/routers/notifications.py
@@ -9,8 +9,10 @@ from pydantic import BaseModel, Field, field_validator
 from sqlalchemy.ext.asyncio import AsyncSession
 
 from db.connection import get_db
+from db.models.smart_notification import Channel, Priority
 from internal.middleware.fastapi_auth import AuthContext, require_auth
 from pkg.constants.notification_constants import VALID_NOTIFICATION_CHANNELS
+from services.notification_routing_service import NotificationRoutingService
 from services.notification_service import NotificationService
 
 notifications_router = APIRouter(prefix="/api/v1", tags=["notifications"])
@@ -78,6 +80,28 @@ class ReminderCreate(BaseModel):
     related_id: int | None = Field(None, ge=1)
 
 
+class SmartNotificationCreate(BaseModel):
+    summarized_content: str = Field(..., min_length=1, max_length=1024)
+    priority: int = Field(..., ge=0, le=2, description="0=urgent, 1=normal, 2=low")
+    channel: int = Field(..., ge=0, le=3, description="0=email, 1=sms, 2=push, 3=in_app")
+    timing: int = Field(..., ge=0, le=1, description="0=immediate, 1=batch")
+    recipient_filter: dict | None = Field(None, description="Filter criteria for routing")
+
+    @field_validator("priority")
+    @classmethod
+    def priority_must_be_valid(cls, v: int) -> int:
+        if v not in {p.value for p in Priority}:
+            raise ValueError(f"priority must be 0 (urgent), 1 (normal), or 2 (low), got {v}")
+        return v
+
+    @field_validator("channel")
+    @classmethod
+    def channel_must_be_valid(cls, v: int) -> int:
+        if v not in {c.value for c in Channel}:
+            raise ValueError(f"channel must be 0 (email), 1 (sms), 2 (push), or 3 (in_app), got {v}")
+        return v
+
+
 # ---------------------------------------------------------------------------
 # Notification endpoints
 # ---------------------------------------------------------------------------
@@ -133,6 +157,46 @@ async def send_notification(
         related_id=body.related_id,
     )
     return {"success": True, "data": data.to_dict(), "message": "通知发送成功"}
+
+
+@notifications_router.post(
+    "/notifications/smart",
+    summary="Create a smart notification with routing",
+)
+async def create_smart_notification(
+    body: SmartNotificationCreate,
+    current_user: AuthContext = Depends(require_auth),
+    session: AsyncSession = Depends(get_db),
+):
+    """Accept a pre-classified event payload, persist a SmartNotification, and route it.
+
+    The LLM classification integration is tracked in issue #41.
+    """
+    if current_user.tenant_id is None or current_user.tenant_id == 0:
+        raise HTTPException(status_code=401, detail="无效的租户信息")
+
+    svc = NotificationService(session)
+    record = await svc.create_smart_notification(
+        summarized_content=body.summarized_content,
+        priority=body.priority,
+        channel=body.channel,
+        timing=body.timing,
+        tenant_id=current_user.tenant_id,
+        recipient_filter=body.recipient_filter,
+    )
+
+    # Route via NotificationRoutingService to determine delivery channels
+    routing_svc = NotificationRoutingService(session)
+    deliveries = await routing_svc.route(record, tenant_id=current_user.tenant_id)
+
+    return {
+        "success": True,
+        "data": {
+            "notification": record.to_dict(),
+            "deliveries": [d.model_dump() for d in deliveries],
+        },
+        "message": "Smart notification created and routed",
+    }
 
 
 @notifications_router.put(
--- a/src/services/notification_service.py
+++ b/src/services/notification_service.py
@@ -10,6 +10,7 @@ from sqlalchemy.ext.asyncio import AsyncSession
 
 from db.models.notification import NotificationModel
 from db.models.reminder import ReminderModel
+from db.models.smart_notification import SmartNotificationModel
 from db.models.user import UserModel
 from pkg.constants.notification_constants import (
     PAYLOAD_PARAMS_ALLOWED_KEYS,
@@ -167,6 +168,32 @@ class NotificationService:
         if notification is None:
             raise NotFoundException("Notification")
         await self.session.delete(notification)
+        await self.session.flush()
+        return notification
+
+    async def create_smart_notification(
+        self,
+        summarized_content: str,
+        priority: int,
+        channel: int,
+        timing: int,
+        tenant_id: int,
+        recipient_filter: dict | None = None,
+    ) -> SmartNotificationModel:
+        """Persist a smart notification record and return it.
+
+        Actual channel routing is performed by NotificationRoutingService.route()
+        in the router — this method only persists.
+        """
+        notification = SmartNotificationModel(
+            tenant_id=tenant_id,
+            summarized_content=summarized_content,
+            priority=priority,
+            channel=channel,
+            timing=timing,
+            recipient_filter=recipient_filter,
+        )
+        self.session.add(notification)
         await self.session.flush()
         return notification
 
--- a/src/services/notification_service.py
+++ b/src/services/notification_service.py
@@ -10,7 +10,7 @@ from sqlalchemy.ext.asyncio import AsyncSession
 
 from db.models.notification import NotificationModel
 from db.models.reminder import ReminderModel
-from db.models.smart_notification import SmartNotificationModel
+from db.models.smart_notification import Channel, Priority, SmartNotificationModel, Timing
 from db.models.user import UserModel
 from pkg.constants.notification_constants import (
     PAYLOAD_PARAMS_ALLOWED_KEYS,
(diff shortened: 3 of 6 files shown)


.plans/issue-595.md
```diff
@@ -0,0 +1,221 @@
...
+- `NotificationRoutingService.route()` is called with the persisted record and `tenant_id`
+- `ruff check src/` exits 0
+- `PYTHONPATH=src pytest tests/unit/test_notifications.py -v` → all passed
+- A `TODO` comment pointing to issue #41 is present in the router or service
+
+---
+
+## Risks / Open Questions
+
+- The `NotificationRoutingService.route()` method reads `getattr(notification, "priority", None)` and expects string values (`"urgent"`, `"normal"`, `"low"`), but `SmartNotificationModel.priority` is an `IntEnum` (values `0`, `1`, `2`). The service will raise `ValidationException` on the integer enum values. **Action needed**: the router or a helper must map the integer `priority` (from the request body) to its string name before passing to `route()`, or `route()` must be updated to handle both. The safest approach is to add a string conversion in the router before calling `route()` (e.g. map `0→"urgent"`, `1→"normal"`, `2→"low"`), keeping the service unchanged and avoiding a breaking change to `route()`.
```
yihua-zhuo, line 268: **CRITICAL** (critical)

**Issue:** The plan's own risk section (line 220) identifies that NotificationRoutingService.route() compares priority against string literals ('urgent', 'normal', 'low') but SmartNotificationModel.priority is an IntEnum (0, 1, 2). The plan states 'Action needed: the router must map integer priority to its string name before calling route(), keeping the service unchanged.' This action was never taken — the implementation silently omits the conversion, so every POST /notifications/smart call will raise ValidationException('Unknown notification priority: 1') in the routing layer.

**Suggestion:** Before calling routing_svc.route(), map the integer priority to its string name: _PRIORITY_MAP = {0: 'urgent', 1: 'normal', 2: 'low'}; then either set record.priority = _PRIORITY_MAP[body.priority] (if SmartNotificationModel.priority is writable) or construct a lightweight wrapper/dict with the mapped string priority and pass that to route(). Alternatively, update route() to accept IntEnum values.

src/api/routers/notifications.py
```diff
@@ -135,6 +159,46 @@ async def send_notification(
...
+        priority=body.priority,
+        channel=body.channel,
+        timing=body.timing,
+        tenant_id=current_user.tenant_id,
+        recipient_filter=body.recipient_filter,
+    )
+
+    # Route via NotificationRoutingService to determine delivery channels
+    routing_svc = NotificationRoutingService(session)
+    deliveries = await routing_svc.route(record, tenant_id=current_user.tenant_id)
```
yihua-zhuo: **CRITICAL** (critical)

**Issue:** NotificationRoutingService.route() compares priority against string literals ('urgent', 'normal', 'low') at lines 29, 45, 55 of routing service, but SmartNotificationModel.priority is an IntEnum (0, 1, 2). The router passes the record directly without converting priority to its string name, so every routed notification will raise ValidationException('Unknown notification priority: 1') — this is the unaddressed risk from the plan's own risk section.

**Suggestion:** Map the integer priority to a string before calling route(). One approach: _PRIORITY_MAP = {0: 'urgent', 1: 'normal', 2: 'low'}; then patch the record's priority attribute to the string or pass a routing-specific dict/object with the mapped string priority to routing_svc.route().

tests/unit/test_notifications.py
```diff
@@ -0,0 +1,292 @@
...
+
+
+async def _mock_get_db():
+    yield make_mock_session([])
+
+
+def _make_auth_override(tenant_id: int = 1, user_id: int = 99):
+    """Mimics real JWT-based AuthContext creation: raises 401 for invalid tenants."""
+    if not isinstance(tenant_id, int) or tenant_id <= 0:
+        raise HTTPException(status_code=401, detail="Token is missing a valid tenant_id")
```
yihua-zhuo, line 87: **CRITICAL** (critical)

**Issue:** _make_auth_override references HTTPException (for the tenant validation branch) but HTTPException is never imported. This causes NameError at test collection or runtime when _make_auth_override() is called, making TestCreateSmartNotificationTenantIsolation entirely broken.

**Suggestion:** Add 'from starlette.exceptions import HTTPException' to the imports near line 6.

tests/unit/test_notifications.py
```diff
@@ -0,0 +1,292 @@
...
+    return TestClient(app, raise_server_exceptions=False)
+
+
+# ---------------------------------------------------------------------------
+# POST /notifications/smart — happy path
+# ---------------------------------------------------------------------------
+
+
+class TestCreateSmartNotification:
+    def test_create_smart_notification_ok(self):
```
yihua-zhuo, line 116: **CRITICAL** (critical)

**Issue:** The happy-path test (test_create_smart_notification_ok) patches NotificationRoutingService at the class level and mocks its .route() method to return a fixed list, so it never exercises the actual routing logic. Combined with the router's missing priority-to-string conversion, this test gives false confidence — it passes but the real endpoint always raises ValidationException in production.

**Suggestion:** Add an integration-style test that patches at a lower level or verifies the actual priority comparison path. Alternatively, add a unit test for the routing service directly that uses a real SmartNotificationModel (with priority converted to string) to catch the type mismatch.

.plans/issue-595.md
```diff
@@ -0,0 +1,221 @@
...
+
+**Completion check**: `PYTHONPATH=src python -c "from services.notification_service import NotificationService; print('ok')"` exits 0.
+
+### Step 2: Add Pydantic request schema and `POST /notifications/smart` endpoint
+
+Open `src/api/routers/notifications.py`. Add the import for the enums and the routing service at the top of the existing imports block:
+
+```python
+from db.models.smart_notification import Priority, Channel, Timing
+from models.channel_delivery import ChannelDelivery
```
yihua-zhuo, line 79: **WARNING** (suggestion)

**Issue:** The plan specifies importing both Channel and Timing enums ('from db.models.smart_notification import Priority, Channel, Timing') but the router imports Priority and Channel only. Timing is never imported yet used in SmartNotificationCreate.timing field — this is inconsistent with the plan and makes the schema less maintainable.

**Suggestion:** Add Timing to the import: 'from db.models.smart_notification import Channel, Priority, Timing'

.plans/issue-595.md
```diff
@@ -0,0 +1,221 @@
...
+from services.notification_routing_service import NotificationRoutingService
+```
+
+Add the `SmartNotificationCreate` Pydantic schema after the existing `ReminderCreate` class (before the `# Notification endpoints` comment):
+
+```python
+class SmartNotificationCreate(BaseModel):
+    summarized_content: str = Field(..., min_length=1, max_length=1024)
+    priority: int = Field(..., ge=0, le=2, description="0=urgent, 1=normal, 2=low")
```
yihua-zhuo, line 88: **INFO** (suggestion)

**Issue:** The plan's SmartNotificationCreate schema includes a field_validator on timing (step 2, not shown in the snippet but implied by symmetry), yet the implementation only validates priority and channel. If timing values outside {0,1} are passed, the Pydantic ge/le guards catch them, so there is no silent failure — but an explicit validator matching the plan's style is missing.

**Suggestion:** Add a timing field_validator for consistency: '@field_validator("timing") @classmethod def timing_must_be_valid(cls, v): if v not in {t.value for t in Timing}: raise ValueError(...); return v'

tests/unit/test_notifications.py
```diff
@@ -0,0 +1,292 @@
...
+        )
+        assert response.status_code == 422
+
+
+# ---------------------------------------------------------------------------
+# POST /notifications/smart — routing integration
+# ---------------------------------------------------------------------------
+
+
+class TestCreateSmartNotificationRouting:
```
yihua-zhuo, line 306: **WARNING** (suggestion)

**Issue:** test_routing_returns_empty_deliveries has a comment 'priority=normal with no user_id returns200 with empty deliveries list.' The return value depends on the string 'normal' comparison in route(), but since route() is fully mocked this test only verifies the endpoint's response envelope shape, not the routing condition. The test name is also misspaced ('returns200' → 'returns 200').

**Suggestion:** Rename the test to fix the typo and clarify it tests the endpoint's empty-delivery handling, not the routing logic: 'test_empty_deliveries_still_returns_200'.

tests/unit/test_notifications.py
```diff
@@ -0,0 +1,292 @@
...
+
+from __future__ import annotations
+
+from datetime import UTC, datetime
+from unittest.mock import AsyncMock, MagicMock, patch
+
+from fastapi import FastAPI
+from fastapi.testclient import TestClient
+from starlette.responses import JSONResponse
+
```
yihua-zhuo, line 13: **INFO** (suggestion)

**Issue:** The test file imports make_mock_session from tests.unit.conftest but the mock DB is never used — all tests patch at the class level. This is not wrong (tests/unit/conftest.py is shared infrastructure, not modified), but the import is dead weight.

**Suggestion:** Remove the unused 'from tests.unit.conftest import make_mock_session' import to keep the file focused.

yihua-zhuo:

**Status:** `fail`
**Inline comments posted:** 8

**Summary:** Reviewed 6 file(s) between origin/master and origin/auto/issue-595.

- `src/services/notification_service.py`: Service correctly persists smart notifications with flush-only semantics per project rule 122; validation is duplicated at both schema and service layer which is defensive but consistent.

`tests/unit/domain_handlers/notification.py` L225: 
**CRITICAL** (critical)

**Issue:** cancel_reminder handler filters only by tenant_id, but the service also filters by user_id (ReminderModel.id == reminder_id AND ReminderModel.tenant_id == tenant_id). Missing user_id check allows cross-user cancellation.

**Suggestion:** Add user_id check: after fetching by id, verify r.get('user_id') == params.get('user_id') before allowing the delete. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`tests/unit/domain_handlers/notification.py` L246: 
**CRITICAL** (critical)

**Issue:** List-reminders handler does not handle order_by(ReminderModel.remind_at) SQL. The service generates .order_by(ReminderModel.remind_at) on the select, but the handler has no branch for it, so it falls through to return None and the mock session returns an empty result.

**Suggestion:** Add a branch checking for 'order by' in sql_text_lower (after the 'from reminders where id' check) that sorts rows by remind_at before applying offset/page_size slicing. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`tests/unit/domain_handlers/notification.py` L182: 
**WARNING** (suggestion)

**Issue:** make_reminder_handler has no ORDER constant — defaults to 100 in the loader, which could cause it to load before handlers that should run first.

**Suggestion:** Add ORDER =<low number> to the module-level namespace so the loader sorts it correctly. Other domain handler modules that need ordering define this. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`tests/unit/domain_handlers/notification.py` L236: 
**INFO** (suggestion)

**Issue:** Reminder count handler does not bind user_id in its ValueError message (line 238), making debug output slightly misleading compared to the notification count handler which correctly names both tenant_id and user_id.

**Suggestion:** Change the ValueError message to mention both tenant_id and user_id. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_

src/api/routers/notifications.py
```diff
@@ -135,6 +168,49 @@ async def send_notification(
...
+        timing=body.timing,
+        tenant_id=current_user.tenant_id,
+        recipient_filter=body.recipient_filter,
+    )
+
+    # Route via NotificationRoutingService to determine delivery channels
+    routing_svc = NotificationRoutingService(session)
+    # Map IntEnum priority to string name before routing
+    routing_priority = _PRIORITY_MAP[record.priority] if isinstance(record.priority, Priority) else record.priority
+    record.priority = routing_priority
```
yihua-zhuo: **CRITICAL** (critical)

**Issue:** Mutates record.priority on the ORM object returned by the service — violates rule 127 (services return domain objects; routers must not mutate them) and risks corrupting in-memory ORM state before serialization.

**Suggestion:** Make a local variable for routing instead: `routing_priority = _PRIORITY_MAP[record.priority] if isinstance(record.priority, Priority) else record.priority; deliveries = await routing_svc.route(record, tenant_id=...)` — do not assign back to record.priority.

tests/unit/test_notifications.py
```diff
@@ -0,0 +1,293 @@
...
+class TestCreateSmartNotification:
+    def test_create_smart_notification_ok(self):
+        """Valid payload returns 200 with notification and deliveries keys."""
+        with (
+            patch("api.routers.notifications.NotificationService") as svc_cls,
+            patch("api.routers.notifications.NotificationRoutingService") as routing_cls,
+        ):
+            svc = svc_cls.return_value
+            mock_record = _MockSmartNotificationModel({"id": 7, "tenant_id": 1})
```
yihua-zhuo: **CRITICAL** (critical)

**Issue:** _MockSmartNotificationModel stores priority as int (e.g. 1) but NotificationRoutingService.route() compares it to strings ('urgent', 'normal', 'low'). When router maps Priority enum to string via _PRIORITY_MAP, it only triggers when the priority is a Priority IntEnum, not a bare int — causing route() to raise ValidationException and return 500 instead of 200.

**Suggestion:** Set priority=0 (Priority.urgent) or priority='urgent' in the mock to match what the routing service expects. Alternatively, mock record.priority as a Priority enum directly.

tests/unit/test_notifications.py
```diff
@@ -0,0 +1,293 @@
...
+            response = client.post(
+                "/api/v1/notifications/smart",
+                json={
+                    "summarized_content": "Deal closed for $50k",
+                    "priority": 1,
+                    "channel": 0,
+                    "timing": 0,
+                },
+            )
+            assert response.status_code == 200
```
yihua-zhuo, line 158: **CRITICAL** (critical)

**Issue:** test_create_smart_notification_ok asserts only body['success'] and presence of keys body['data']['notification'] and body['data']['deliveries'] — it never validates that deliveries contains the correct ChannelDelivery items, so a 500 from route() raising ValidationException is silently passed.

**Suggestion:** Add assertions: assert body['data']['deliveries'][0]['channel'] == 'in_app' and assert body['data']['deliveries'][0]['target'] == '99'. Also assert response status is 200 and body['success'] is True after the route call succeeds.

src/services/notification_service.py
```diff
@@ -170,6 +171,32 @@ async def delete_notification(self, notification_id: int, tenant_id: int) -> Not
...
+        timing: int,
+        tenant_id: int,
+        recipient_filter: dict | None = None,
+    ) -> SmartNotificationModel:
+        """Persist a smart notification record and return it.
+
+        Actual channel routing is performed by NotificationRoutingService.route()
+        in the router — this method only persists.
+        """
+        notification = SmartNotificationModel(
```
yihua-zhuo, line 195: **WARNING** (suggestion)

**Issue:** create_smart_notification persists priority/channel/timing integers with no validation — any direct call to this method (bypassing the router's Pydantic schema) can write out-of-range values to the DB.

**Suggestion:** Add guard clauses: `if priority not in {p.value for p in Priority}: raise ValidationException(...); same for channel and timing.`

tests/unit/test_notifications.py
```diff
@@ -0,0 +1,293 @@
...
+                json={
+                    "summarized_content": "Update",
+                    "priority": 1,
+                    "channel": 0,
+                    "timing": 0,
+                },
+            )
+            assert response.status_code == 200
+            body = response.json()
+            assert body["data"]["deliveries"] == []
```
yihua-zhuo, line 338: **WARNING** (suggestion)

**Issue:** test_empty_deliveries_still_returns_200 checks body['data']['deliveries'] == [] but never validates the notification field or delivery count, and does not assert that the routing service was called with the correct record — making the test pass even if route() threw a 500.

**Suggestion:** Add assertions for the notification field (id, priority) and assert routing_svc.route.assert_called_once() to ensure the routing path was actually exercised.

tests/unit/test_notifications.py
```diff
@@ -0,0 +1,293 @@
...
+        response = client.post(
+            "/api/v1/notifications/smart",
+            json={
+                "summarized_content": "Test",
+                "priority": 99,
+                "channel": 0,
+                "timing": 0,
+            },
+        )
+        assert response.status_code == 422
```
yihua-zhuo, line 244: **INFO** (suggestion)

**Issue:** test_invalid_priority only asserts response.status_code == 422 — it does not assert which field caused the error or the specific error message, making it possible for wrong-field validation (e.g. channel constraint catching priority's invalid value) to silently pass.

**Suggestion:** Add error field and message assertions: error_fields = {e.get('loc')[-1] for e in errors}; assert 'priority' in error_fields.

tests/unit/test_notifications.py
```diff
@@ -0,0 +1,293 @@
...
+        response = client.post(
+            "/api/v1/notifications/smart",
+            json={
+                "summarized_content": "Test",
+                "priority": 1,
+                "channel": 99,
+                "timing": 0,
+            },
+        )
+        assert response.status_code == 422
```
yihua-zhuo, line 261: **INFO** (suggestion)

**Issue:** test_invalid_channel only asserts response.status_code == 422 without confirming the 'channel' field is the cause, same weakness as test_invalid_priority.

**Suggestion:** Add error field assertion: error_fields = {e.get('loc')[-1] for e in errors}; assert 'channel' in error_fields.

tests/unit/test_notifications.py
```diff
@@ -0,0 +1,293 @@
...
+# POST /notifications/smart — tenant isolation
+# ---------------------------------------------------------------------------
+
+
+class TestCreateSmartNotificationTenantIsolation:
+    def _app_invalid_tenant(self) -> TestClient:
+        app = _make_test_app(lambda: _make_auth_ctx(tenant_id=0, user_id=99))
+        return TestClient(app, raise_server_exceptions=False)
+
+    def test_invalid_tenant_returns_401(self):
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** test_invalid_tenant_returns_401 calls _make_auth_override which raises 401 inside the auth override before require_auth is even called — the test exercises the override's validation, not the router's own if current_user.tenant_id == 0 guard. These are two separate enforcement points and only the override is being tested.

**Suggestion:** Rename the test to reflect it tests the override guard, or add a second test that passes a valid override (tenant_id=1) but with a mock that bypasses the override so the router's own tenant check is exercised.

tests/unit/test_notifications.py
```diff
@@ -0,0 +1,293 @@
...
+        response = client.post(
+            "/api/v1/notifications/smart",
+            json={
+                "summarized_content": "",
+                "priority": 1,
+                "channel": 0,
+                "timing": 0,
+            },
+        )
+        assert response.status_code == 422
```
yihua-zhuo, line 295: **INFO** (suggestion)

**Issue:** test_empty_summarized_content asserts status_code == 422 but does not verify the error targets the 'summarized_content' field specifically, which matters if the schema changes and a different field becomes the first validation failure.

**Suggestion:** Add assertion: error_fields = {e.get('loc')[-1] for e in errors}; assert 'summarized_content' in error_fields.

src/api/routers/notifications.py
```diff
@@ -135,6 +187,49 @@ async def send_notification(
...
+    "/notifications/smart",
+    summary="Create a smart notification with routing",
+)
+async def create_smart_notification(
+    body: SmartNotificationCreate,
+    current_user: AuthContext = Depends(require_auth),
+    session: AsyncSession = Depends(get_db),
+):
+    """Accept a pre-classified event payload, persist a SmartNotification, and route it.
+
```
yihua-zhuo, line 213: **CRITICAL** (critical)

**Issue:** Router mutates the persisted ORM object's priority attribute to a string value after routing, corrupting in-memory ORM state and causing to_dict() to return the wrong type for the response.

**Suggestion:** Do not mutate record.priority. Create a routing helper dict or copy with routing_priority assigned: routing_record = type(record)(..., priority=routing_priority, ...) and pass that to route(). Alternatively pass routing_priority as a separate parameter to route() if the service signature allows.

tests/unit/test_notifications.py
```diff
@@ -0,0 +1,317 @@
...
+    def test_create_smart_notification_ok(self):
+        """Valid payload returns 200 with notification and deliveries keys."""
+        with (
+            patch("api.routers.notifications.NotificationService") as svc_cls,
+            patch("api.routers.notifications.NotificationRoutingService") as routing_cls,
+        ):
+            svc = svc_cls.return_value
+            mock_record = _MockSmartNotificationModel({"id": 7, "tenant_id": 1})
+            svc.create_smart_notification = AsyncMock(return_value=mock_record)
+
```
yihua-zhuo, line 125: **CRITICAL** (critical)

**Issue:** Test creates ChannelDelivery with priority='normal' (str) but NotificationRoutingService.route() returns ChannelDelivery instances where priority is validated as str — this passes but validates nothing about what the real service actually returns.

**Suggestion:** Call the real NotificationRoutingService.route() in the test mock instead of constructing ChannelDelivery manually, or assert the returned deliveries match the routing rules for the given priority.

tests/unit/test_notifications.py
```diff
@@ -0,0 +1,317 @@
...
+                json={
+                    "summarized_content": "Deal closed for $50k",
+                    "priority": 1,
+                    "channel": 0,
+                    "timing": 0,
+                },
+            )
+            assert response.status_code == 200
+            body = response.json()
+            assert body["success"] is True
```
yihua-zhuo, line 160: **CRITICAL** (critical)

**Issue:** Test asserts deliveries length without checking the actual routing outcome — routing_svc.route is mocked to return an arbitrary ChannelDelivery list, so the assertion exercises nothing about the real routing logic (rule 144: weak for import/routing flows).

**Suggestion:** Remove the manual ChannelDelivery mock; let the real NotificationRoutingService.route() execute and assert on its actual output, or verify specific delivery values match what the real service would produce for the given priority.

.plans/issue-595.md
```diff
@@ -0,0 +1,221 @@
...
+
+## Goal
+
+Add `POST /notifications/smart` to the existing `notifications_router` at `src/api/routers/notifications.py`. The endpoint accepts a pre-classified event payload, persists a `SmartNotificationModel` record, calls `NotificationRoutingService.route()`, and returns the delivery status. It requires `AuthContext`, enforces tenant isolation, and includes a `TODO` comment pointing to issue #41 for future LLM integration. Unit tests live in `tests/unit/test_notifications.py`.
+
+---
+
+## Source Contract
+
+Dev-plan target: `/home/runner/work/agent-job/agent-job/docs/dev-plan/40-campaigns/0595-add-post-notifications-smart-endpoint.md`
```
yihua-zhuo, line 15: **WARNING** (suggestion)

**Issue:** Plan references a source dev-plan document at docs/dev-plan/40-campaigns/0595-add-post-notifications-smart-endpoint.md that was not verified to exist in the repo.

**Suggestion:** Confirm the source document exists before citing it as a reading-order prerequisite.

.plans/issue-595.md
```diff
@@ -0,0 +1,221 @@
...
+        },
+        "message": "Smart notification created and routed",
+    }
+```
+
+**Completion check**: `ruff check src/api/routers/notifications.py` exits 0.
+
+### Step 3: Add unit tests in `tests/unit/test_notifications.py`
+
+Create the file `tests/unit/test_notifications.py` (the issue specifically requests this path, distinct from `test_notifications_router.py`). Follow the established pattern from `test_notifications_router.py`:
```
yihua-zhuo, line 203: **INFO** (suggestion)

**Issue:** Plan says to follow the pattern from test_notifications_router.py but does not show or describe that file, so a new contributor cannot reproduce the test pattern from the plan alone.

**Suggestion:** Either inline a condensed version of the patching pattern or reference it with a concrete excerpt.

src/api/routers/notifications.py
```diff
@@ -135,6 +187,49 @@ async def send_notification(
...
+    The LLM classification integration is tracked in issue #41.
```
yihua-zhuo, line 214: **WARNING** (suggestion)

**Issue:** Rule 121 violation: the router-owned ORM object is mutated after routing, which breaks the service/router transaction boundary contract.

**Suggestion:** Remove the record.priority reassignment. Use a separate object for routing input.

src/api/routers/notifications.py
yihua-zhuo, line 213: **INFO** (suggestion)

**Issue:** _PRIORITY_MAP uses Priority enum keys; if record.priority is ever a plain int (not a Priority instance), isinstance check returns False and the int is passed directly to route(), which may cause a KeyError or ValidationException inside NotificationRoutingService.

**Suggestion:** Handle the plain-int fallback explicitly: routing_priority = _PRIORITY_MAP.get(record.priority) or (str(record.priority) if isinstance(record.priority, int) else record.priority).

tests/unit/test_notifications.py
```diff
@@ -0,0 +1,317 @@
...
+            svc = svc_cls.return_value
+            mock_record = _MockSmartNotificationModel({"id": 7, "tenant_id": 1})
+            svc.create_smart_notification = AsyncMock(return_value=mock_record)
+
+            routing_svc = routing_cls.return_value
+            routing_svc.route = AsyncMock(
+                return_value=[
+                    ChannelDelivery(
+                        channel="in_app",
+                        target="99",
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** _MockSmartNotificationModel has no __slots__ but the real AuthContext it cooperates with does — jsonable_encoder falls back to __iter__ for AuthContext but the interaction between __slots__ and __iter__ yielding tuples is fragile and not explicitly tested.

**Suggestion:** Add a test that directly exercises a response containing AuthContext (if any code path returns it) or explicitly document that AuthContext is never serialized directly in responses.

tests/unit/test_notifications.py
```diff
@@ -0,0 +1,317 @@
...
+        assert response.status_code == 422
+        errors = response.json().get("detail", [])
+        error_fields = {e.get("loc")[-1] for e in errors}
+        assert "channel" in error_fields
+
+    def test_empty_summarized_content(self):
+        """Empty summarized_content string returns 422."""
+        client = _app()
+        response = client.post(
+            "/api/v1/notifications/smart",
```
yihua-zhuo, line 287: **WARNING** (suggestion)

**Issue:** test_empty_deliveries_still_returns_200 mocks NotificationRoutingService.route to return [] for priority=normal, but this is impossible in the real service — normal priority with no user_id returns [] at line 48 of notification_routing_service.py, so the mock matches the real behavior, yet the assertion only checks an empty list without verifying the priority/user_id state that drives it (rule 144).

**Suggestion:** Assert that the notification's priority and user_id values would produce an empty delivery list in the real routing service, or remove the mock and test the real routing path.

tests/unit/test_notifications.py
```diff
@@ -0,0 +1,317 @@
...
+        ):
+            svc = svc_cls.return_value
+            mock_record = _MockSmartNotificationModel({"priority": Priority.normal})
+            svc.create_smart_notification = AsyncMock(return_value=mock_record)
+
+            routing_svc = routing_cls.return_value
+            routing_svc.route = AsyncMock(return_value=[])
+
+            client = _app()
+            response = client.post(
```
yihua-zhuo, line 324: **WARNING** (suggestion)

**Issue:** TestCreateSmartNotificationTenantIsolation only tests the invalid-tenant (401) path for create_smart_notification; there is no cross-tenant test asserting that tenant A cannot create notifications visible to tenant B (rule 126).

**Suggestion:** Add a test that creates a notification as tenant_id=1 and verifies that routing_svc.route is called with tenant_id=1, not another tenant's ID; optionally add a test that mocks the DB to return a record with a mismatched tenant_id and asserts the routing rejects it.

tests/unit/test_notifications.py
```diff
@@ -0,0 +1,317 @@
...
+
+            svc.create_smart_notification.assert_called_once()
+            call_kwargs = svc.create_smart_notification.call_args.kwargs
+            assert call_kwargs["summarized_content"] == "Deal closed for $50k"
+            assert call_kwargs["priority"] == 1
+            assert call_kwargs["channel"] == 0
+            assert call_kwargs["timing"] == 0
+            assert call_kwargs["tenant_id"] == 1
+            assert call_kwargs["recipient_filter"] is None
+
```
yihua-zhuo, line 177: **WARNING** (suggestion)

**Issue:** Test test_create_smart_notification_with_recipient_filter sends channel=3 (in_app) but _MockSmartNotificationModel defaults channel=0 (email) — the channel value is never asserted in the body, so the test does not verify that the request channel is preserved through to the stored record.

**Suggestion:** Assert body['data']['notification']['channel'] == 3 to verify the request channel value is persisted and returned in the response.

tests/unit/test_notifications.py
```diff
@@ -0,0 +1,317 @@
...
+# ---------------------------------------------------------------------------
+# POST /notifications/smart — happy path
+# ---------------------------------------------------------------------------
+
+
+class TestCreateSmartNotification:
+    def test_create_smart_notification_ok(self):
+        """Valid payload returns 200 with notification and deliveries keys."""
+        with (
+            patch("api.routers.notifications.NotificationService") as svc_cls,
```
yihua-zhuo, line 119: **INFO** (suggestion)

**Issue:** _MockSmartNotificationModel.to_dict() returns raw Python ints for priority/channel/timing, but SmartNotificationModel.to_dict() returns the PostgreSQL enum representation (which may be int or a string depending on SQLAlchemy config) — this difference means to_dict round-trip serialization is not being verified against the real model.

**Suggestion:** Make _MockSmartNotificationModel store priority/channel/timing as the actual Priority/Channel/Timing enum values so the serialization format matches the real ORM model's to_dict() output.

tests/unit/test_notifications.py
```diff
@@ -0,0 +1,317 @@
...
+    """Build a FastAPI app with the notifications router and given auth override."""
+    app = FastAPI()
+    app.include_router(notifications_router)
+    app.dependency_overrides[require_auth] = auth_override
+    app.dependency_overrides[get_db] = _mock_get_db
+
+    @app.exception_handler(AppException)
+    async def _handler(request, exc):
+        return JSONResponse(status_code=exc.status_code, content={"success": False, "message": exc.detail})
+
```
yihua-zhuo, line 101: **INFO** (suggestion)

**Issue:** _app() re-creates a TestClient on every call, which is correct per-TestClass pattern, but _make_test_app is called with a lambda that ignores its argument — the lambda pattern is correct but the inline lambda in _app (lambda: _make_auth_override(...)) discards the notifier callback parameter for consistency with how require_auth is called.

**Suggestion:** Align the lambda signature with how require_auth is called (async def require_auth(...)) or simplify to _make_auth_override directly as the override since neither takes parameters in this usage.

tests/unit/test_notifications.py
```diff
@@ -0,0 +1,317 @@
...
+            assert call_kwargs["tenant_id"] == 1
+            assert call_kwargs["recipient_filter"] is None
+
+            routing_svc.route.assert_called_once()
+            route_call = routing_svc.route.call_args
+            assert route_call.kwargs["tenant_id"] == 1
+            from api.routers.notifications import _MockRoutingRecord
+
+            routed_record = route_call.args[0]
+            assert isinstance(routed_record, _MockRoutingRecord)
```
yihua-zhuo: **CRITICAL** (critical)

**Issue:** No cross-tenant negative test exists for the smart notification endpoint. Per guideline 126, cross-tenant negative tests are required for new or changed multi-tenant behavior.

**Suggestion:** Add a test that creates a notification under tenant_id=1 then verifies a request with tenant_id=2 cannot access or influence it (e.g., patch the service to return a record with tenant_id=1 and assert routing svc is called with tenant_id=1, not tenant_id=2).

.plans/issue-595.md
yihua-zhuo, line 268: **INFO** (suggestion)

**Issue:** The Risks/Open Questions section correctly identifies the Priority IntEnum vs string mismatch, but the implementation steps do not describe the solution (the _MockRoutingRecord adapter + _PRIORITY_MAP + isinstance guard) that was actually used.

**Suggestion:** Add a step in §Implementation Steps describing the adapter class: _PRIORITY_MAP constant, _MockRoutingRecord class, and the isinstance guard before calling route().

.plans/issue-595.md
```diff
@@ -0,0 +1,221 @@
...
+        "success": True,
+        "data": {
+            "notification": record.to_dict(),
+            "deliveries": [d.model_dump() for d in deliveries],
+        },
+        "message": "Smart notification created and routed",
+    }
+```
+
+**Completion check**: `ruff check src/api/routers/notifications.py` exits 0.
```
yihua-zhuo: **INFO** (suggestion)

**Issue:** Completion check in Step 2 is 'ruff check src/api/routers/notifications.py' but CLAUDE.md §Full check pipeline uses 'ruff check src/', which is the project-wide lint command. The plan's narrower check may miss cross-file issues.

**Suggestion:** Use 'ruff check src/' as the completion check to match project convention.

src/api/routers/notifications.py
```diff
@@ -9,8 +9,10 @@
 from sqlalchemy.ext.asyncio import AsyncSession
 
 from db.connection import get_db
+from db.models.smart_notification import Channel, Priority, Timing
 from internal.middleware.fastapi_auth import AuthContext, require_auth
```
yihua-zhuo, line 13: **INFO** (suggestion)

**Issue:** AuthContext is imported but not used as a type annotation in the router — it is only accessed via the require_auth dependency. The import is harmless but unused as a type hint.

**Suggestion:** Remove the 'AuthContext' import, or annotate current_user: 'AuthContext = Depends(require_auth)' to make the import meaningful as a type annotation.

src/services/notification_service.py
```diff
@@ -170,6 +171,39 @@ async def delete_notification(self, notification_id: int, tenant_id: int) -> Not
         await self.session.flush()
         return notification
 
+    async def create_smart_notification(
```
yihua-zhuo, line 174: **WARNING** (suggestion)

**Issue:** The new create_smart_notification method has no unit test in tests/unit/test_notifications.py — only the router endpoint is tested via class-level patches, leaving the service's SQL generation and enum validation logic untested.

**Suggestion:** Add tests in tests/unit/ that use a mock session with a smart-notification handler to test create_smart_notification directly: valid call, invalid priority, invalid channel, invalid timing, and tenant isolation.

yihua-zhuo, line 174: **INFO** (suggestion)

**Issue:** The timing field accepted by create_smart_notification is passed to SmartNotificationModel but NotificationRoutingService.route() never reads it. This is acceptable per the plan ('this method only persists') but unused data is stored.

**Suggestion:** No action required; this is by design per the plan. Document in a comment if timing routing is planned for #41.

tests/unit/test_notifications.py
```diff
@@ -0,0 +1,317 @@
...
+            assert call_kwargs["timing"] == 0
+            assert call_kwargs["tenant_id"] == 1
+            assert call_kwargs["recipient_filter"] is None
+
+            routing_svc.route.assert_called_once()
+            route_call = routing_svc.route.call_args
+            assert route_call.kwargs["tenant_id"] == 1
+            from api.routers.notifications import _MockRoutingRecord
+
+            routed_record = route_call.args[0]
```
yihua-zhuo, line 182: **WARNING** (suggestion)

**Issue:** Test imports _MockRoutingRecord — a private class (underscore-prefixed) from the production router module — to assert isinstance checks, coupling the test tightly to a router implementation detail.

**Suggestion:** Remove the import of _MockRoutingRecord. Assert routing_record attributes directly (e.g., assert routed_record.priority == 'urgent' and routed_record.id == 7) without requiring the private class to be importable from the router.

tests/unit/test_notifications.py
```diff
@@ -0,0 +1,317 @@
...
+                "channel": 99,
+                "timing": 0,
+            },
+        )
+        assert response.status_code == 422
+        errors = response.json().get("detail", [])
+        error_fields = {e.get("loc")[-1] for e in errors}
+        assert "channel" in error_fields
+
+    def test_empty_summarized_content(self):
```
yihua-zhuo, line 283: **INFO** (suggestion)

**Issue:** SmartNotificationCreate.timing field has ge=0/le=1 Pydantic constraint and a validator, but no test exercises an out-of-range timing value (e.g., timing=99).

**Suggestion:** Add a validation test analogous to test_invalid_channel: send json with timing=99 and assert 422 with 'timing' in error_fields.

tests/unit/test_notifications.py
```diff
@@ -0,0 +1,317 @@
+"""Unit tests for POST /notifications/smart endpoint."""
```
yihua-zhuo, line 1: **INFO** (suggestion)

**Issue:** The test file exercises the /notifications/smart endpoint but never touches /notifications/send (NotificationCreate) or its content_keys_must_be_allowed validator that blocks credential-class fields in content.

**Suggestion:** Add a test for /notifications/send that exercises content validation, e.g., json with content containing 'password' should return 422.

tests/unit/test_notifications.py
```diff
@@ -0,0 +1,317 @@
...
+
+class TestCreateSmartNotificationRouting:
+    def test_empty_deliveries_still_returns_200(self):
+        """priority=normal with no user_id returns 200 with an empty deliveries list."""
+        with (
+            patch("api.routers.notifications.NotificationService") as svc_cls,
+            patch("api.routers.notifications.NotificationRoutingService") as routing_cls,
+        ):
+            svc = svc_cls.return_value
+            mock_record = _MockSmartNotificationModel({"priority": Priority.normal})
```
yihua-zhuo, line 319: **WARNING** (suggestion)

**Issue:** _MockSmartNotificationModel.channel and .timing store bare ints, but SmartNotificationCreate sends channel=0 (int) and the router passes it as-is to the service. The test mock mirrors this, but no test asserts that non-zero channel values route correctly through the system.

**Suggestion:** Add at least one test with channel=3 (in_app) or channel=2 (push) to verify the channel integer is faithfully passed through to the service and routing service, not just tested at channel=0.

tests/unit/test_notifications.py
```diff
@@ -0,0 +1,564 @@
...
+                    "priority": 1,
+                    "channel": 0,
+                    "timing": 0,
+                },
+            )
+            assert response.status_code == 200
+            body = response.json()
+            assert body["success"] is True
+            assert "notification" in body["data"]
+            assert "deliveries" in body["data"]
```
yihua-zhuo, line 162: **CRITICAL** (critical)

**Issue:** route_call.kwargs is empty dict — tenant_id is passed positionally in NotificationRoutingService.route(self, notification, tenant_id), so .kwargs['tenant_id'] raises KeyError.

**Suggestion:** Replace route_call.kwargs['tenant_id'] with list(route_call.kwargs.keys()) or check route_call.args directly.

tests/unit/test_notifications.py
```diff
@@ -0,0 +1,564 @@
...
+            assert body["data"]["notification"]["id"] == 7
```
yihua-zhuo, line 163: **CRITICAL** (critical)

**Issue:** route_call.args[0] is the MagicMock auto-created by routing_cls.return_value, not the actual _MockRoutingRecord produced by the router — assertions on routed_record.id and routed_record.priority are vacuously true against mock attributes.

**Suggestion:** Assert routing_svc.route.call_args list instead: assert len(route_call.args) == 2; record_arg = route_call.args[0]; assert record_arg.id == 7; assert record_arg.priority == 'urgent'.

tests/unit/test_notifications.py
```diff
@@ -0,0 +1,564 @@
...
+            assert len(body["data"]["deliveries"]) == 2
```
yihua-zhuo, line 164: **CRITICAL** (critical)

**Issue:** Same issue as line 162: assert route_call.kwargs['tenant_id'] == 1 raises KeyError because tenant_id is passed positionally.

**Suggestion:** Assert tenant_id positionally: assert route_call.args[1] == 1.

tests/unit/test_notifications.py
```diff
@@ -0,0 +1,564 @@
...
+            assert body["data"]["deliveries"][0]["channel"] == "in_app"
+            assert body["data"]["deliveries"][1]["channel"] == "email"
+            assert body["message"] == "Smart notification created and routed"
```
yihua-zhuo, line 167: **CRITICAL** (critical)

**Issue:** isinstance(routed_record, _MockRoutingRecord) always passes because routing_cls.return_value.route returns a MagicMock auto-created by MagicMock.__getattr__, and routing_cls.return_value.route.call_args.args[0] is that same MagicMock, not a real _MockRoutingRecord.

**Suggestion:** Inspect the actual call arguments directly from the router's perspective, or assert specific attributes on routing_svc.route.call_args list to avoid mocking-incorrect assertions.

.plans/issue-595.md
```diff
@@ -0,0 +1,269 @@
...
+
+## Goal
+
+Add `POST /notifications/smart` to the existing `notifications_router` at `src/api/routers/notifications.py`. The endpoint accepts a pre-classified event payload, persists a `SmartNotificationModel` record, calls `NotificationRoutingService.route()`, and returns the delivery status. It requires `AuthContext`, enforces tenant isolation, and includes a `TODO` comment pointing to issue #41 for future LLM integration. Unit tests live in `tests/unit/test_notifications.py`.
+
+---
+
+## Source Contract
+
+Dev-plan target: `/home/runner/work/agent-job/agent-job/docs/dev-plan/40-campaigns/0595-add-post-notifications-smart-endpoint.md`
```
yihua-zhuo, line 15: **WARNING** (suggestion)

**Issue:** Plan references a source dev-plan document at `docs/dev-plan/40-campaigns/0595-add-post-notifications-smart-endpoint.md` that does not exist in the repository. The plan also lists this path as part of its Source Contract, yet the file cannot be read or verified.

**Suggestion:** Remove or correct the Source Contract section pointing to the missing file. The plan itself should be self-contained and not require external documents that may not exist.

src/api/routers/notifications.py
```diff
@@ -78,6 +80,69 @@ class ReminderCreate(BaseModel):
...
+
+    def __init__(self, record):
+        self.id = record.id
+        self.tenant_id = record.tenant_id
+        self.priority = _priority_to_string(record.priority)
+        self.channel = record.channel
+        self.timing = record.timing
+        self.summarized_content = record.summarized_content
+        self.recipient_filter = record.recipient_filter
+
```
yihua-zhuo, line 145: **WARNING** (suggestion)

**Issue:** PEP 8 violation: no blank line after class definition `_MockRoutingRecord` before the next section comment (rule #20).

**Suggestion:** Add a blank line between line 145 and line 146.

src/services/notification_service.py
```diff
@@ -170,6 +171,39 @@ async def delete_notification(self, notification_id: int, tenant_id: int) -> Not
...
+        timing: int,
+        tenant_id: int,
+        recipient_filter: dict | None = None,
+    ) -> SmartNotificationModel:
+        """Persist a smart notification record and return it.
+
+        Actual channel routing is performed by NotificationRoutingService.route()
+        in the router — this method only persists.
+        """
+        if priority not in {p.value for p in Priority}:
```
yihua-zhuo, line 188: **INFO** (suggestion)

**Issue:** `create_smart_notification` re-validates priority/channel/timing integers that Pydantic validators in `SmartNotificationCreate` already checked at the router layer (rule #127: routers own validation). These service-side checks are defense-in-depth with no test coverage.

**Suggestion:** Remove the three ValidationException blocks in `create_smart_notification` and rely solely on the Pydantic schema, or add unit tests for the ValidationException paths.

tests/unit/test_notifications.py
```diff
@@ -0,0 +1,564 @@
...
+        """timing outside {0,1} returns 422."""
+        client = _app()
+        response = client.post(
+            "/api/v1/notifications/smart",
+            json={
+                "summarized_content": "Test",
+                "priority": 1,
+                "channel": 0,
+                "timing": 99,
+            },
```
yihua-zhuo, line 276: **WARNING** (suggestion)

**Issue:** Docstring claims 'The real NotificationRoutingService.route() is called' (line 280), but NotificationRoutingService is not patched, so the real service IS called. However, _mock_get_db() yields make_mock_session([]) which has no DB handlers — the real service may execute real SQL (getattr calls) against an unprepared mock session.

**Suggestion:** Either patch NotificationRoutingService like other tests do, or document clearly that the test exercises the real service with a mock session and accept that some routing paths may raise AttributeError against the bare mock.

tests/unit/conftest.py
```diff
@@ -20,6 +20,7 @@
 from dotenv import load_dotenv
 from sqlalchemy import insert, select, table, text
 from sqlalchemy.exc import CompileError, MultipleResultsFound, UnsupportedCompilationError
```
yihua-zhuo, line 22: **CRITICAL** (critical)

**Issue:** conftest.py re-exports make_smart_notification_handler directly, violating project rule 125 which requires domain-owned handlers to live in domain files and prohibits editing conftest.py for domain changes.

**Suggestion:** Remove make_smart_notification_handler from conftest.py re-exports; tests/unit/test_notifications.py already imports it directly from tests/unit/domain_handlers/notification.py, which is the correct path per rule 125.

.plans/issue-595.md
```diff
@@ -0,0 +1,269 @@
...
+        "data": {
+            "notification": record.to_dict(),
+            "deliveries": [d.model_dump() for d in deliveries],
+        },
+        "message": "Smart notification created and routed",
+    }
+```
+
+#### Priority IntEnum → string adapter (part of Step 2)
+
```
yihua-zhuo, line 152: **INFO** (suggestion)

**Issue:** The Priority IntEnum → string adapter section describes an approach that was implemented correctly in the router, but the plan does not verify whether the same conversion is needed in the service layer when the ORM object is constructed vs when routing occurs — this distinction is not clearly documented.

**Suggestion:** Clarify in the plan that the enum-to-string conversion for priority only needs to happen at the routing step (router layer), not in the service layer. The service stores the integer; routing converts it.

.plans/issue-595.md
```diff
@@ -0,0 +1,269 @@
...
+## Acceptance Criteria
+
+- `POST /api/v1/notifications/smart` with a valid payload returns `200` with `{"success": true, "data": {"notification": {...}, "deliveries": [...]}, "message": "..."}`
+- Pydantic schema rejects missing `summarized_content`, out-of-range `priority`, and out-of-range `channel` with `422`
+- `tenant_id=0` returns `401`
+- `NotificationRoutingService.route()` is called with the persisted record and `tenant_id`
+- `ruff check src/` exits 0
+- `PYTHONPATH=src pytest tests/unit/test_notifications.py -v` → all passed
+- A `TODO` comment pointing to issue #41 is present in the router or service
```
yihua-zhuo, line 262: **INFO** (suggestion)

**Issue:** The acceptance criterion 'TODO comment pointing to issue #41' is a non-machine-checkable social contract; there is no automated test or lint rule enforcing that the TODO remains.

**Suggestion:** Consider adding a code comment with an explicit TODO marker or tracking the TODO in the external issue tracker rather than relying on a static code comment.

src/api/routers/notifications.py
```diff
@@ -9,8 +9,10 @@
 from sqlalchemy.ext.asyncio import AsyncSession
 
 from db.connection import get_db
+from db.models.smart_notification import Channel, Priority, Timing
 from internal.middleware.fastapi_auth import AuthContext, require_auth
 from pkg.constants.notification_constants import VALID_NOTIFICATION_CHANNELS
```
yihua-zhuo, line 14: **WARNING** (suggestion)

**Issue:** Import statement for ChannelDelivery was removed during implementation but the plan shows it should be present; the router uses ChannelDelivery only in type-less route() return values, so it is functionally unused but creates confusion.

**Suggestion:** Remove the unused import: delete line 14 'from models.channel_delivery import ChannelDelivery' — the model is constructed in the routing service, not the router.

src/api/routers/notifications.py
```diff
@@ -78,6 +80,69 @@ class ReminderCreate(BaseModel):
...
+    def __init__(self, record):
+        self.id = record.id
+        self.tenant_id = record.tenant_id
+        self.priority = _priority_to_string(record.priority)
+        self.channel = record.channel
+        self.timing = record.timing
+        self.summarized_content = record.summarized_content
+        self.recipient_filter = record.recipient_filter
+
 # ---------------------------------------------------------------------------
```
yihua-zhuo, line 146: **INFO** (suggestion)

**Issue:** Missing blank line between _MockRoutingRecord class and the Notification endpoints comment — style inconsistency with the rest of the file (every other section break has a blank line).

**Suggestion:** Add a blank line before line 146 '# ---------------------------------------------------------------------------' for consistent formatting.

tests/unit/test_notifications.py
```diff
@@ -0,0 +1,564 @@
...
+            "timing": self.timing,
+            "recipient_filter": self.recipient_filter,
+            "created_at": self.created_at.isoformat() if self.created_at else None,
+        }
+
+    def __iter__(self):
+        yield from self.to_dict().items()
+
+
+class _TestRoutingRecord:
```
yihua-zhuo, line 67: **WARNING** (suggestion)

**Issue:** _TestRoutingRecord class is defined but never instantiated or called anywhere in the test file — dead code per rule 17.

**Suggestion:** Remove the _TestRoutingRecord class if it is not used, or remove the import of _priority_to_string from line 14 if _TestRoutingRecord is the only consumer.

tests/unit/test_notifications.py
```diff
@@ -0,0 +1,564 @@
...
+                summarized_content="Test",
+                priority=0,
+                channel=0,
+                timing=99,
+                tenant_id=1,
+            )
+        assert "timing" in str(exc_info.value.detail)
+
+    @pytest.mark.asyncio
+    async def test_tenant_isolation(self, mock_db_session):
```
yihua-zhuo, line 538: **INFO** (suggestion)

**Issue:** test_tenant_isolation documents that the mock handler 'does not filter by tenant on read' — this is a known limitation. The comment correctly notes the service enforces tenant scoping, but a future handler bug could go undetected.

**Suggestion:** No action needed for existing tests; be aware this comment documents a gap in mock coverage rather than a test deficiency.

PR #787, opened 2026-06-04
Auto: implement issue #606

### LINKED ISSUE #606: Add accuracy metrics endpoint and basic reporting
Subtask of #45
Depends on #605

Add `GET /tickets/categorization/metrics` endpoint returning: total categorized, override count, override rate (%), average confidence, breakdown by type and priority. Query the new models and compute stats in a service method. Unit test the metrics response shape.

Dev-Plan: docs/dev-plan/30-tickets/0606-add-accuracy-metrics-endpoint-and-basic-reporting.md

yihua-zhuo:
Closes #606

Plan: `.plans/issue-606.md`

1. **Plan** — Claude wrote a detailed implementation plan to `.plans/issue-606.md`.

Please review before merging.

## Summary by CodeRabbit

* **New Features**
  * Added endpoint to retrieve ticket categorization metrics, including overall accuracy statistics, human override information, confidence scores, and detailed breakdowns by category type and priority.

* **Tests**
  * Added comprehensive unit tests for metrics retrieval, response validation, and service integration.

### DIFF
--- a/.plans/issue-606.md
+++ b/.plans/issue-606.md
@@ -1,0 +1,89 @@
+I have all the information needed. Here is the implementation plan:
+
+# Implementation Plan — Issue #606
+
+## Goal
+Add a `GET /tickets/categorization/metrics` endpoint that returns aggregated accuracy metrics for the LLM-based ticket categorization system. The endpoint exposes total categorized count, human override count/rate, average confidence, and breakdowns by category_type and priority, all scoped to the authenticated tenant. This enables dashboards/reporting to quantify categorization quality.
+
+## Source Contract
+Dev-plan target: `/home/runner/work/agent-job/agent-job/docs/dev-plan/30-tickets/0606-add-accuracy-metrics-endpoint-and-basic-reporting.md`
+Template depth: `deep`
+Reading order followed:
+1. `/home/runner/work/agent-job/agent-job/docs/dev-plan/README.md`
+2. `/home/runner/work/agent-job/agent-job/docs/dev-plan/_template-deep.md`
+3. `/home/runner/work/agent-job/agent-job/docs/dev-plan/30-tickets/0606-add-accuracy-metrics-endpoint-and-basic-reporting.md`
+
+## Affected Files
+- `src/services/ticket_categorization_service.py` — **new file**: add `get_metrics(tenant_id)` method that runs three SQL aggregations (overall, by `category_type`, by `priority`) against `TicketCategorizationModel` and returns a `dict` with `total_categorized`, `override_count`, `override_rate`, `average_confidence`, `by_type`, `by_priority`.
+- `src/api/routers/tickets.py` — append `GET /tickets/categorization/metrics` endpoint after line 496 (after `patch_categorization_feedback`). The import `from services.ticket_categorization_service import TicketCategorizationService` already exists at line 16. Follow the same `tickets_router` + `Depends(require_auth)` + `Depends(get_db)` pattern used by `get_sla_summary` at lines 458–470.
+- `tests/unit/test_ticket_categorization_metrics.py` — **new file**: unit tests for `TicketCategorizationService.get_metrics()` covering happy path, empty result (zero division guard), and NULL confidence (coalesce guard).
+- `tests/unit/test_tickets_router.py` — add a `test_get_categorization_metrics` function (appended after line 815) using the existing `client_with_service` fixture pattern (see the `test_*_sla_summary` pattern at lines 760–815) that patches `TicketCategorizationService.get_metrics` and asserts the JSON envelope.
+
+## Implementation Steps
+
+1. **Add `get_metrics` method to `TicketCategorizationService` in `src/services/ticket_categorization_service.py`**
+   - The file already exists; append the method to the existing `TicketCategorizationService` class.
+   - `__init__` already takes `session: AsyncSession` with no default — do not change.
+   - Run three queries, all `WHERE tenant_id = :tenant_id`:
+     - **Overall**: `func.count(id)`, `func.coalesce(func.avg(confidence), 0.0)`, `func.sum(case((human_override == True, 1), else_=0))`.
+     - **By category_type**: same three aggregations, `.group_by(category_type)`.
+     - **By priority**: same three aggregations, `.group_by(priority)`.
+   - Compute `override_rate = override_count / total if total > 0 else 0.0` in Python (avoid `None`/division).
+   - Return `dict` with keys: `total_categorized` (int), `override_count` (int), `override_rate` (float, 4 dp), `average_confidence` (float, 4 dp), `by_type` (dict keyed by category_type string), `by_priority` (dict keyed by priority string). Each breakdown value is a dict `{count, avg_confidence, overrides}`.
+   - Do **not** call `.to_dict()` — there is no ORM object to serialize; return a plain dict (matches dev-plan §4.1).
+   - Do **not** raise `AppException` for empty results — return zeros.
+
+2. **Add `GET /tickets/categorization/metrics` endpoint to `src/api/routers/tickets.py`**
+   - Append after the last existing endpoint (`patch_categorization_feedback` at line 496).
+   - Use the same dependency pattern as `get_sla_summary`:
+     ```python
+     @tickets_router.get("/tickets/categorization/metrics")
+     async def get_categorization_metrics(
+         ctx: AuthContext = Depends(require_auth),
+         session: AsyncSession = Depends(get_db),
+     ):
+         """Return categorization accuracy metrics for the current tenant."""
+         svc = TicketCategorizationService(session)
+         metrics = await svc.get_metrics(tenant_id=ctx.tenant_id or 0)
+         return {"success": True, "data": metrics}
+     ```
+   - `tickets_router` is defined with `prefix="/api/v1"` (line 20), so the full path is `/api/v1/tickets/categorization/metrics` — matches the dev-plan.
+   - The import for `TicketCategorizationService` already exists at line 16 — no new import needed.
+
+3. **Create `tests/unit/test_ticket_categorization_metrics.py`**
+   - Define `MockResult` (with `.one()`) and `MockScalarResult` (with `.all()`) helpers locally — do not import from `tests/unit/conftest.py` (keep the test self-contained, matching the dev-plan example).
+   - Use `AsyncMock` for the session. Set up a `side_effect` list of three return values for the three sequential `session.execute()` calls (overall, by_type, by_priority).
+   - **Test 1 — `test_get_metrics_returns_structure`**: assert all six top-level keys present, `override_rate` is a float, `by_type` / `by_priority` are dicts.
+   - **Test 2 — `test_get_metrics_empty_result`**: mock `total=0`, `override_count=0`, `avg_confidence=None`; assert `override_rate == 0.0`, `total_categorized == 0`, `average_confidence == 0.0` (coalesce).
+   - **Test 3 — `test_get_metrics_null_confidence_defaults_to_zero`**: mock overall row with `avg_confidence=None` and total > 0; assert `average_confidence == 0.0`.
+   - All tests pass `tenant_id=1` — verify the WHERE clause is applied by checking the session mock receives a SQLAlchemy `select` with the tenant filter (assertion optional; the main contract is correct return shape).
+
+4. **Append `test_get_categorization_metrics` to `tests/unit/test_tickets_router.py`**
+   - Follow the pattern from the `test_requires_auth` (line 800) and the SLA summary tests (lines 760–798) in the same file.
+   - Use the existing `client_with_service` fixture (line 111) which already wires up a mock `TicketCategorizationService` — or monkeypatch `api.routers.tickets.TicketCategorizationService` directly.
+   - Mock `TicketCategorizationService.get_metrics` with `AsyncMock(return_value=mock_metrics_dict)`.
+   - Assert `response.status_code == 200`, `data["success"] is True`, and all expected fields are present in `data["data"]`.
+
+## Test Plan
+- Unit tests in `tests/unit/`:
+  - **New: `tests/unit/test_ticket_categorization_metrics.py`** — 3 tests for `TicketCategorizationService.get_metrics()`: structure shape, empty-result zero-division guard, NULL-confidence coalesce guard. All use `AsyncMock` for the session, no DB.
+  - **Modified: `tests/unit/test_tickets_router.py`** — add 1 test `test_get_categorization_metrics` validating the HTTP envelope (`{"success": true, "data": {...}}`) and that `get_metrics` is called with `tenant_id=1`.
+- Integration tests in `tests/integration/`: **none** — the dev-plan §5 has no Step for an integration test, and the dev-plan §6 acceptance list does not include one. The dev-plan budget is 0.5–1 day; unit coverage of the service + router is the contract.
+- Dev-plan verification (from §6 acceptance):
+  1. `ruff check src/services/ticket_categorization_service.py` → 0 errors
+  2. `ruff check src/api/routers/tickets.py` → 0 errors
+  3. `PYTHONPATH=src pytest tests/unit/test_ticket_categorization_metrics.py -v` → ≥ 3 passed
+  4. `PYTHONPATH=src pytest tests/unit/test_tickets_router.py::test_get_categorization_metrics -v` → passed
+  5. `curl http://localhost:8000/api/v1/tickets/categorization/metrics -H "Authorization: Bearer …"` → 200 with JSON containing all six keys
+
+## Acceptance Criteria
+- `GET /api/v1/tickets/categorization/metrics` returns 200 with `{"success": true, "data": {total_categorized, override_count, override_rate, average_confidence, by_type, by_priority}}`.
+- All aggregate values are filtered by `tenant_id` derived from the authenticated context.
+- When `total_categorized == 0`, `override_rate` is `0.0` (no `ZeroDivisionError`).
+- When all `confidence` values are `NULL`, `average_confidence` is `0.0` (coalesce applied at SQL level).
+- `ruff check src/services/ticket_categorization_service.py src/api/routers/tickets.py` → 0 errors.
+- `PYTHONPATH=src pytest tests/unit/test_ticket_categorization_metrics.py -v` → 3 passed.
+- `PYTHONPATH=src pytest tests/unit/test_tickets_router.py::test_get_categorization_metrics -v` → passed.
+
+## Risks / Open Questions
+- **Column name mismatch between dev-plan code and actual model.** The dev-plan's Step 1 example code references `TicketCategorizationModel.category` and `TicketCategorizationModel.overridden`. The actual model (created by #605 at `src/db/models/ticket_categorization.py:13`) uses `category_type` and `human_override`. The implementation must use the real column names (`category_type` for the type-breakdown `.group_by()`, `human_override` for the override count), not the names in the dev-plan example. The plan uses the real names.
--- a/src/api/routers/tickets.py
+++ b/src/api/routers/tickets.py
@@ -494,3 +494,14 @@ async def patch_categorization_feedback(
         corrected_priority=body.priority,
     )
     return {"success": True, "data": feedback.to_dict()}
+
+
+@tickets_router.get("/tickets/categorization/metrics")
+async def get_categorization_metrics(
+    ctx: AuthContext = Depends(require_auth),
+    session: AsyncSession = Depends(get_db),
+):
+    """Return categorization accuracy metrics for the current tenant."""
+    svc = TicketCategorizationService(session)
+    metrics = await svc.get_metrics(tenant_id=ctx.tenant_id or 0)
+    return {"success": True, "data": metrics}
--- a/src/services/ticket_categorization_service.py
+++ b/src/services/ticket_categorization_service.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 from datetime import UTC, datetime
 from decimal import Decimal
 
-from sqlalchemy import and_, select
+from sqlalchemy import and_, case, func, select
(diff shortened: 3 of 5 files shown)


.plans/issue-606.md
```diff
@@ -0,0 +1,89 @@
...
+## Source Contract
+Dev-plan target: `/home/runner/work/agent-job/agent-job/docs/dev-plan/30-tickets/0606-add-accuracy-metrics-endpoint-and-basic-reporting.md`
+Template depth: `deep`
+Reading order followed:
+1. `/home/runner/work/agent-job/agent-job/docs/dev-plan/README.md`
+2. `/home/runner/work/agent-job/agent-job/docs/dev-plan/_template-deep.md`
+3. `/home/runner/work/agent-job/agent-job/docs/dev-plan/30-tickets/0606-add-accuracy-metrics-endpoint-and-basic-reporting.md`
+
+## Affected Files
+- `src/services/ticket_categorization_service.py` — **new file**: add `get_metrics(tenant_id)` method that runs three SQL aggregations (overall, by `category_type`, by `priority`) against `TicketCategorizationModel` and returns a `dict` with `total_categorized`, `override_count`, `override_rate`, `average_confidence`, `by_type`, `by_priority`.
```
yihua-zhuo, line 17: **INFO** (suggestion)

**Issue:** Plan states the service 'returns a dict' but per Rule 127, services should return domain objects or ORM results — returning a raw dict from a service is borderline; however since this is a pure aggregation with no ORM object, a dict is acceptable here. No action needed.

**Suggestion:** No change required; this is an appropriate exception to the rule for aggregation results.

.plans/issue-606.md
```diff
@@ -0,0 +1,89 @@
...
+     ```python
+     @tickets_router.get("/tickets/categorization/metrics")
+     async def get_categorization_metrics(
+         ctx: AuthContext = Depends(require_auth),
+         session: AsyncSession = Depends(get_db),
+     ):
+         """Return categorization accuracy metrics for the current tenant."""
+         svc = TicketCategorizationService(session)
+         metrics = await svc.get_metrics(tenant_id=ctx.tenant_id or 0)
+         return {"success": True, "data": metrics}
```
yihua-zhuo, line 48: **INFO** (suggestion)

**Issue:** Plan example code uses `ctx.tenant_id or 0` which silently coerces a null tenant to 0, potentially allowing cross-tenant data access if auth context is malformed. This is a defense-in-depth concern (Rule 126).

**Suggestion:** Document why `or 0` is used (it means a missing tenant returns an empty result set rather than an error) so future reviewers understand the intent.

src/api/routers/tickets.py
```diff
@@ -494,3 +494,14 @@ async def patch_categorization_feedback(
         corrected_priority=body.priority,
     )
     return {"success": True, "data": feedback.to_dict()}
+
+
+@tickets_router.get("/tickets/categorization/metrics")
```
yihua-zhuo, line 499: **WARNING** (suggestion)

**Issue:** The new endpoint `get_categorization_metrics` does not declare a `response_model` or Pydantic schema for its response. For a public API endpoint returning a fixed structured payload, an explicit schema (e.g., `CategorizationMetricsResponse`) would improve OpenAPI documentation and validation (Rule 93, 100).

**Suggestion:** Add a Pydantic model for the metrics response and pass it as `response_model` to the route decorator.

src/api/routers/tickets.py
```diff
@@ -494,3 +494,14 @@ async def patch_categorization_feedback(
...
+async def get_categorization_metrics(
+    ctx: AuthContext = Depends(require_auth),
+    session: AsyncSession = Depends(get_db),
+):
+    """Return categorization accuracy metrics for the current tenant."""
+    svc = TicketCategorizationService(session)
+    metrics = await svc.get_metrics(tenant_id=ctx.tenant_id or 0)
+    return {"success": True, "data": metrics}
```
yihua-zhuo, line 507: **WARNING** (suggestion)

**Issue:** Response envelope uses `{'success': True, 'data': metrics}` (Rule 127 requires `data` to be a serialised domain object or list). Returning a raw service dict is acceptable here since there is no ORM object, but it would be cleaner to wrap in a named model.

**Suggestion:** Define a `CategorizationMetricsResponse` Pydantic model to formalize the response shape.

src/api/routers/tickets.py
```diff
@@ -494,3 +494,14 @@ async def patch_categorization_feedback(
...
+
+
+@tickets_router.get("/tickets/categorization/metrics")
+async def get_categorization_metrics(
+    ctx: AuthContext = Depends(require_auth),
+    session: AsyncSession = Depends(get_db),
+):
+    """Return categorization accuracy metrics for the current tenant."""
+    svc = TicketCategorizationService(session)
+    metrics = await svc.get_metrics(tenant_id=ctx.tenant_id or 0)
```
yihua-zhuo, line 506: **INFO** (suggestion)

**Issue:** `ctx.tenant_id or 0` silently coerces a null tenant to 0; if the auth dependency ever returns `None` for `tenant_id`, this would allow queries scoped to tenant 0. This is consistent with other endpoints in this file so likely intentional, but the fallback warrants a brief comment (Rule 126).

**Suggestion:** Add a comment: `tenant_id=ctx.tenant_id or 0  # fallback to tenant 0 if unauthenticated (returns empty set)`

src/services/ticket_categorization_service.py
```diff
@@ -80,3 +80,69 @@ async def categorize_ticket(self, ticket_id: int, tenant_id: int) -> TicketCateg
         await self.session.flush()
         await self.session.refresh(record)
         return record
+
+    async def get_metrics(self, tenant_id: int) -> dict:
```
yihua-zhuo, line 84: **WARNING** (suggestion)

**Issue:** `get_metrics` returns a raw `dict` instead of a dataclass or Pydantic model. A typed return value would make the contract explicit, aid IDE completion, and serve as the single source of truth for the response schema (Rule 27, 127).

**Suggestion:** Define a `CategorizationMetrics` dataclass with fields: `total_categorized`, `override_count`, `override_rate`, `average_confidence`, `by_type`, `by_priority` and return an instance.

src/services/ticket_categorization_service.py
```diff
@@ -80,3 +80,69 @@ async def categorize_ticket(self, ticket_id: int, tenant_id: int) -> TicketCateg
...
+        total_result = await self.session.execute(
+            select(
+                func.count(TicketCategorizationModel.id).label("total"),
+                func.coalesce(func.avg(TicketCategorizationModel.confidence), 0.0).label("avg_confidence"),
+                func.sum(
+                    case((TicketCategorizationModel.human_override, 1), else_=0)
+                ).label("override_count"),
+            ).where(TicketCategorizationModel.tenant_id == tenant_id)
+        )
+        row = total_result.one()
```
yihua-zhuo, line 94: **INFO** (suggestion)

**Issue:** `total_result.one()` is used but the method could return `None` if the query produces no rows; the code accesses `row.total` unconditionally, which would raise `AttributeError` on `None` (Rule 41, 46).

**Suggestion:** Use `.one_or_none()` and handle the `None` case: `row = total_result.one_or_none(); if row is None: return _empty_metrics()`

src/services/ticket_categorization_service.py
```diff
@@ -80,3 +80,69 @@ async def categorize_ticket(self, ticket_id: int, tenant_id: int) -> TicketCateg
...
+        total = row.total or 0
+        override_count = int(row.override_count or 0)
```
yihua-zhuo, line 96: **INFO** (suggestion)

**Issue:** `func.count()` in SQLAlchemy always returns an integer, so `row.total or 0` is slightly defensive but fine. However `func.avg()` returns a float that could theoretically be `None` even with coalesce, and `row.override_count` from `func.sum()` of integers could also be `None` — the `or 0` guards handle this correctly (Rule 122).

**Suggestion:** No change needed; the coalesce + `or 0` pattern is correct.

src/services/ticket_categorization_service.py
```diff
@@ -80,3 +80,69 @@ async def categorize_ticket(self, ticket_id: int, tenant_id: int) -> TicketCateg
         await self.session.flush()
```
yihua-zhuo, line 80: **INFO** (suggestion)

**Issue:** `categorize_ticket` uses `flush() → refresh()` pattern (Rule 123 compliance — no `commit()` in service). Good. But line 81 calls `refresh(record)` after `flush()`; this is correct per the rule but adds a DB roundtrip that could be avoided if the caller doesn't need server-generated defaults. Minor performance note (Rule 36).

**Suggestion:** Acceptable as-is; `refresh` is needed to populate `created_at`/`updated_at` server defaults for the response.

src/services/ticket_categorization_service.py
```diff
@@ -80,3 +80,69 @@ async def categorize_ticket(self, ticket_id: int, tenant_id: int) -> TicketCateg
...
         await self.session.refresh(record)
         return record
+
+    async def get_metrics(self, tenant_id: int) -> dict:
+        total_result = await self.session.execute(
+            select(
+                func.count(TicketCategorizationModel.id).label("total"),
+                func.coalesce(func.avg(TicketCategorizationModel.confidence), 0.0).label("avg_confidence"),
+                func.sum(
+                    case((TicketCategorizationModel.human_override, 1), else_=0)
```
yihua-zhuo, line 90: **INFO** (suggestion)

**Issue:** The `case((..., 1), else_=0)` pattern is correct and the `func.sum` wrapper is proper SQLAlchemy 2.x usage (Rule 132).

tests/unit/test_ticket_categorization_metrics.py
```diff
@@ -0,0 +1,102 @@
+"""Unit tests for TicketCategorizationService.get_metrics()."""
+
+import pytest
+from unittest.mock import AsyncMock
+
+from services.ticket_categorization_service import TicketCategorizationService
+
+
+class MockResult:
```
yihua-zhuo, line 9: **WARNING** (suggestion)

**Issue:** Test defines its own `MockResult` and `MockScalarResult` classes locally instead of using the shared `MockRow`/`MockResult`/`MockScalarResult` primitives from `tests/unit/conftest.py` (Rule 135 — prefer composable mock state). This also means the mock doesn't validate bind parameter names (Rule 133).

**Suggestion:** Import the shared mock primitives from `tests.unit.conftest` and use `make_mock_session` with a custom handler, or at minimum import the existing `MockResult` class.

tests/unit/test_ticket_categorization_metrics.py
```diff
@@ -0,0 +1,102 @@
...
+    def all(self):
+        return self._rows
+
+    def __iter__(self):
+        return iter(self._rows)
+
+
+class TestGetMetrics:
+    @pytest.fixture
+    def mock_session(self):
```
yihua-zhuo, line 30: **WARNING** (suggestion)

**Issue:** Test uses `AsyncMock` directly for the session without any tenant-isolation enforcement in the mock (Rule 134). The `side_effect` list approach bypasses any handler that would verify `tenant_id` is passed as a bind parameter.

**Suggestion:** Add an assertion that `session.execute` is called with a SQLAlchemy `select` whose `.where()` clause includes `tenant_id == 1`, e.g. via `execute.call_args_list[0][0][0]` inspection or by using a handler that checks bind params.

tests/unit/test_ticket_categorization_metrics.py
```diff
@@ -0,0 +1,102 @@
...
+        return AsyncMock()
+
+    @pytest.fixture
+    def service(self, mock_session):
+        return TicketCategorizationService(mock_session)
+
+    async def test_get_metrics_returns_structure(self, service, mock_session):
```
yihua-zhuo, line 37: **INFO** (suggestion)

**Issue:** Test does not verify that the `tenant_id` bind parameter is correctly applied to any of the three SQL queries (Rule 133). The plan acknowledged this is optional but the test would be stronger with a `assert_called_with` on the first `execute` call's `where` clause.

**Suggestion:** Capture the `where` clause from the first `execute` call and assert it contains a `tenant_id == 1` predicate.

tests/unit/test_ticket_categorization_metrics.py
```diff
@@ -0,0 +1,102 @@
...
+        """Happy path: verify all required keys are present and types are correct."""
+        overall_row = type("Row", (), {"total": 100, "avg_confidence": 0.85, "override_count": 12})()
```
yihua-zhuo, line 39: **INFO** (suggestion)

**Issue:** Mock row uses `type('Row', (), {...})()` anonymous class pattern; this works but is fragile — any attribute access typo in the service would raise `AttributeError` rather than a clear test failure. A named class or `SimpleNamespace` would be slightly more maintainable (Rule 67).

**Suggestion:** Use `from types import SimpleNamespace; overall_row = SimpleNamespace(total=100, avg_confidence=0.85, override_count=12)` for readability.

tests/unit/test_ticket_categorization_metrics.py
```diff
@@ -0,0 +1,102 @@
...
+        assert "override_rate" in result
+        assert "average_confidence" in result
+        assert "by_type" in result
+        assert "by_priority" in result
+        assert isinstance(result["override_rate"], float)
+        assert isinstance(result["by_type"], dict)
+        assert isinstance(result["by_priority"], dict)
+        assert result["total_categorized"] == 100
+        assert result["override_count"] == 12
+        assert result["override_rate"] == 0.12
```
yihua-zhuo, line 67: **INFO** (suggestion)

**Issue:** Asserts `override_rate == 0.12` which depends on `12 / 100` and `round(..., 4)` — this is correct but a comment showing the expected math would help future maintainers.

**Suggestion:** Add a brief comment: `# 12 overrides / 100 total = 0.12`

tests/unit/test_ticket_categorization_metrics.py
```diff
@@ -0,0 +1,102 @@
...
+        result = await service.get_metrics(tenant_id=1)
+
+        assert result["override_rate"] == 0.0
+        assert result["total_categorized"] == 0
+        assert result["override_count"] == 0
+        assert result["average_confidence"] == 0.0
+        assert result["by_type"] == {}
+        assert result["by_priority"] == {}
+
+    async def test_get_metrics_null_confidence_defaults_to_zero(self, service, mock_session):
```
yihua-zhuo, line 89: **INFO** (suggestion)

**Issue:** Test name says 'null_confidence_defaults_to_zero' but the mock sets `avg_confidence=None` and `total=5` — the coalesce at the SQL level handles this, but the test doesn't verify what happens when `total > 0` and `avg_confidence` is genuinely `None` vs. `0.0`. The assertion `result['average_confidence'] == 0.0` is correct (Rule 144).

**Suggestion:** No change needed; the test adequately verifies the coalesce-to-zero behavior.

yihua-zhuo:

**Status:** `pass`
**Inline comments posted:** 16

**Summary:** Reviewed 5 file(s) between origin/master and origin/auto/issue-606.

`tests/unit/test_tickets_router.py` L118: 
**WARNING** (suggestion)

**Issue:** The fixture uses raw `MagicMock()`/`AsyncMock()` for the entire service layer instead of the composable `MockState` + domain handler pattern described in project rule 135. `tests/unit/conftest.py` provides `make_mock_session`, `make_ticket_handler`, etc., which would make tests more representative and catch real SQL-pattern drift.

**Suggestion:** Replace `MagicMock()` services with a mock session built via `make_mock_session([make_ticket_handler(state), make_ticket_reply_handler(state), ...])` and use `mock_db_session` to inject it. This also allows asserting bind parameters and tenant filtering. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`tests/unit/test_tickets_router.py` L163: 
**WARNING** (suggestion)

**Issue:** Nearly every endpoint test omits a `tenant_id` assertion on the service call. Per rule 126, tenant isolation is a security requirement, not a style choice. Only two tests (lines 768-777 and 855-863) verify `assert_awaited_once_with(tenant_id=1)`. The create, list, get, update, assign, reply, status, and categorize tests have no such guard.

**Suggestion:** Add `svc.create_ticket.assert_awaited_once_with(tenant_id=1, ...)` (or equivalent) to every test that calls a tenant-scoped endpoint. Use `call_args` to inspect keyword args. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`tests/unit/test_tickets_router.py` L660: 
**WARNING** (suggestion)

**Issue:** `MockCategorization` (line 660) and `CATEGORIZATION_ROW` (line 683) are defined after the test class that uses them (`TestCategorizeTicketEndpoint`, line 474). `MockActivity` (line 630) and `ACTIVITY_ROW` (line 648) are defined after `TestGetTicketActivityEndpoint` (line 700). Forward references work in Python but reduce readability (rules 11, 20).

**Suggestion:** Move each mock class and its data constant to immediately above the test class that consumes it, or group them by endpoint at the top of the file. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`tests/unit/test_tickets_router.py` L800: 
**WARNING** (suggestion)

**Issue:** `test_requires_auth` duplicates the entire FastAPI app + monkeypatch setup from `client_with_service` rather than reusing it. This violates DRY (rule 110) and makes it harder to keep auth-override behaviour in sync across tests.

**Suggestion:** Refactor the fixture to accept a parameter like `with_auth: bool = True` and clear the `require_auth` override conditionally, or expose a separate `unauth_client` fixture from the same setup. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`tests/unit/test_tickets_router.py` L122: 
**INFO** (suggestion)

**Issue:** `mock_user_service.get_user_by_id = AsyncMock(return_value=object())` returns a bare `object()`, making it impossible for any consumer to inspect the return value. If the router uses attributes of the returned user in serialization or further logic, the test silently papers over bugs.

**Suggestion:** Use a lightweight named mock or a simple class with the expected attributes (e.g. `SimpleNamespace(id=1, tenant_id=1)`). — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`tests/unit/test_tickets_router.py` L245: 
**INFO** (suggestion)

**Issue:** `test_auto_categorize_on_create_field_default_false` only checks the 201 status code. The test name promises to verify the default value, but the default (`False`) is never asserted on the service call.

**Suggestion:** Add `import_call = svc.create_ticket.call_args` and assert `import_call.kwargs.get('auto_categorize_on_create') is False` (or the corresponding positional arg). — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`tests/unit/test_tickets_router.py` L118: 
**INFO** (suggestion)

**Issue:** The mock services accept any call with any `tenant_id` without filtering, so a unit test can pass even if the router fails to forward the authenticated tenant. Per rule 134, mocks should preserve tenant isolation semantics where feasible.

**Suggestion:** When migrating to the handler pattern (issue above), configure the mock handlers to assert that `tenant_id` in the call matches the expected value, or at minimum add a wrapper that validates the bound parameter. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`tests/unit/test_tickets_router.py` L35: 
**INFO** (suggestion)

**Issue:** The `MockTicket.__init__` default-then-override pattern (lines 35-38) uses `hasattr` to avoid clobbering values from the data dict. This is needlessly clever for a test fixture (rule 15).

**Suggestion:** Simplify to `self.status = data.get('status', TicketStatus.OPEN)` inside a single dict-merge block, e.g. `self.__dict__.update({defaults, **(data or {})})`. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`tests/unit/test_tickets_router.py` L23: 
**INFO** (suggestion)

**Issue:** `tenant_id=1` and `user_id=99` are magic constants threaded through the entire test module without a named constant (rule 19). Changing the default auth context would require edits throughout.

**Suggestion:** Promote to module-level constants like `DEFAULT_TENANT_ID = 1` and `DEFAULT_USER_ID = 99` and use them in `_make_auth_ctx` and the assertions that depend on them. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_

PR #788, opened 2026-06-04
Auto: implement issue #642

### LINKED ISSUE #642: Add role and permission management API endpoints
Subtask of #38
Depends on #641

Create src/api/routers/rbac.py with: (1) GET /roles — list all roles for tenant, (2) POST /roles — create a custom role (body: name, permission list), admin+ only, (3) GET /roles/{id}/permissions — return permission list for role, (4) PUT /roles/{id}/permissions — update permission list for custom roles (admin+), (5) POST /users/{id}/role — assign role to user (admin+), (6) GET /permissions — list all available (resource, action) pairs. Create a RoleService in src/services/role_service.py. Use @require_permission where appropriate. Return standard envelope. No DB migration needed — uses models from task 1.

Dev-Plan: docs/dev-plan/70-platform/0642-add-role-and-permission-management-api-endpoints.md

yihua-zhuo:
Closes #642

Plan: `.plans/issue-642.md`

1. **Plan** — Claude wrote a detailed implementation plan to `.plans/issue-642.md`.

Please review before merging.

## Summary by CodeRabbit

* **New Features**
  * Added role management API for creating, listing, and managing custom roles.
  * Added ability to assign roles to users and update role permissions.
  * Added permission enumeration to view available system permissions.
  * Implemented tenant isolation to keep role configurations separate per organization.

### DIFF
--- a/.plans/issue-642.md
+++ b/.plans/issue-642.md
@@ -1,0 +1,98 @@
+I now have full context. Let me write the implementation plan.
+
+# Implementation Plan — Issue #642
+
+## Goal
+
+Create a `RoleService` and a 6-endpoint HTTP API for role and permission management under `/api/v1/rbac/`, enabling tenant administrators to list/create custom roles, manage role-to-permission assignments, assign roles to users, and enumerate all available system permissions. The endpoints in the dev-plan overlap significantly with the existing `RBACService` / `rbac_router` (introduced by #641) — the existing code already covers most endpoints. The new `RoleService` will be a thin facade that re-uses the existing DB-backed models (`RoleModel`, `PermissionModel`, `RolePermissionModel`, `UserRoleModel`) rather than introducing a new JSONB column design, and a new `RoleManagementRouter` in `src/api/routers/rbac.py` will expose the six endpoints described in the issue body, aligning with the path/method shape the issue requires.
+
+## Source Contract
+
+Dev-plan target: `/home/runner/work/agent-job/agent-job/docs/dev-plan/70-platform/0642-add-role-and-permission-management-api-endpoints.md`
+Template depth: `deep`
+Reading order followed:
+1. `/home/runner/work/agent-job/agent-job/docs/dev-plan/README.md`
+2. `/home/runner/work/agent-job/agent-job/docs/dev-plan/_template-deep.md`
+3. `/home/runner/work/agent-job/agent-job/docs/dev-plan/70-platform/0642-add-role-and-permission-management-api-endpoints.md`
+
+## Affected Files
+
+- `src/services/role_service.py` — **new**: `RoleService` class providing six methods (`list_roles`, `create_custom_role`, `get_role_permissions`, `update_role_permissions`, `assign_role_to_user`, `list_all_permissions`). Delegates the heavy lifting to existing `RBACService` or to direct ORM queries on `RoleModel` / `PermissionModel` / `RolePermissionModel` / `UserRoleModel`.
+- `src/api/routers/rbac.py` — **modify**: add a `RoleManagementRouter` (`APIRouter` with prefix `/api/v1/rbac`, tags `["rbac-roles"]`) containing the six endpoints the issue specifies. The existing `rbac_router` stays untouched (it was created by #641 and serves a different granularity of endpoints).
+- `src/dependencies/rbac.py` — **read only**: `require_permission(resource, action)` already exists and is used to gate admin+ endpoints.
+- `tests/unit/test_role_service.py` — **new**: unit tests for all six `RoleService` methods.
+- `tests/integration/test_rbac_integration.py` — **new**: end-to-end HTTP tests against the real Postgres-backed FastAPI app.
+
+## Implementation Steps
+
+1. **Create `src/services/role_service.py`** with `RoleService`:
+   - `__init__(self, session: AsyncSession)` — no default, per CLAUDE.md §Service Pattern.
+   - `list_roles(tenant_id: int) -> list[RoleModel]` — returns roles where `tenant_id == tenant_id OR tenant_id == 0` (system roles visible to all tenants), ordered by `priority DESC, id ASC`.
+   - `create_custom_role(tenant_id: int, name: str, permissions: list[str]) -> RoleModel` — creates a non-system role (`is_system=False`) and assigns the given permission names via `RolePermissionModel` rows. Raises `ValidationException` if any permission name does not exist in `PermissionModel`. Raises `ConflictException` if a role with the same `name` already exists for the tenant.
+   - `get_role_permissions(role_id: int, tenant_id: int) -> list[str]` — returns permission name strings (not ORM objects) for the given role, scoped to the tenant or system (`tenant_id == 0`).
+   - `update_role_permissions(role_id: int, tenant_id: int, permissions: list[str]) -> RoleModel` — replaces `RolePermissionModel` rows for the role. Raises `ForbiddenException` if `role.is_system == True`. Raises `ValidationException` if any permission name is unknown.
+   - `assign_role_to_user(user_id: int, role_id: int, tenant_id: int, granted_by: int) -> dict` — validates the user exists in the tenant, validates the role exists for the tenant or system, then inserts a `UserRoleModel` row (idempotent: if already assigned, returns `{"already_assigned": True}`).
+   - `list_all_permissions() -> list[dict[str, str]]` — returns the static `DEFAULT_PERMISSIONS` list from `services.rbac_service` as `[{"resource": ..., "action": ...}, ...]` pairs derived from the permission name `resource:action` format.
+   - Internal helpers: `_permission_pairs()` to derive `[{resource, action}]` from the permission name string by splitting on `:`.
+
+2. **Modify `src/api/routers/rbac.py`** to add a `RoleManagementRouter`:
+   - Define `role_management_router = APIRouter(prefix="/api/v1/rbac", tags=["rbac-roles"])`.
+   - `GET /roles` — calls `RoleService.list_roles(ctx.tenant_id)`, returns `{"success": True, "data": {"items": [r.to_dict() for r in roles], "total": len(roles)}}`.
+   - `POST /roles` — Pydantic body `CreateRoleRequest { name: str, permissions: list[str] }`. Gated with `Depends(require_permission("admin", "all"))` (admin+ only). Calls `RoleService.create_custom_role`.
+   - `GET /roles/{role_id}/permissions` — returns `{"success": True, "data": {"role_id": role_id, "permissions": [...]}}`.
+   - `PUT /roles/{role_id}/permissions` — body `UpdatePermissionsRequest { permissions: list[str] }`. Gated with `Depends(require_permission("admin", "all"))`. Calls `RoleService.update_role_permissions`.
+   - `POST /users/{user_id}/role` — body `AssignRoleRequest { role_id: int }`. Gated with `Depends(require_permission("admin", "all"))`. Calls `RoleService.assign_role_to_user`.
+   - `GET /permissions` — returns `{"success": True, "data": [{"resource": ..., "action": ...}, ...]}`.
+   - All routes inject `ctx: AuthContext = Depends(require_auth)` and `session: AsyncSession = Depends(get_db)`.
+   - Discovery note: `src/api/__init__.py::iter_routers()` yields all `APIRouter` instances whose attribute name is `router` or ends with `_router`. Since `rbac_router` (existing) and `role_management_router` (new) both use the prefix `/api/v1/rbac`, FastAPI will route them independently by the unique path/method pairs. No changes to `src/main.py` are required — the auto-discovery picks up both routers.
+
+3. **Add `tests/unit/test_role_service.py`** with `mock_db_session` fixture using `RBACMockState` and the auto-discovered handlers:
+   - `test_list_roles_returns_tenant_and_system_roles` — seed tenant-specific role + system role, verify both appear.
+   - `test_create_custom_role_persists_role_and_permission_links` — call `create_custom_role`, verify `state.roles` and `state.role_permissions` updated.
+   - `test_create_custom_role_raises_conflict_on_duplicate_name` — second call with same name raises `ConflictException`.
+   - `test_create_custom_role_raises_validation_on_unknown_permission` — pass `["nonexistent:perm"]`, expect `ValidationException`.
+   - `test_get_role_permissions_returns_name_list` — verify return type is `list[str]`, not ORM objects.
+   - `test_update_role_permissions_blocks_system_role` — call on `is_system=True` role, expect `ForbiddenException`.
+   - `test_update_role_permissions_replaces_existing_links` — verify old `role_permissions` rows for the role are deleted before new ones inserted.
+   - `test_assign_role_to_user_checks_tenant_membership` — seed user in a different tenant, expect `NotFoundException("用户")`.
+   - `test_assign_role_to_user_is_idempotent` — call twice, second returns `{"already_assigned": True}`.
+   - `test_list_all_permissions_returns_resource_action_pairs` — verify every entry is `{"resource": str, "action": str}` with non-empty values.
+
+4. **Add `tests/integration/test_rbac_integration.py`** using the real `api_client` fixture (which includes `auth_headers_web` for a seeded admin user):
+   - `test_list_roles_returns_empty_initially` — `GET /api/v1/rbac/roles` → `data.items == []`.
+   - `test_create_role_then_list_includes_it` — `POST /api/v1/rbac/roles` with `{"name": "custom_support", "permissions": ["customer:read"]}` → 201, then `GET` shows it.
+   - `test_create_role_with_unknown_permission_returns_422` — body includes `["fake:perm"]` → 422.
+   - `test_update_role_permissions_replaces_existing` — `PUT` new list, `GET` confirms replacement.
+   - `test_update_system_role_permissions_returns_403` — try to update `admin` (system role) → 403.
+   - `test_assign_role_to_user_persists_binding` — `POST /api/v1/rbac/users/{uid}/role` with `{"role_id": 2}` → 200; then `GET /api/v1/rbac/users/{uid}/roles` (existing endpoint) shows the binding.
+   - `test_list_permissions_returns_all_system_pairs` — `GET /api/v1/rbac/permissions` → ≥ 6 pairs, all `resource`/`action` non-empty.
+   - `test_tenant_isolation_roles` — create role in tenant 1, switch to `api_client_tenant_2`, verify `GET /roles` does not show tenant 1's custom role (only system roles shared via `tenant_id == 0`).
+
+## Test Plan
+
+- Unit tests in `tests/unit/test_role_service.py`: mock DB via `make_mock_session` + `RBACMockState` (auto-loaded from `tests/unit/domain_handlers/rbac.py`). Covers all six service methods across success, tenant-isolation, conflict, validation, and forbidden paths. Expected: `PYTHONPATH=src pytest tests/unit/test_role_service.py -v` → ≥ 10 passed.
+- Integration tests in `tests/integration/test_rbac_integration.py`: uses `api_client`, `auth_headers_web`, `api_client_tenant_2`, `db_schema`, `async_session` fixtures from `tests/integration/conftest.py`. Verifies HTTP round-trip including 201/200/403/422 status codes, envelope shape, and cross-tenant data isolation. Expected: `PYTHONPATH=src pytest tests/integration/test_rbac_integration.py -v` → ≥ 8 passed.
+- Dev-plan verification (§6): the target board lists these checks:
+  - `ruff check src/services/role_service.py src/api/routers/rbac.py` → 0 errors.
+  - `PYTHONPATH=src pytest tests/unit/test_role_service.py -v` → 6+ passed.
+  - `PYTHONPATH=src pytest tests/integration/test_rbac_integration.py -v` → 6+ passed.
+  - `PYTHONPATH=src mypy src/services/role_service.py src/api/routers/rbac.py` → 0 errors.
+  - `curl http://localhost:8000/api/v1/rbac/permissions` → `{"success": true, "data": [...]}`.
+
+## Acceptance Criteria
+
+- `ruff check src/services/role_service.py src/api/routers/rbac.py` → 0 errors.
+- `PYTHONPATH=src pytest tests/unit/test_role_service.py -v` → all tests passed.
+- `PYTHONPATH=src pytest tests/integration/test_rbac_integration.py -v` → all tests passed.
+- `GET /api/v1/rbac/roles` returns 200 with `{"success": true, "data": {"items": [...], "total": N}}` envelope.
+- `POST /api/v1/rbac/roles` with admin auth returns 201; without admin role returns 403.
+- `GET /api/v1/rbac/roles/{id}/permissions` returns `{"success": true, "data": {"role_id": id, "permissions": [string...]}}`.
+- `PUT /api/v1/rbac/roles/{id}/permissions` on a system role returns 403; on a custom role returns 200.
+- `POST /api/v1/rbac/users/{id}/role` returns 200 with `{"success": true, "data": {"user_id": id, "role_id": id}}`.
+- `GET /api/v1/rbac/permissions` returns 200 with `{"success": true, "data": [{"resource": ..., "action": ...}, ...]}` containing all 15 permissions from `DEFAULT_PERMISSIONS`.
+- All SQL queries filter by `tenant_id`; cross-tenant roles are not visible in `GET /roles`.
+
+## Risks / Open Questions
+
+- **Path collision with existing `rbac_router`**: Both `rbac_router` (from #641, prefix `/api/v1/rbac`) and the new `role_management_router` use the same prefix. FastAPI handles this by matching the first registered router that owns a given path+method pair. The existing `rbac_router` registers paths like `POST /roles`, `GET /roles/{role_id}`, `PUT /roles/{role_id}` — the new router must not duplicate these. The six endpoints in the issue body map to paths that are already covered by the existing router (`GET /roles`, `POST /roles`, `GET /roles/{id}/permissions`, `PUT /roles/{id}/permissions`, `POST /users/{id}/roles` vs issue's `POST /users/{id}/role`, `GET /permissions`). This raises a real risk: the new endpoints will conflict with existing ones, causing FastAPI to raise `AssertionError: route conflict` at startup. **Mitigation**: Register `role_management_router` with the same prefix only if path/method pairs are genuinely new; otherwise, the existing `rbac_router` already satisfies the issue. If a separate `RoleManagementRouter` is mandated by the issue contract, paths must be differentiated (e.g., add a `/v2/` or distinct prefix), or the new router should only add genuinely new routes not covered by `rbac_router` (e.g., `GET /permissions` is the only one not present in `rbac_router` since it lists `PermissionModel` paginated, not resource/action pairs).
+- **`@require_permission` gate semantics**: The existing `require_permission(resource, action)` dependency checks `ctx.roles` (a list of role names like `"admin"`, `"viewer"`) against `ROLE_PERMISSIONS` from `permission_service.py`. It does not consult the DB-backed `UserRoleModel` assignments. So a user assigned a custom role at runtime via the new `POST /users/{id}/role` endpoint will not be granted `admin:all` permission through the `require_permission` guard. This is a pre-existing limitation in the auth flow (not caused by this issue) but the dev-plan's claim that admin+ endpoints are protected by `@require_permission` is only true for users whose JWT contains `"admin"` in `roles` — not for users who get admin via the DB-backed custom-role mechanism.
--- a/src/api/routers/rbac.py
+++ b/src/api/routers/rbac.py
@@ -11,8 +11,10 @@ from sqlalchemy.ext.asyncio import AsyncSession
 from db.connection import get_db
 from internal.middleware.fastapi_auth import AuthContext, require_auth
 from services.rbac_service import RBACService
+from services.role_service import RoleService
 
 rbac_router = APIRouter(prefix="/api/v1/rbac", tags=["rbac"])
+role_management_router = APIRouter(prefix="/api/v1/rbac/mgmt", tags=["rbac-roles"])
 
 
 def _paginated(items, total, page, page_size):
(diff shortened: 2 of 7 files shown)


.plans/issue-642.md
```diff
@@ -0,0 +1,98 @@
...
...s by matching the first registered router that owns a given path+method pair. The existing `rbac_router` registers paths like `POST /roles`, `GET /roles/{role_id}`, `PUT /roles/{role_id}` — the new router must not duplicate these. The six endpoints in the issue body map to paths that are already covered by the existing router (`GET /roles`, `POST /roles`, `GET /roles/{id}/permissions`, `PUT /roles/{id}/permissions`, `POST /users/{id}/roles` vs issue's `POST /users/{id}/role`, `GET /permissions`). This raises a real risk: the new endpoints will conflict with existing ones, causing FastAPI to raise `AssertionError: route conflict` at startup. **Mitigation**: Register `role_management_router` with the same prefix only if path/method pairs are genuinely new; otherwise, the existing `rbac_router` already satisfies the issue. If a separate `RoleManagementRouter` is mandated by the issue contract, paths must be differentiated (e.g., add a `/v2/` or distinct prefix), or the new router should only add genuinely new routes not covered by `rbac_router` (e.g., `GET /permissions` is the only one not present in `rbac_router` since it lists `PermissionModel` paginated, not resource/action pairs).
```
yihua-zhuo, line 97: **CRITICAL** (critical)

**Issue:** The plan acknowledges a path-collision risk between rbac_router and role_management_router but claims it is mitigated; in reality the new router uses prefix '/api/v1/rbac/mgmt' (not '/api/v1/rbac') in code, contradicting the plan text which says the new router registers at the same prefix — the plan and implementation are out of sync.

**Suggestion:** Update the plan to reflect the actual /api/v1/rbac/mgmt prefix used in rbac.py:17, or align the implementation with the plan description.

src/api/routers/rbac.py
```diff
@@ -11,8 +11,10 @@
 from db.connection import get_db
 from internal.middleware.fastapi_auth import AuthContext, require_auth
 from services.rbac_service import RBACService
+from services.role_service import RoleService
 
 rbac_router = APIRouter(prefix="/api/v1/rbac", tags=["rbac"])
+role_management_router = APIRouter(prefix="/api/v1/rbac/mgmt", tags=["rbac-roles"])
```
yihua-zhuo, line 17: **CRITICAL** (critical)

**Issue:** rule 125: Two routers in the same file (rbac_router at /api/v1/rbac and role_management_router at /api/v1/rbac/mgmt) violate the project rule that feature domains should own their own router files — the mgmt router should live in its own file (e.g. src/api/routers/role_management.py) rather than being appended to rbac.py.

**Suggestion:** Move the RoleManagementRouter and its Pydantic schemas to src/api/routers/role_management.py and import it.

src/api/routers/rbac.py
```diff
@@ -254,3 +256,105 @@ async def list_users_with_role(
...
+    ctx: AuthContext = Depends(require_auth),
+    session: AsyncSession = Depends(get_db),
+):
+    svc = RoleService(session)
+    roles = await svc.list_roles(tenant_id=ctx.tenant_id)
+    items = [r.to_dict() for r in roles]
+    return {"success": True, "data": {"items": items, "total": len(items)}}
+
+
+@role_management_router.post("/roles", status_code=201)
```
yihua-zhuo, line 293: **CRITICAL** (critical)

**Issue:** rule 125/149: The 'mgmt_create_role' endpoint has no admin permission gate (no Depends(require_permission(...))), so any authenticated user can create custom roles and assign them to other users — the plan itself says this should be admin+ only.

**Suggestion:** Add Depends(require_permission('admin', 'all')) or a similar admin gate to mgmt_create_role, mgmt_update_role_permissions, and mgmt_assign_role.

src/api/routers/rbac.py
```diff
@@ -254,3 +256,105 @@ async def list_users_with_role(
...
+    svc = RoleService(session)
+    role = await svc.update_role_permissions(
+        role_id=role_id,
+        tenant_id=ctx.tenant_id,
+        permissions=body.permissions,
+    )
+    return {"success": True, "data": role.to_dict(), "message": "权限分配成功"}
+
+
+@role_management_router.post("/users/{user_id}/role")
```
yihua-zhuo, line 335: **CRITICAL** (critical)

**Issue:** rule 125/149: mgmt_assign_role has no admin permission gate — any authenticated user can assign any role to any other user in the tenant, a privilege-escalation vector.

**Suggestion:** Add Depends(require_permission('admin', 'all')) to the mgmt_assign_role endpoint.

yihua-zhuo, line 335: **CRITICAL** (critical)

**Issue:** rule 125/149: The singular '/users/{user_id}/role' path in mgmt_assign_role does not include the user_id in the body, but the path is not validated against the AuthContext's tenant — a user in tenant A could potentially assign roles to users in tenant B if tenant_id filtering in the service is bypassed.

**Suggestion:** Ensure RoleService.assign_role_to_user validates tenant_id on both user and role, and consider adding tenant_id to the API contract.

src/services/role_service.py
```diff
@@ -0,0 +1,206 @@
...
+
+    async def list_roles(self, tenant_id: int) -> list[RoleModel]:
+        result = await self.session.execute(
+            select(RoleModel)
+            .where(or_(RoleModel.tenant_id == tenant_id, RoleModel.tenant_id == 0))
+            .order_by(RoleModel.priority.desc(), RoleModel.id.asc())
+        )
+        return list(result.scalars().all())
+
+    async def create_custom_role(
```
yihua-zhuo, line 56: **CRITICAL** (critical)

**Issue:** rule 122: The service calls 'await self.session.flush()' on lines 89 and 93 (create_custom_role) and 152 (update_role_permissions) and 202 (assign_role_to_user) — flushing is acceptable per the rule, but the pattern of flush→add→flush→flush without an intervening refresh means auto-generated timestamps (created_at on UserRoleModel) may not be populated when the ORM object is returned to the router.

**Suggestion:** Add 'await self.session.refresh(role)' after the final flush in each method that returns the ORM object, so the caller sees the persisted state.

yihua-zhuo, line 56: **CRITICAL** (critical)

**Issue:** rule 123: update_role_permissions uses delete + add + flush, which is acceptable, but the method signature returns RoleModel and then the router does role.to_dict() — the role object was loaded before the delete, so the returned object is stale (no refreshed id/timestamp from the new permission links).

**Suggestion:** Return the role after refreshing it, or return a summary object that includes the new permission list.

src/services/role_service.py
```diff
@@ -0,0 +1,206 @@
...
+
+        await self.session.execute(
+            delete(RolePermissionModel).where(RolePermissionModel.role_id == role_id)
+        )
+        for perm in found:
+            self.session.add(RolePermissionModel(role_id=role_id, permission_id=perm.id))
+        await self.session.flush()
+        return role
+
+    async def assign_role_to_user(
```
yihua-zhuo, line 155: **CRITICAL** (critical)

**Issue:** rule 126: assign_role_to_user validates the user belongs to tenant_id and the role belongs to tenant_id or 0, but the UserRoleModel insert includes tenant_id from the parameter — this is correct, but there is no check that the user does not already have the same role with a different tenant_id, which could create cross-tenant role-leakage if the unique constraint on (user_id, role_id) does not include tenant_id.

**Suggestion:** Verify the UserRoleModel unique constraint includes tenant_id, or add a composite unique index.

tests/integration/test_rbac_integration.py
```diff
@@ -397,3 +402,172 @@ async def test_custom_role_invisible_to_other_tenant(self, db_schema, tenant_id,
...
+            select(RoleModel).where(RoleModel.name == "admin")
+        )
+        admin_role = result.scalar_one()
+        resp = await api_client.put(
+            f"/api/v1/rbac/mgmt/roles/{admin_role.id}/permissions",
+            json={"permissions": ["customer:read"]},
+        )
+        assert resp.status_code == 403
+
+    async def test_assign_role_to_user_persists_binding(
```
yihua-zhuo, line 489: **CRITICAL** (critical)

**Issue:** rule 124: test_assign_role_to_user_persists_binding uses both api_client (HTTP) and async_session (direct DB query) to look up entities, mixing HTTP testing with direct DB inspection — this is acceptable for setup, but the test never verifies the HTTP response body shape matches the API contract (no assertion on 'already_assigned' field or message).

**Suggestion:** Add assertions on body['data']['already_assigned'] and body['message'] to validate the full API response contract.

.plans/issue-642.md
```diff
@@ -0,0 +1,98 @@
+I now have full context. Let me write the implementation plan.
+
+# Implementation Plan — Issue #642
+
+## Goal
+
+Create a `RoleService` and a 6-endpoint HTTP API for role and permission management under `/api/v1/rbac/`, enabling tenant administrators to list/create custom roles, manage role-to-permission assignments, assign roles to users, and enumerate all available system permissions. The endpoints in the dev-plan overlap significantly with the existing `RBACService` / `rbac_router` (introduced by #641) — the existing code already covers most endpoints. The new `RoleService` will be a thin facade that re-uses the existing DB-backed models (`RoleModel`, `PermissionModel`, `RolePermissionModel`, `UserRoleModel`) rather than introducing a new JSONB column design, and a new `RoleManagementRouter` in `src/api/routers/rbac.py` will expose the six endpoints described in the issue body, aligning with the path/method shape the issue requires.
```
yihua-zhuo, line 7: **WARNING** (suggestion)

**Issue:** The plan says RoleService 're-uses existing RBACService', but the service docstring (role_service.py:9-10) explicitly states it does NOT re-use RBACService — contradiction between plan and implementation.

**Suggestion:** Update the plan to match the actual decision: RoleService uses direct ORM queries, not RBACService delegation.

.plans/issue-642.md
yihua-zhuo, line 97: **WARNING** (suggestion)

**Issue:** Plan claims path collision is mitigated by using '/users/{user_id}/role' (singular) vs '/users/{user_id}/roles' (plural), but FastAPI's router dispatch is per-prefix+routes; the real mitigation is the '/mgmt' sub-prefix added to role_management_router, not singular/plural path differences.

**Suggestion:** Correct the risk analysis: the /mgmt sub-prefix is what prevents route conflicts, not singular vs plural.

.plans/issue-642.md
```diff
@@ -0,0 +1,98 @@
...
+   - `assign_role_to_user(user_id: int, role_id: int, tenant_id: int, granted_by: int) -> dict` — validates the user exists in the tenant, validates the role exists for the tenant or system, then inserts a `UserRoleModel` row (idempotent: if already assigned, returns `{"already_assigned": True}`).
+   - `list_all_permissions() -> list[dict[str, str]]` — returns the static `DEFAULT_PERMISSIONS` list from `services.rbac_service` as `[{"resource": ..., "action": ...}, ...]` pairs derived from the permission name `resource:action` format.
+   - Internal helpers: `_permission_pairs()` to derive `[{resource, action}]` from the permission name string by splitting on `:`.
+
+2. **Modify `src/api/routers/rbac.py`** to add a `RoleManagementRouter`:
+   - Define `role_management_router = APIRouter(prefix="/api/v1/rbac", tags=["rbac-roles"])`.
+   - `GET /roles` — calls `RoleService.list_roles(ctx.tenant_id)`, returns `{"success": True, "data": {"items": [r.to_dict() for r in roles], "total": len(roles)}}`.
```
yihua-zhuo, line 40: **WARNING** (suggestion)

**Issue:** Plan specifies DELETE /api/v1/rbac/roles/{id} should return '200' (line 88 acceptance criteria) but the plan's step 2 does not include any DELETE endpoint for the six new endpoints — the acceptance criteria lists endpoints that don't exist in the implementation steps.

**Suggestion:** Reconcile the acceptance criteria with the actual six endpoints implemented; remove the DELETE criterion or add a DELETE step.

.plans/issue-642.md
```diff
@@ -0,0 +1,98 @@
...
+   - `test_update_role_permissions_replaces_existing` — `PUT` new list, `GET` confirms replacement.
+   - `test_update_system_role_permissions_returns_403` — try to update `admin` (system role) → 403.
+   - `test_assign_role_to_user_persists_binding` — `POST /api/v1/rbac/users/{uid}/role` with `{"role_id": 2}` → 200; then `GET /api/v1/rbac/users/{uid}/roles` (existing endpoint) shows the binding.
+   - `test_list_permissions_returns_all_system_pairs` — `GET /api/v1/rbac/permissions` → ≥ 6 pairs, all `resource`/`action` non-empty.
+   - `test_tenant_isolation_roles` — create role in tenant 1, switch to `api_client_tenant_2`, verify `GET /roles` does not show tenant 1's custom role (only system roles shared via `tenant_id == 0`).
+
+## Test Plan
+
+- Unit tests in `tests/unit/test_role_service.py`: mock DB via `make_mock_session` + `RBACMockState` (auto-loaded from `tests/unit/domain_handlers/rbac.py`). Covers all six service methods across success, tenant-isolation, conflict, validation, and forbidden paths. Expected: `PYTHONPATH=src pytest tests/unit/test_role_service.py -v` → ≥ 10 passed.
```
yihua-zhuo, line 73: **INFO** (suggestion)

**Issue:** Test plan claims unit tests use 'RBACMockState and auto-discovered handlers from tests/unit/domain_handlers/rbac.py', but the plan should verify this domain handler file exists and supports RoleService's queries (flush, delete, join on RolePermissionModel).

**Suggestion:** Verify tests/unit/domain_handlers/rbac.py provides flush, delete, and join handling for RolePermissionModel before relying on auto-discovery.

src/api/routers/rbac.py
```diff
@@ -254,3 +256,105 @@ async def list_users_with_role(
...
+
+# ---------------------------------------------------------------------------
+# RoleManagementRouter — thin facade with the six endpoints specified in #642.
+# Uses a different path shape (POST /users/{user_id}/role, singular) than
+# the existing rbac_router (POST /users/{user_id}/roles, plural) so the two
+# routers do not conflict at FastAPI registration time.
+# ---------------------------------------------------------------------------
+
+
+class CreateRoleRequest(BaseModel):
```
yihua-zhuo, line 269: **WARNING** (suggestion)

**Issue:** rule 136: CreateRoleRequest and UpdatePermissionsRequest define 'permissions: list[str] = Field(..., min_length=0)' — an empty list is allowed for update but a role with no permissions may be a logical error; the update endpoint should at least require min_length=1 to prevent removing all permissions.

**Suggestion:** Change UpdatePermissionsRequest.permissions to Field(..., min_length=1) to prevent zero-permission roles.

src/api/routers/rbac.py
yihua-zhuo, line 293: **WARNING** (suggestion)

**Issue:** rule 17: CreateRoleRequest does not include is_system, priority, display_name, or description fields that RoleModel supports — this is intentional simplification, but it means the mgmt API cannot fully manage roles, creating dead-end behavior for clients.

**Suggestion:** Document the limitation in the endpoint docstring or add optional fields for display_name and description.

src/api/routers/rbac.py
```diff
@@ -254,3 +256,105 @@ async def list_users_with_role(
...
+
+class UpdatePermissionsRequest(BaseModel):
+    permissions: list[str] = Field(..., min_length=0)
+
+
+class AssignRoleRequest(BaseModel):
+    role_id: int = Field(..., description="Role to assign to the user")
+
+
+@role_management_router.get("/roles")
```
yihua-zhuo, line 282: **WARNING** (suggestion)

**Issue:** rule 94: mgmt_list_roles does not implement pagination — returns all roles in a single response which could be a large payload for tenants with many custom roles.

**Suggestion:** Add page and page_size Query parameters consistent with rbac_router.list_roles.

src/api/routers/rbac.py
```diff
@@ -254,3 +256,105 @@ async def list_users_with_role(
...
+    svc = RoleService(session)
+    role = await svc.create_custom_role(
+        tenant_id=ctx.tenant_id,
+        name=body.name,
+        permissions=body.permissions,
+    )
+    return {"success": True, "data": role.to_dict(), "message": "角色创建成功"}
+
+
+@role_management_router.get("/roles/{role_id}/permissions")
```
yihua-zhuo, line 308: **INFO** (suggestion)

**Issue:** rule 15: mgmt_get_role_permissions and mgmt_list_role_permissions (existing) have overlapping semantics with slightly different return shapes (list[str] vs list[dict]) — this divergence is confusing for API consumers.

**Suggestion:** Unify the return shape across both endpoints or document the intentional difference.

src/services/role_service.py
```diff
@@ -0,0 +1,206 @@
...
+from db.models.user import UserModel
+from pkg.errors.app_exceptions import (
+    ConflictException,
+    ForbiddenException,
+    NotFoundException,
+    ValidationException,
+)
+
+
+def _permission_pairs() -> list[dict[str, str]]:
```
yihua-zhuo, line 28: **WARNING** (suggestion)

**Issue:** rule 146: _permission_pairs() imports DEFAULT_PERMISSIONS inside the function body (line 29) — this is fine for avoiding circular imports at module load, but it means the function is not import-safe if services.rbac_service has side effects on import.

**Suggestion:** Move the import to module level if there is no circular dependency, or add a comment explaining why it is deferred.

src/services/role_service.py
```diff
@@ -0,0 +1,206 @@
...
+        )
+        self.session.add(role)
+        await self.session.flush()
+
+        for perm in found:
+            self.session.add(RolePermissionModel(role_id=role.id, permission_id=perm.id))
+        await self.session.flush()
+        return role
+
+    async def get_role_permissions(self, role_id: int, tenant_id: int) -> list[str]:
```
yihua-zhuo, line 96: **WARNING** (suggestion)

**Issue:** rule 126: get_role_permissions validates the role exists for the tenant, but does not enforce that the role is active or that the calling user has permission to view its permissions — this is a minor authorization gap since role permissions are not considered secret, but the service should at least be consistent with the pattern of tenant-scoped reads.

**Suggestion:** Document the authorization model or add a check that the caller has at least one role in common with the target role.

src/services/role_service.py
```diff
@@ -0,0 +1,206 @@
...
+                )
+            )
+        )
+        role = role_result.scalar_one_or_none()
+        if role is None:
+            raise NotFoundException("角色")
+        if role.is_system:
+            raise ForbiddenException("系统角色不可修改权限")
+
+        permissions = list(dict.fromkeys(permissions))
```
yihua-zhuo, line 136: **INFO** (suggestion)

**Issue:** rule 15: 'permissions = list(dict.fromkeys(permissions))' silently deduplicates the input list — this is reasonable but should be documented in the method docstring so callers know the service mutates the parameter.

**Suggestion:** Add a docstring note that duplicate permission names are silently de-duplicated.

src/services/role_service.py
```diff
@@ -0,0 +1,206 @@
...
+                role_id=role_id,
+                tenant_id=tenant_id,
+                granted_by=granted_by,
+                granted_at=datetime.now(UTC),
+            )
+        )
+        await self.session.flush()
+        return {"user_id": user_id, "role_id": role_id}
+
+    async def list_all_permissions(self) -> list[dict[str, str]]:
```
yihua-zhuo, line 205: **INFO** (suggestion)

**Issue:** rule 15: list_all_permissions returns a hardcoded list from DEFAULT_PERMISSIONS (static data) rather than querying the PermissionModel table — if the DB has permissions not in the constant, they are invisible. The method name suggests it returns 'all' permissions.

**Suggestion:** Rename to 'list_default_permissions' or query PermissionModel for the complete list.

tests/integration/test_rbac_integration.py
yihua-zhuo, line 489: **WARNING** (suggestion)

**Issue:** rule 125/149: TestRoleManagementAPI class does not test the admin-permission gate (no test verifies that a non-admin user gets 403) — the implementation lacks the gate, so the test passes vacuously, but this means the security boundary is untested.

**Suggestion:** Add a test that calls a mgmt write endpoint without admin role and asserts 403; or document that admin gating is out of scope for this issue.

tests/integration/test_rbac_integration.py
```diff
@@ -14,9 +14,14 @@
...
+from sqlalchemy.ext.asyncio import AsyncSession
 
+from db.models.rbac import RoleModel
+from db.models.user import UserModel
 from pkg.errors.app_exceptions import NotFoundException, ValidationException
-from services.rbac_service import RBACService
+from services.rbac_service import DEFAULT_PERMISSIONS, RBACService
 
 
 async def _seed_rbac_data():
```
yihua-zhuo, line 27: **WARNING** (suggestion)

**Issue:** rule 146: _seed_rbac_data() opens a new asyncpg connection from a raw DSN string — if the DATABASE_URL is not set, the test fails with a RuntimeError, but this means every integration test in the file requires the env var even if the test does not use RBAC data. The seed function is also not idempotent if the migration is not applied first.

**Suggestion:** Check for migration existence before seeding, or wrap the seed in a try/except that skips RBAC tests gracefully.

tests/integration/test_rbac_integration.py
```diff
@@ -397,3 +402,172 @@ async def test_custom_role_invisible_to_other_tenant(self, db_schema, tenant_id,
...
+        resp = await api_client.get("/api/v1/rbac/mgmt/permissions")
+        assert resp.status_code == 200
+        perms = resp.json()["data"]
+        assert len(perms) == len(DEFAULT_PERMISSIONS)
+        for entry in perms:
+            assert set(entry.keys()) == {"resource", "action"}
+            assert entry["resource"]
+            assert entry["action"]
+
+    async def test_tenant_isolation_roles(
```
yihua-zhuo, line 525: **WARNING** (suggestion)

**Issue:** rule 144: test_tenant_isolation_roles only asserts that a role name is absent/in present, but does not verify the count or the role's tenant_id field in the response — a test that passes a role with the same name in both tenants would also pass.

**Suggestion:** Assert that the response items include the role's tenant_id field and that it matches the expected tenant.

tests/integration/test_rbac_integration.py
```diff
@@ -14,9 +14,14 @@
 
 import asyncpg
```
yihua-zhuo, line 15: **INFO** (suggestion)

**Issue:** rule 146: The file imports asyncpg at module level (line 15) and httpx at line 17 — module-level imports of optional drivers can cause ImportError in environments that do not have asyncpg installed, even if a particular test does not need it.

**Suggestion:** Move asyncpg import inside the _seed_rbac_data function to keep the module import-safe.

tests/unit/test_role_service.py
```diff
@@ -0,0 +1,313 @@
...
+
+        custom = RoleModel(
+            id=900,
+            tenant_id=1,
+            name="custom_tenant_role",
+            display_name="Custom Tenant Role",
+            description="",
+            is_system=False,
+            priority=0,
+            created_at=__import__("datetime").datetime.now(__import__("datetime").UTC),
```
yihua-zhuo, line 57: **WARNING** (suggestion)

**Issue:** Repeated `__import__("datetime").datetime.now(__import__("datetime").UTC)` (lines 57, 83, 151, 189, 221) is an unusual and hard-to-read pattern used to avoid adding an import line.

**Suggestion:** Add `from datetime import datetime, UTC` at the top of the module and use `datetime.now(UTC)`.

tests/unit/test_role_service.py
```diff
@@ -0,0 +1,313 @@
...
+@pytest.fixture
+def role_service(mock_db_session):
+    return RoleService(mock_db_session)
+
+
+class TestListRoles:
+    async def test_list_roles_returns_tenant_and_system_roles(self, role_service, state):
+        from db.models.rbac import RoleModel
+
+        custom = RoleModel(
```
yihua-zhuo, line 49: **INFO** (suggestion)

**Issue:** Constructing a full RoleModel and then manually copying its fields into a plain dict (lines 49-68, 74-84, 143-152, 180-189, 213-222) is redundant — the mock handler reads from `state.roles` as a dict, not from ORM objects.

**Suggestion:** Drop the RoleModel construction and seed `state.roles[custom_id] = {"id": ..., "tenant_id": 1, ...}` directly; add a small helper if the repetition bothers you.

tests/unit/test_role_service.py
```diff
@@ -0,0 +1,313 @@
...
+            "tenant_id": 1,
+            "name": "replaceable",
+            "display_name": "Replaceable",
+            "description": "",
+            "is_system": False,
+            "priority": 0,
+            "created_at": __import__("datetime").datetime.now(__import__("datetime").UTC),
+        }
+        state.role_permissions.append(
+            {"id": 1, "role_id": custom_id, "permission_id": 1}
```
yihua-zhuo, line 192: **INFO** (suggestion)

**Issue:** Hard-coded role_permission ids `1` and `2` in the test will collide with the mock's `role_permissions_next_id` (which starts at 1 and auto-increments), risking id reuse across tests run in the same session.

**Suggestion:** Use ids in a reserved range (e.g. 9000+) or call `state.role_permissions_next_id` to allocate the next id before appending.

tests/unit/test_role_service.py
```diff
@@ -0,0 +1,313 @@
...
+
+
+@pytest.fixture
+def state() -> RBACMockState:
+    return RBACMockState()
+
+
+@pytest.fixture
+def mock_db_session(state: RBACMockState):
+    from tests.unit.domain_handlers.rbac import get_handlers
```
yihua-zhuo, line 32: **INFO** (suggestion)

**Issue:** The `mock_db_session` fixture re-imports `get_handlers` inside the fixture body even though it is only used here.

**Suggestion:** Move `from tests.unit.domain_handlers.rbac import get_handlers` to module-level imports next to the other handler imports.

yihua-zhuo:

**Status:** `fail`
**Inline comments posted:** 29

**Summary:** Reviewed 7 file(s) between origin/master and origin/auto/issue-642.

`tests/integration/test_rbac_integration.py` L393: 
**CRITICAL** (critical)

**Issue:** rule 124: test_custom_role_invisible_to_other_tenant uses 'other_tenant = tenant_id + 99999' to simulate a second tenant, but there is no guarantee this tenant_id does not collide with a real seeded tenant — the test would silently pass or fail depending on seed data.

**Suggestion:** Use the tenant_id_2_web fixture or a properly isolated tenant_id from the conftest, not an arbitrary offset. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`src/api/routers/rbac.py` L61: 
**WARNING** (suggestion)

**Issue:** rule 4: rbac_router and role_management_router are defined in the same file with overlapping responsibility — the file serves two distinct API surfaces (full RBAC CRUD and the six-endpoint mgmt facade) without clear separation.

**Suggestion:** Split into rbac.py (existing CRUD) and role_management.py (new six endpoints) per single-responsibility principle. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`tests/integration/test_rbac_integration.py` L393: 
**WARNING** (suggestion)

**Issue:** rule 63: test_custom_role_invisible_to_other_tenant is not deterministic — the assertion 'assert total_same >= 5' depends on the seed data count being exactly 5, and the test does not assert that the custom role is absent from the second tenant's list (it only checks by name, not by id).

**Suggestion:** Assert that no row with tenant_id == tenant_id exists in the result of list_roles(tenant_id=other_tenant). — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`tests/integration/test_rbac_integration.py` L135: 
**INFO** (suggestion)

**Issue:** rule 143: seed_rbac_data is an async fixture but it calls _seed_rbac_data() which uses await asyncpg.connect — this is fine for async, but the fixture is 'autouse=True' and 'scope=function', meaning every test in this file pays the asyncpg connection cost even for tests that do not touch RBAC tables.

**Suggestion:** Make the fixture non-autouse or scope it to only the TestRoleManagementAPI class. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`tests/unit/domain_handlers/rbac.py` L144: 
**WARNING** (suggestion)

**Issue:** Rule 134: The role_permission DELETE handler filters only by role_id with no tenant_id check, so a test could delete permissions for roles owned by other tenants without any mock-level guard.

**Suggestion:** Extract tenant_id from the params and include it in the filter: `state.role_permissions = [rp for rp in state.role_permissions if not (rp["role_id"] == role_id and rp.get("tenant_id") == tid)]` — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`tests/unit/domain_handlers/rbac.py` L169: 
**WARNING** (suggestion)

**Issue:** Rule 134: tenant_id is parsed unconditionally at the top of the role handler; for UPDATE/DELETE the guards at lines 226 and 236 compare `state.roles[rid]["tenant_id"] == tenant_id`, which becomes `== None` if the query omitted the param, causing every legitimate call to silently no-op.

**Suggestion:** Parse tenant_id only inside the branches that need it, or raise/warn if a write-path query arrives without a tenant_id bind. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`tests/unit/domain_handlers/rbac.py` L43: 
**INFO** (suggestion)

**Issue:** Local import of `RoleModel` (and `PermissionModel` on line 59) inside factory functions — works, but leaks import-time coupling into every call and makes the module harder to lint or type-check.

**Suggestion:** Move `from db.models.rbac import RoleModel, PermissionModel` to the top of the module. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`tests/unit/domain_handlers/rbac.py` L85: 
**INFO** (suggestion)

**Issue:** `import re` inside `_parse_int_param` is executed on every call; same for `import re as _re` on line 263 and `import re` on line 297.

**Suggestion:** Hoist `import re` to the module-level imports. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`tests/unit/domain_handlers/rbac.py` L316: 
**INFO** (suggestion)

**Issue:** user_roles INSERT uses `or 0` fallbacks for tenant_id and granted_by, silently turning a missing bind into a valid system-tenant row and masking bugs where the service forgot to pass those values.

**Suggestion:** Pass through the raw value (or raise) instead of substituting 0: `_extract_int(...) if "tenant_id" in (k for k in params) else None`. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`tests/unit/domain_handlers/users.py` L63: 
**WARNING** (suggestion)

**Issue:** Rule 19/135: Magic numbers `7` and `3` hard-coded for tenant_id==1 vs. others in the COUNT fallback — this ties the shared mock to a specific test's expectations and is a 'utility.py' style dumping ground for test fixtures.

**Suggestion:** Remove the fallback; have tests that need a non-empty count seed state.users directly. If a default is unavoidable, make it one shared constant. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`tests/unit/domain_handlers/users.py` L71: 
**WARNING** (suggestion)

**Issue:** Rule 135: Hard-coded fixture dicts (existing/alice/john_doe/jane_doe) and (lines 111-114, 128-132) are embedded in the shared handler, mixing test-specific seed data with reusable mock logic.

**Suggestion:** Move per-username/id fixtures into the test files that need them, or expose a `seed_user(state, ...)` helper. Keep the handler generic. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`tests/unit/domain_handlers/users.py` L79: 
**WARNING** (suggestion)

**Issue:** `dt.utcnow()` is deprecated in Python 3.12+ (also lines 91, 119, 127).

**Suggestion:** Use `from datetime import UTC, datetime as dt` and `dt.now(UTC)` everywhere. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`tests/unit/domain_handlers/users.py` L50: 
**INFO** (suggestion)

**Issue:** Rule 132/133: `("set " in sql_text or sql_text.startswith("update")) and "where id" in sql_text` is fragile substring matching — a column named anything_set or a table aliased near 'id' will false-match and route to the user handler.

**Suggestion:** Narrow with a regex or check for the specific table token (e.g. `re.search(r'\bUPDATE\s+users\b', sql_text, re.I)`). — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`tests/unit/domain_handlers/users.py` L125: 
**INFO** (suggestion)

**Issue:** Unconditional fall-through returns every user in state when no pattern matches — this swallows unknown query shapes and can mask service bugs by always returning a non-empty result.

**Suggestion:** Return `MockResult([])` on no-match and let the service raise NotFoundException, or log/raise a clear 'unhandled query' error. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_

PR #789, opened 2026-06-04
auto/issue-642 → master

yihua-zhuo:
Auto-created by CI from branch `auto/issue-642`.


src/api/routers/role_management.py
```diff
@@ -0,0 +1,142 @@
...
+    """At least one permission is required to avoid creating a zero-permission role."""
+
+    permissions: list[str] = Field(..., min_length=1)
+
+
+class AssignRoleRequest(BaseModel):
+    role_id: int = Field(..., description="Role to assign to the user")
+
+
+@role_management_router.get("/roles")
```
yihua-zhuo, line 48: **CRITICAL** (critical)

**Issue:** Pagination is done in Python after fetching the full list from the service — this loads all roles into memory and then slices, which does not scale and is an O(N) memory hit per request (Rules 31, 34).

**Suggestion:** Push page/page_size into the service call: `await svc.list_roles(tenant_id=ctx.tenant_id, page=page, page_size=page_size)` and return `(items, total)` from the service like the legacy RBAC router does.

src/api/routers/role_management.py
```diff
@@ -0,0 +1,142 @@
...
+async def mgmt_list_roles(
+    page: int = Query(1, ge=1),
+    page_size: int = Query(50, ge=1, le=200),
+    ctx: AuthContext = Depends(require_auth),
+    session: AsyncSession = Depends(get_db),
+):
+    svc = RoleService(session)
```
yihua-zhuo, line 55: **CRITICAL** (critical)

**Issue:** `RoleService.list_roles(tenant_id=...)` is called without page/page_size, then the full result set is serialized with `.to_dict()` on every item before pagination — for tenants with thousands of roles this is both a memory and CPU problem (Rule 90).

**Suggestion:** Have the service accept page/page_size and return a tuple so the router only serializes the page slice.

tests/integration/test_rbac_integration.py
```diff
@@ -571,3 +637,68 @@ async def test_create_role_duplicate_name_returns_409(
 
         second = await api_client.post("/api/v1/rbac/mgmt/roles", json=body)
         assert second.status_code == 409
+
+    async def test_create_role_denies_non_admin(
```
yihua-zhuo, line 641: **CRITICAL** (critical)

**Issue:** `test_create_role_denies_non_admin` mutates `_app.dependency_overrides` directly and uses its own ASGITransport client — this duplicates the `client` / `api_client` fixture logic and creates a second dependency override that races with the `client` fixture's override if both are used in the same test session (Rule 112, Rule 124).

**Suggestion:** Extract the ASGI client + dependency override into a shared fixture (e.g., `non_admin_client`) parameterized by auth headers.

src/api/routers/rbac.py
```diff
@@ -11,10 +11,8 @@
 from db.connection import get_db
 from internal.middleware.fastapi_auth import AuthContext, require_auth
 from services.rbac_service import RBACService
-from services.role_service import RoleService
 
 rbac_router = APIRouter(prefix="/api/v1/rbac", tags=["rbac"])
-role_management_router = APIRouter(prefix="/api/v1/rbac/mgmt", tags=["rbac-roles"])
 
 
 def _paginated(items, total, page, page_size):
```
yihua-zhuo, line 18: **INFO** (suggestion)

**Issue:** Helper `_paginated` is a clean DRY utility; consistent envelope shape across endpoints.

src/api/routers/role_management.py
```diff
@@ -0,0 +1,142 @@
...
+    roles = await svc.list_roles(tenant_id=ctx.tenant_id)
```
yihua-zhuo, line 56: **INFO** (suggestion)

**Issue:** Mixes service classes (RoleService vs RBACService in the legacy router) for overlapping functionality — potential for divergent behavior (e.g. different tenant filtering semantics) between the two routers (Rule 112).

**Suggestion:** Consolidate into a single RBAC service or document the boundary explicitly in the module docstring.

src/api/routers/role_management.py
```diff
@@ -0,0 +1,142 @@
...
+    return {"success": True, "data": result, "message": msg}
+
+
+@role_management_router.get("/permissions")
+async def mgmt_list_permissions(
+    ctx: AuthContext = Depends(require_auth),
+    session: AsyncSession = Depends(get_db),
+):
+    svc = RoleService(session)
+    perms = await svc.list_all_permissions()
```
yihua-zhuo, line 141: **INFO** (suggestion)

**Issue:** `mgmt_list_permissions` returns a raw list without pagination envelope — acceptable if the permission set is small and static, but lacks consistency with the paginated `/roles` endpoint (Rule 91).

**Suggestion:** Wrap in a consistent envelope: `{"data": {"items": perms, "total": len(perms)}}` for API consistency.

src/api/routers/role_management.py
```diff
@@ -0,0 +1,142 @@
...
+from db.connection import get_db
+from dependencies.rbac import require_permission
+from internal.middleware.fastapi_auth import AuthContext, require_auth
+from services.role_service import RoleService
+
+role_management_router = APIRouter(prefix="/api/v1/rbac/mgmt", tags=["rbac-roles"])
+
+
+class CreateRoleRequest(BaseModel):
+    name: str = Field(..., min_length=1, max_length=50)
```
yihua-zhuo, line 34: **INFO** (suggestion)

**Issue:** `CreateRoleRequest.permissions: list[str] = Field(..., min_length=0)` allows zero permissions at creation time — the comment on line 39 explains this is intentional for `UpdatePermissionsRequest` but the create schema allows the same loophole, contradicting the stated invariant.

**Suggestion:** Set `min_length=1` on CreateRoleRequest.permissions to match the invariant that every role has at least one permission.

src/api/routers/role_management.py
```diff
@@ -0,0 +1,142 @@
...
+from fastapi import APIRouter, Depends, Path, Query
+from pydantic import BaseModel, Field
+from sqlalchemy.ext.asyncio import AsyncSession
+
+from db.connection import get_db
+from dependencies.rbac import require_permission
+from internal.middleware.fastapi_auth import AuthContext, require_auth
+from services.role_service import RoleService
+
+role_management_router = APIRouter(prefix="/api/v1/rbac/mgmt", tags=["rbac-roles"])
```
yihua-zhuo, line 30: **INFO** (suggestion)

**Issue:** Path prefix `/api/v1/rbac/mgmt` is a sub-prefix of the legacy router at `/api/v1/rbac` — the comment block explains the collision avoidance, but this creates two divergent APIs for the same domain (Rule 112, Rule 120).

**Suggestion:** Deprecate the legacy RBAC router or merge the management endpoints into it; avoid maintaining two parallel APIs.

tests/integration/test_rbac_integration.py
```diff
@@ -32,6 +31,10 @@ async def _seed_rbac_data():
     This function inserts the seed data into already-existing tables.
 
     Uses asyncpg (async) so it is safe to call inside async test fixtures.
+
+    asyncpg is imported lazily here (not at module top) so the rest of the
+    file can be imported in environments where asyncpg is not installed
+    (e.g. type-checker only runs).
     """
     db_url = os.environ.get("DATABASE_URL")
```
yihua-zhuo, line 39: **WARNING** (suggestion)

**Issue:** `_seed_rbac_data` opens a separate `asyncpg` connection and inserts seed data via raw SQL with `ON CONFLICT DO NOTHING` — this bypasses the ORM session and the `db_schema` truncation may not see the data if the connection pool has a different snapshot. Also uses f-string-less but still hardcoded SQL (Rule 132).

**Suggestion:** Use the `async_session` fixture to insert seed data via ORM or bound parameters; avoid opening a separate asyncpg connection.

tests/integration/test_rbac_integration.py
```diff
@@ -434,6 +456,8 @@ async def test_create_role_then_list_includes_it(
         body = resp.json()
         assert body["success"] is True
         assert body["data"]["name"] == "custom_support"
+        assert body["data"]["tenant_id"] is not None
+        assert "admin" not in body["data"] or body["data"]["is_system"] is False
```
yihua-zhuo, line 460: **WARNING** (suggestion)

**Issue:** Assertion `assert 'admin' not in body['data'] or body['data']['is_system'] is False` is a tautology — `'admin' not in body['data']` is almost always True because `body['data']` is a dict containing the created role's fields, not a list. This assertion never fails and provides no coverage (Rule 69).

**Suggestion:** Remove this dead assertion; the test already verifies the role was created and is non-system implicitly via the `custom_support` name check.

tests/integration/test_recommendation_integration.py
```diff
@@ -36,9 +36,10 @@ async def _seed_user(async_session, tenant_id: int = 1) -> int:
 
 async def _seed_customer(async_session, tenant_id: int = 1, **overrides):
     """Create a customer and return the CustomerModel."""
+    from db.repositories.customer import CustomerRepository
```
yihua-zhuo, line 39: **INFO** (suggestion)

**Issue:** Imports for CustomerRepository and CustomerService are inside the _seed_customer function body rather than at module top, which is harder to audit and breaks the convention of module-level imports.

**Suggestion:** Move the imports to the top of the file unless there is a known circular-import reason to keep them lazy.

yihua-zhuo:

**Status:** `fail`
**Inline comments posted:** 11

**Summary:** Reviewed 6 file(s) between origin/master and origin/auto/issue-642.

- `src/api/routers/rbac.py`: Routers serialize correctly via .to_dict() and services return domain objects — layering is sound.
- `tests/integration/conftest.py`: The conftest is large shared test infrastructure — modifying it for domain changes would violate Rule 149. The current state is appropriate as shared scaffolding.
- `tests/integration/test_rbac_integration.py`: Integration tests correctly use real DB via async_session and db_schema fixtures — no mocking violations (Rule 124 satisfied).
- `tests/integration/test_rbac_integration.py`: Good test coverage of RBAC operations including tenant isolation, duplicate handling, and permission validation — tests are deterministic and use real DB (Rule 63, Rule 65).

`tests/integration/conftest.py` L130: 
**CRITICAL** (critical)

**Issue:** `_test_get_db_session` is an async context manager that commits on normal exit and rolls back on exception — this conflicts with the router-layer transaction boundary (Rule 121) if any service or router under test uses this session directly. The `client` fixture overrides `get_db` to share `async_session`, so the override is used there, but this standalone session manager is a footgun.

**Suggestion:** Add a docstring warning that `_test_get_db_session` should not be used in tests that also use the `client` fixture; or remove it if unused. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`tests/integration/conftest.py` L197: 
**CRITICAL** (critical)

**Issue:** `Base.metadata.create_all()` is called in the test session-scope fixture — this is acceptable for test setup (Rule 130 allows it), but it bypasses Alembic migrations, meaning integration tests never exercise the actual migration chain. A migration that works in production may fail in tests.

**Suggestion:** Run `alembic upgrade head` against the test DB instead of `create_all()` for at least one test suite to validate migration correctness. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`src/api/routers/rbac.py` L131: 
**WARNING** (suggestion)

**Issue:** `list_permissions` accepts no `tenant_id` — consistent with the permissions table being global, but the endpoint advertises no tenant scoping. A future tenant-scoped permissions refactor will need a breaking change here (Rule 126).

**Suggestion:** Add a docstring noting permissions are intentionally global; revisit when permissions become tenant-owned. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`src/api/routers/rbac.py` L69: 
**INFO** (suggestion)

**Issue:** `data.display_name or data.name` silently coerces None to a default — acceptable for a required-but-optional field, but a Pydantic default would be clearer.

**Suggestion:** Set `display_name: str | None = None` on RoleCreate and default to `data.name` in the service instead. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`src/api/routers/rbac.py` L256: 
**WARNING** (suggestion)

**Issue:** `list_users_with_role` serializes via `u.to_dict()` — the UserModel.to_dict() allow-list correctly excludes `password_hash` (Rule 137), but no test or docstring confirms this contract here.

**Suggestion:** Add a comment asserting that the allow-list in UserModel.to_dict() intentionally omits credential fields. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`tests/integration/conftest.py` L156: 
**WARNING** (suggestion)

**Issue:** `_install_test_db_session` monkey-patches `get_db_session` on every service module by iterating `pkgutil.iter_modules` — this is fragile because services that import `get_db_session` lazily (inside functions) will still get the original reference. The `conn_mod.engine` and `conn_mod.async_session_maker` reassignment is also global state mutation (Rule 28).

**Suggestion:** Use FastAPI's `dependency_overrides` mechanism (already done for `get_db` in the `client` fixture) instead of monkey-patching service modules. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`tests/integration/conftest.py` L81: 
**WARNING** (suggestion)

**Issue:** Hardcoded `TEST_JWT_SECRET = "integration-test-jwt-secret-key-32"` is acceptable for tests, but also sets `os.environ["JWT_SECRET"]` and `os.environ["JWT_SECRET_KEY"]` globally — any other test process or subprocess inherits this (Rule 28).

**Suggestion:** Use `monkeypatch.setenv` in fixtures instead of mutating `os.environ` at module level. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`tests/integration/conftest.py` L243: 
**INFO** (suggestion)

**Issue:** `tenant_id` uses `random.randint(10_000_000, 99_999_999)` — collision probability is very low for typical test suites, but parallel CI runs with thousands of tests could theoretically collide. Not critical given the truncation strategy.

**Suggestion:** No change needed; the collision space is large enough for practical use. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`tests/integration/conftest.py` L298: 
**INFO** (suggestion)

**Issue:** The `_cleanup_after_module` autouse fixture is described as 'mandatory' in the comment but is not enforced — tests that don't use `db_schema` will still get the cleanup, which is fine, but the comment language could mislead contributors into thinking it's a strict contract.

**Suggestion:** Soften the comment from 'mandatory' to 'recommended as defense-in-depth'. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`tests/integration/conftest.py` L327: 
**WARNING** (suggestion)

**Issue:** `client` fixture's docstring says `async_session` is 'intentionally consumed here (not used directly)' to guarantee write ordering — this is an implicit ordering contract that's easy to break. The dependency-injection chain alone doesn't guarantee flush ordering between fixtures (Rule 112).

**Suggestion:** Use an explicit `await async_session.flush()` or a `finalizer` to make the ordering contract explicit rather than relying on dependency-consumption side effects. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`tests/integration/test_rbac_integration.py` L76: 
**WARNING** (suggestion)

**Issue:** System roles are inserted with `tenant_id=0` — this relies on a magic value (0) to mean 'global/system' (Rule 19). The convention should be documented or use a nullable/sentinel value instead.

**Suggestion:** Document the `tenant_id=0` convention in the RBAC model docstring or use a proper `is_system` boolean with a separate global-roles table. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`tests/integration/test_rbac_integration.py` L174: 
**INFO** (suggestion)

**Issue:** Tests query `list_roles(tenant_id=0)` to find system roles — this hardcodes the tenant_id=0 convention in test code, coupling tests to the same magic value (Rule 19).

**Suggestion:** Add a `get_system_roles()` service method or fixture that encapsulates the tenant_id=0 lookup. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`tests/integration/test_rbac_integration.py` L400: 
**INFO** (suggestion)

**Issue:** Cross-tenant test uses `other_tenant = tenant_id + 99999` — this offset is fragile if `tenant_id` is near the upper bound of the int range. Also, the 'other tenant' is never seeded as a TenantModel, so if the service ever starts requiring a valid tenant FK, the test breaks (Rule 138).

**Suggestion:** Use the `tenant_id_2` fixture and `_seed_tenant_2` to create a proper second tenant record. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`tests/integration/test_rbac_integration.py` L572: 
**INFO** (suggestion)

**Issue:** `test_list_permissions_returns_all_system_pairs` asserts `len(perms) == len(DEFAULT_PERMISSIONS)` — this ties the test to the exact count of permissions in the service constant, which is correct but brittle if the constant changes (Rule 63).

**Suggestion:** Assert specific permission names are present (e.g., 'customer:create') rather than exact count, or add a comment explaining the count is intentional. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`tests/integration/test_recommendation_integration.py` L24: 
**WARNING** (suggestion)

**Issue:** Cross-domain seed helpers (_seed_user, _seed_customer, _seed_opportunity) are defined locally in this test file instead of in tests/integration/domain_fixtures/, violating the domain-owned fixtures convention (rule 125).

**Suggestion:** Move the seed helpers to domain-owned fixture modules (e.g. tests/integration/domain_fixtures/customers.py, users.py, opportunities.py) and import them, or reuse existing cross-service seed helpers from the recommended location. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`tests/integration/test_recommendation_integration.py` L364: 
**WARNING** (suggestion)

**Issue:** test_invalid_risk_level_string catches sa_exc.DBAPIError, which is the broadest base class and will match any database failure — a false positive risk if the schema or connection changes.

**Suggestion:** Use sa_exc.IntegrityError or the psycopg2 CheckViolation error class to specifically assert the CHECK constraint violation. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`tests/integration/test_recommendation_integration.py` L133: 
**INFO** (suggestion)

**Issue:** test_recommendation_tenant_isolation and test_risk_signal_tenant_isolation (line 202) are near-identical in structure, duplicating the same seed-then-cross-check pattern.

**Suggestion:** Parameterize the two isolation tests with pytest.mark.parametrize over the model class and tenant-filter assertion, or extract the shared setup/teardown into a helper. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`tests/unit/test_role_service.py` L236: 
**WARNING** (suggestion)

**Issue:** The user-seeding dict literal is copy-pasted three times (lines 236-248, 255-267, 280-292) with identical structure and only id/username/email varying, violating the no-duplication rule (rule 110).

**Suggestion:** Extract a helper such as _seed_user(state, user_id, tenant_id, username, email) in this file or import one from tests/unit/domain_handlers/users.py. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`tests/unit/test_role_service.py` L271: 
**WARNING** (suggestion)

**Issue:** assert first.get("already_assigned") is not True is a weak assertion — a missing key returns None, and falsy values like 0 or empty string would also pass, hiding real bugs.

**Suggestion:** Use assert "already_assigned" not in first or assert first.get("already_assigned") is False to make the negative case unambiguous. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`tests/unit/test_role_service.py` L236: 
**INFO** (suggestion)

**Issue:** Tests bypass the loaded make_user_handler and mutate state.users dicts directly. While workable, this partially circumvents the composable mock-state pattern (rule 135) and couples tests to internal state shape.

**Suggestion:** Use a seed helper from tests/unit/domain_handlers/users.py (e.g. add_user) so the handler mediates state creation, keeping tests independent of dict key names. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_

Issue #77, opened 2026-05-02
[Enhancement] Smart Lead Routing — Automated territory management


yihua-zhuo:
## Problem

当前线索（leads）分配靠销售手动抢单或管理员分配，无法根据地域/行业/时间段自动分配。导致分配不均、响应慢、线索流失。

## Goals

- 基于规则 + ML 的自动线索分配
- 地域 / 行业 / 产品线 / 时间段 多维度分配策略
- 分配时考虑销售负载（均衡分配）
- 线索回收机制（24h 不响应自动回收）

## Dependencies

- `leads` 表（已有）
- `users` 表（已有，关联 role）
- RBAC service

### 分配策略配置
```
routing_rules: id, tenant_id, name, conditions_json, assignee_type, assignee_id, priority
```
- conditions: {region: ["APAC", "EMEA"], industry: ["SaaS"], employee_count: ">50"}
- assignee_type: "user" | "team" | "round_robin"

### 分配引擎
- `src/services/lead_routing_service.py` — 线索分配服务
- `src/api/routers/lead_routing.py` — REST API
- 策略优先级：explicit rule → ML predicted → round_robin
- 负载均衡：每次分配选择当前负载最低的销售

### 线索回收
- 定时任务：每 30 分钟检查未分配线索
- 超过 24h 不响应 → 重新入池（标记为 recycled）
- 最多回收 3 次，第 4 次标记为 disqualified

### Frontend
- 分配规则配置页面（条件 + 动作可视化）
- 线索池列表（未分配 / 已分配 / 回收）
- 销售负载仪表板

### Lead Routing Configuration (`/settings/routing`)
- [ ] Rule list: name, conditions summary, assignee, priority
- [ ] Create rule: condition builder (field + operator + value)
- [ ] Supported fields: region, industry, employee_count, source, created_date
- [ ] Test rule with sample lead
- [ ] Enable / disable rule
- [ ] Rule priority ordering (drag to reorder)

### Lead Distribution (`/sales/leads`)
- [ ] Unassigned leads queue (sorted by creation time)
- [ ] "Assign" button → show matched rules + assignee preview
- [ ] Manual override: reassign to different user
- [ ] Lead detail: assigned_to, assigned_at, SLA countdown

### Auto-Assignment
- [ ] Real-time assignment on lead creation
- [ ] Assignment notification to assigned sales rep
- [ ] Round-robin distribution across team
- [ ] Load balancing: skip reps with >20 active leads

[...]

PR #804, opened 2026-06-06
Auto: implement issue #586

### LINKED ISSUE #586: Add lead tier filter and auto-rank to customer list endpoints + engagement webhook
Subtask of #49
Depends on #585

Add optional `lead_tier: str` query param to `GET /customers/` that filters by score tier (hot/warm/cold). Add `order_by_score: bool` to auto-rank. Add an engagement event webhook that triggers score recalculation on new email open / website visit events (calls ScoreService.calculate_score). Touch src/api/routers/customers.py, src/services/customer_service.py, and src/api/routers/events.py (new webhook endpoint), with corresponding tests.

Dev-Plan: docs/dev-plan/10-customers/0586-add-lead-tier-filter-and-auto-rank-to-customer-list-endpoint.md

yihua-zhuo:
Closes #586

Plan: `.plans/issue-586.md`

1. **Plan** — Claude wrote a detailed implementation plan to `.plans/issue-586.md`.

Please review before merging.

## Release Notes

* **New Features**
  * New webhook endpoint to record customer engagement events (email opens and website visits) for tracking interaction history
  * Added lead tier filtering to the customer list endpoint to segment customers by tier level (hot, warm, cold)
  * Added engagement score-based sorting to the customer list endpoint for automatic customer ranking by activity

### DIFF
--- a/.plans/issue-586.md
+++ b/.plans/issue-586.md
@@ -1,0 +1,79 @@
+Now I have all the context needed to write the plan. Here's the implementation plan:
+
+# Implementation Plan — Issue #586
+
+## Goal
+Add an optional `lead_tier` filter and `order_by_score` auto-rank to `GET /customers/`, plus a `POST /events/engagement` webhook that records engagement events and triggers score recalculation via `ScoreService.calculate_score`, closing the lead-scoring feedback loop.
+
+## Source Contract
+Dev-plan target: `/home/runner/work/agent-job/agent-job/docs/dev-plan/10-customers/0586-add-lead-tier-filter-and-auto-rank-to-customer-list-endpoint.md`
+Template depth: `deep`
+Reading order followed:
+1. `/home/runner/work/agent-job/agent-job/docs/dev-plan/README.md`
+2. `/home/runner/work/agent-job/agent-job/docs/dev-plan/_template-deep.md`
+3. `/home/runner/work/agent-job/agent-job/docs/dev-plan/10-customers/0586-add-lead-tier-filter-and-auto-rank-to-customer-list-endpoint.md`
+
+## Affected Files
+- `src/api/routers/customers.py` — add `lead_tier` and `order_by_score` query params to `list_customers`; thread them to `CustomerService.list_customers`
+- `src/services/customer_service.py` — extend `list_customers` signature with `lead_tier` and `order_by_score`; validate `lead_tier` against allowed set; pass through to `CustomerRepository.list_customers`
+- `src/db/repositories/customer.py` — extend `list_customers` signature with `lead_tier` and `order_by_score`; add `CustomerModel.tier` equality filter and conditional `ORDER BY score DESC` (with `COALESCE(score, 0)` to handle NULL)
+- `src/db/models/engagement.py` — new file; define `EngagementEventModel` (id, tenant_id, customer_id, event_type, event_metadata JSONB, created_at)
+- `src/services/event_service.py` — new file; define `EventService` with `record_engagement_event(tenant_id, customer_id, event_type, metadata) -> EngagementEventModel`
+- `src/api/routers/events.py` — new file; `APIRouter(prefix="/api/v1/events", tags=["events"])` with `POST /engagement` endpoint that calls `EventService.record_engagement_event` then `ScoreService.calculate_score`, returns `{"success": True, "data": {"customer_id": N, "score": M, "tier": "A"}}`
+- `alembic/versions/<id>_add_engagement_events.py` — new migration; `CREATE TABLE engagement_events` with `tenant_id` index and `event_metadata` JSONB; manually verify `sa.JSONB()` and `sa.DateTime(timezone=True)`
+- `tests/unit/test_customer_service.py` — add 4 new cases under `TestListCustomers`: hot filter, warm filter, cold filter, `order_by_score` parameter pass-through; plus 1 validation-error case for invalid tier value
+- `tests/unit/test_events_webhook.py` — new file; unit tests for the events router: valid engagement POST returns 200 with score; invalid event_type returns 422; mock `ScoreService.calculate_score` assertion
+- `tests/integration/test_engagement_webhook_integration.py` — new file; integration tests: POST engagement event → customer score/tier updated in DB; multi-tenant isolation; `order_by_score` returns highest-scoring customers first; `lead_tier` filter returns only matching tier
+
+## Implementation Steps
+
+1. **Extend `CustomerRepository.list_customers`** in `src/db/repositories/customer.py` (L38-64): add `lead_tier: str | None = None` and `order_by_score: bool = False` parameters; in the conditions list, add `CustomerModel.tier == lead_tier` when `lead_tier` is not None; for ordering, use `func.coalesce(CustomerModel.score, 0).desc()` when `order_by_score` is True, otherwise retain the existing `CustomerModel.created_at.desc()`.
+
+2. **Extend `CustomerService.list_customers`** in `src/services/customer_service.py` (L48-63): add the same two parameters; validate `lead_tier` against the set `{"hot", "warm", "cold"}` and raise `ValidationException` for unknown values; forward both parameters to `self.repository.list_customers`.
+
+3. **Update `GET /customers/` in `src/api/routers/customers.py`** (L213-244): add `lead_tier: str | None = Query(None, description="Filter by lead tier: hot, warm, cold")` and `order_by_score: bool = Query(False, description="Auto-rank by score descending")` parameters; pass both to `service.list_customers`. No changes to the enrichment enrichment loop or response envelope.
+
+4. **Create `EngagementEventModel`** in `src/db/models/engagement.py`: columns `id` (Integer, PK), `tenant_id` (Integer, nullable=False, index=True), `customer_id` (Integer, nullable=False, index=True), `event_type` (String(50), nullable=False), `event_metadata` (JSON, nullable=True, default=dict), `created_at` (DateTime(timezone=True), server_default=func.now()). The model will be auto-discovered by `src/db/models/__init__.py` via `pkgutil.iter_modules` — no manual import needed there.
+
+5. **Generate and fix Alembic migration**: run `alembic revision --autogenerate -m "add engagement_events table"` against the `alembic_dev` DB; manually correct `sa.JSON()` → `sa.JSONB()` on the `event_metadata` column; verify `sa.DateTime(timezone=True)` on `created_at`; run `alembic upgrade head && alembic downgrade -1 && alembic upgrade head` three-cycle; run a second `autogenerate` to confirm empty drift, then delete the drift-check file.
+
+6. **Create `EventService`** in `src/services/event_service.py`: `__init__(self, session: AsyncSession)` with no default; `record_engagement_event(tenant_id, customer_id, event_type, metadata=None) -> EngagementEventModel` that constructs an `EngagementEventModel` instance, adds it, commits, refreshes, and returns. Validate `event_type` is in `{"email_open", "website_visit"}` and raise `ValidationException` otherwise.
+
+7. **Create `POST /engagement` router** in `src/api/routers/events.py`: define `router = APIRouter(prefix="/api/v1/events", tags=["events"])` (auto-discovered by `iter_routers` — no `main.py` change needed). Define a Pydantic `EngagementEventRequest(BaseModel)` with `customer_id: int` (gt=0), `event_type: str` (pattern=`^(email_open|website_visit)$`), `metadata: dict | None = None`. The endpoint handler instantiates `EventService(session)` and `ScoreService(session)`, calls `record_engagement_event` then `calculate_score`, and returns `{"success": True, "data": {"customer_id": ..., "score": result.score, "tier": result.tier_label}}`. Auth via `Depends(require_auth)`.
+
+8. **Add unit tests for `list_customers` parameter pass-through** in `tests/unit/test_customer_service.py`: a new `TestListCustomers` class with 4 cases — `test_list_customers_passes_lead_tier_hot` verifies `repo.list_customers` receives `lead_tier="hot"`; `test_list_customers_passes_lead_tier_warm`; `test_list_customers_passes_lead_tier_cold`; `test_list_customers_passes_order_by_score_true` verifies `order_by_score=True` is forwarded. Plus `test_list_customers_invalid_tier_raises_validation` which calls `service.list_customers(lead_tier="invalid")` and asserts `ValidationException`.
+
+9. **Create unit tests for the engagement webhook** in `tests/unit/test_events_webhook.py`: use a mock `AsyncSession` with patched `EventService.record_engagement_event` and `ScoreService.calculate_score`; test that a valid POST returns the expected envelope; test that an invalid `event_type` returns 422 (Pydantic validation); assert `ScoreService.calculate_score` is awaited exactly once after the event is recorded.
+
+10. **Create integration tests** in `tests/integration/test_engagement_webhook_integration.py`: seed a customer in a tenant, POST to `/api/v1/events/engagement` with `event_type="email_open"`, then query the customer row and assert `score` and `tier` are updated. Add a cross-tenant test: POST with `tenant_id=A` customer, query from `tenant_id=B` → should not see the event. Add a `lead_tier` filter test: seed two customers with different tier values, GET `/api/v1/customers/?lead_tier=hot` returns only the hot one. Add an `order_by_score` test: seed three customers with scores 10/50/90, GET `/api/v1/customers/?order_by_score=true` returns them in descending order.
+
+11. **Run the full verification pipeline**: `ruff check src/ && ruff format --check src/` must exit 0; `PYTHONPATH=src pytest tests/unit/ -v` must pass; `PYTHONPATH=src pytest tests/integration/test_engagement_webhook_integration.py -v` must pass; `alembic upgrade head && alembic downgrade -1 && alembic upgrade head` must complete without error.
+
+## Test Plan
+- Unit tests in `tests/unit/`: 
+  - `test_customer_service.py` — add `TestListCustomers` class (4 pass-through cases + 1 validation case)
+  - `test_events_webhook.py` — new file; mock-based tests for the `/engagement` endpoint covering valid POST, invalid event_type, and `ScoreService` invocation assertion
+- Integration tests in `tests/integration/`:
+  - `test_engagement_webhook_integration.py` — new file; full-DB tests for webhook trigger → score update, cross-tenant isolation, `lead_tier` filter accuracy, and `order_by_score` sort order
+- Dev-plan verification:
+  - §6 acceptance #1: `ruff check src/api/routers/customers.py src/api/routers/events.py src/services/customer_service.py src/services/event_service.py` → 0 errors
+  - §6 acceptance #2: `PYTHONPATH=src pytest tests/unit/test_customer_service.py -v` → 4+ new cases passed
+  - §6 acceptance #3: `PYTHONPATH=src pytest tests/unit/test_events_webhook.py -v` → all passed
+  - §6 acceptance #4: `PYTHONPATH=src pytest tests/integration/test_engagement_webhook_integration.py -v` → all passed
+  - §6 acceptance #5: `alembic upgrade head && alembic downgrade -1 && alembic upgrade head` → three exit 0 (covers Step 5 migration)
+  - §6 E2E webhook: `curl -X POST /api/v1/events/engagement` with valid body → returns `{"success": true, "data": {"customer_id": N, "score": M, "tier": "..."}}`
+  - §6 E2E list: `curl /api/v1/customers/?lead_tier=hot&order_by_score=true` → returns filtered, score-sorted list
+
+## Acceptance Criteria
+- `GET /api/v1/customers/?lead_tier=hot` (and `warm`, `cold`) returns only customers whose `tier` column matches; invalid tier values produce a 422 with `ValidationException`
+- `GET /api/v1/customers/?order_by_score=true` returns customers sorted by `score DESC` (with NULLs treated as 0 via `COALESCE`)
+- Both query parameters are optional and independent — omitting both preserves the existing `ORDER BY created_at DESC` behavior with no regression
+- `POST /api/v1/events/engagement` with a valid body (authenticated) records an `EngagementEventModel` row and invokes `ScoreService.calculate_score` exactly once, returning the recalculated `score` and `tier` in the response
+- The engagement event row is persisted with the correct `tenant_id`, `customer_id`, `event_type`, and `event_metadata`
+- Cross-tenant isolation holds: a webhook POST for `tenant_id=A` is not visible or actionable from `tenant_id=B`
+- All unit and integration tests pass; `ruff check` and `ruff format --check` are clean; migration applies and reverts cleanly in a three-cycle
+
+## Risks / Open Questions
+- **Tier value mapping**: The dev-plan body says `lead_tier` accepts `hot/warm/cold`, but the existing `CustomerModel.tier` column stores the `ScoreTier` enum values (`A`/`B`/`C`/`D`). The plan adds a `{"hot", "warm", "cold"}` validation set in the service layer, but the actual SQL filter operates on the stored letter tier. If callers expect hot/warm/cold to map to A/B/C, the mapping must be defined at the service or router boundary — not in the SQL query. This decision should be confirmed with the issue owner before implementation, as it may require updating the filter to translate input values to the stored enum.
+- **Webhook auth model**: The dev-plan Step 3 uses `Depends(require_auth)`, which expects a user JWT. External systems (email trackers, website analytics) typically authenticate via a shared secret or API key, not user credentials. The plan follows the dev-plan's auth choice, but a production webhook would more likely use a header-based secret check. This should be confirmed with the issue owner if the webhook is intended for external system integration.
+- **ScoreService persistence**: `ScoreService.calculate_score` (in `src/services/score_service.py`) is currently read-only — it computes a `ScoreResult` from the persisted `score_factors` JSON column but does not write the computed `score`/`tier` back to `CustomerModel`. The webhook calls `calculate_score` and then must persist the result. The plan assumes the webhook handler (or a follow-up) updates `customer.score` and `customer.tier` after calling `calculate_score`; if `ScoreService` is extended to persist as part of #585, this becomes simpler. This should be verified against the #585 implementation before starting work.
--- a/alembic/versions/add_engagement_events_table.py
+++ b/alembic/versions/add_engagement_events_table.py
@@ -1,0 +1,59 @@
+"""add_engagement_events_table
+
+Revision ID: add_engagement_events
+Revises: e8ad92ca62b3
+Create Date: 2026-06-06 10:00:00.000000
+
+Create the engagement_events table used by the POST /events/engagement webhook
+to record email_open and website_visit events. The webhook handler reads back
+the row and uses it to trigger ScoreService.calculate_score for the customer.
+"""
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = "add_engagement_events"
+down_revision: str | None = "e8ad92ca62b3"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    op.create_table(
+        "engagement_events",
+        sa.Column("id", sa.Integer(), autoincrement=True, nullable=False),
+        sa.Column("tenant_id", sa.Integer(), nullable=False),
+        sa.Column("customer_id", sa.Integer(), nullable=False),
+        sa.Column("event_type", sa.String(length=50), nullable=False),
+        sa.Column("event_metadata", postgresql.JSONB(astext_type=sa.Text()), nullable=True),
+        sa.Column(
+            "created_at",
+            sa.DateTime(timezone=True),
+            server_default=sa.text("now()"),
+            nullable=False,
+        ),
+        sa.PrimaryKeyConstraint("id"),
+    )
+    op.create_index(
+        op.f("ix_engagement_events_tenant_id"),
+        "engagement_events",
+        ["tenant_id"],
+        unique=False,
+    )
+    op.create_index(
+        op.f("ix_engagement_events_customer_id"),
+        "engagement_events",
+        ["customer_id"],
+        unique=False,
+    )
+
+
+def downgrade() -> None:
+    op.drop_index(op.f("ix_engagement_events_customer_id"), table_name="engagement_events")
+    op.drop_index(op.f("ix_engagement_events_tenant_id"), table_name="engagement_events")
+    op.drop_table("engagement_events")
--- a/alembic/versions/add_engagement_events_table.py
+++ b/alembic/versions/add_engagement_events_table.py
@@ -1,6 +1,6 @@
 """add_engagement_events_table
 
-Revision ID: add_engagement_events
+Revision ID: a1b2c3d4e5f6
 Revises: e8ad92ca62b3
 Create Date: 2026-06-06 10:00:00.000000
 
@@ -17,7 +17,7 @@ from sqlalchemy.dialects import postgresql
 from alembic import op
 
 # revision identifiers, used by Alembic.
-revision: str = "add_engagement_events"
+revision: str = "a1b2c3d4e5f6"
 down_revision: str | None = "e8ad92ca62b3"
 branch_labels: str | Sequence[str] | None = None
 depends_on: str | Sequence[str] | None = None
@@ -27,10 +27,25 @@ def upgrade() -> None:
     op.create_table(
         "engagement_events",
         sa.Column("id", sa.Integer(), autoincrement=True, nullable=False),
-        sa.Column("tenant_id", sa.Integer(), nullable=False),
-        sa.Column("customer_id", sa.Integer(), nullable=False),
+        sa.Column(
+            "tenant_id",
+            sa.Integer(),
+            sa.ForeignKey("tenants.id", ondelete="CASCADE"),
+            nullable=False,
+        ),
+        sa.Column(
+            "customer_id",
+            sa.Integer(),
+            sa.ForeignKey("customers.id", ondelete="CASCADE"),
+            nullable=False,
+        ),
         sa.Column("event_type", sa.String(length=50), nullable=False),
-        sa.Column("event_metadata", postgresql.JSONB(astext_type=sa.Text()), nullable=True),
+        sa.Column(
+            "event_metadata",
+            postgresql.JSONB(astext_type=sa.Text()),
+            nullable=True,
+            server_default=sa.text("'{}'"),
+        ),
         sa.Column(
             "created_at",
             sa.DateTime(timezone=True),
@@ -51,9 +66,23 @@ def upgrade() -> None:
         ["customer_id"],
         unique=False,
     )
+    op.create_index(
+        "ix_engagement_events_tenant_customer",
+        "engagement_events",
+        ["tenant_id", "customer_id"],
+        unique=False,
+    )
+    op.create_index(
+        "ix_engagement_events_tenant_event_type",
+        "engagement_events",
+        ["tenant_id", "event_type"],
+        unique=False,
+    )
 
 
 def downgrade() -> None:
+    op.drop_index("ix_engagement_events_tenant_event_type", table_name="engagement_events")
+    op.drop_index("ix_engagement_events_tenant_customer", table_name="engagement_events")
     op.drop_index(op.f("ix_engagement_events_customer_id"), table_name="engagement_events")
     op.drop_index(op.f("ix_engagement_events_tenant_id"), table_name="engagement_events")
     op.drop_table("engagement_events")
--- a/src/api/routers/customers.py
+++ b/src/api/routers/customers.py
@@ -216,7 +216,7 @@ async def list_customers(
     page_size: int = Query(20, ge=1, le=100),
     status: str | None = None,
     owner_id: int | None = Query(None, ge=0),
-    lead_tier: str | None = Query(None, description="Filter by lead tier: hot, warm, cold"),
+    lead_tier: str | None = Query(None, pattern="^(hot|warm|cold)$", description="Filter by lead tier: hot, warm, cold"),
     order_by_score: bool = Query(False, description="Auto-rank by score descending"),
     ctx: AuthContext = Depends(require_auth),
     session: AsyncSession = Depends(get_db),
--- a/src/api/routers/customers.py
+++ b/src/api/routers/customers.py
@@ -216,6 +216,8 @@ async def list_customers(
     page_size: int = Query(20, ge=1, le=100),
     status: str | None = None,
     owner_id: int | None = Query(None, ge=0),
+    lead_tier: str | None = Query(None, description="Filter by lead tier: hot, warm, cold"),
+    order_by_score: bool = Query(False, description="Auto-rank by score descending"),
     ctx: AuthContext = Depends(require_auth),
     session: AsyncSession = Depends(get_db),
 ):
@@ -226,6 +228,8 @@ async def list_customers(
         page_size=page_size,
         status=status,
         owner_id=owner_id,
+        lead_tier=lead_tier,
+        order_by_score=order_by_score,
         tenant_id=ctx.tenant_id,
     )
     customer_ids = [getattr(c, "id", None) for c in items]
--- a/src/api/routers/events.py
+++ b/src/api/routers/events.py
@@ -1,0 +1,50 @@
+"""Events router — POST /engagement webhook for customer engagement events."""
+
+from __future__ import annotations
+
+from fastapi import APIRouter, Depends
+from pydantic import BaseModel, Field
+from sqlalchemy.ext.asyncio import AsyncSession
+
+from db.connection import get_db
+from internal.middleware.fastapi_auth import AuthContext, require_auth
+from services.event_service import EventService
+from services.score_service import ScoreService
+
+router = APIRouter(prefix="/api/v1/events", tags=["events"])
+
+
+class EngagementEventRequest(BaseModel):
+    customer_id: int = Field(..., gt=0)
+    event_type: str = Field(..., pattern="^(email_open|website_visit)$")
+    metadata: dict | None = None
+
+
+@router.post("/engagement")
+async def create_engagement_event(
+    body: EngagementEventRequest,
+    ctx: AuthContext = Depends(require_auth),
+    session: AsyncSession = Depends(get_db),
+):
+    """Record an engagement event and trigger a score recalculation for the customer."""
+    event_svc = EventService(session)
+    score_svc = ScoreService(session)
+
+    await event_svc.record_engagement_event(
+        tenant_id=ctx.tenant_id,
+        customer_id=body.customer_id,
+        event_type=body.event_type,
+        metadata=body.metadata,
+    )
+    result = await score_svc.calculate_score(
+        customer_id=body.customer_id,
+        tenant_id=ctx.tenant_id,
+    )
+    return {
+        "success": True,
+        "data": {
+            "customer_id": body.customer_id,
+            "score": result.score,
+            "tier": result.tier_label,
+        },
+    }
--- a/src/api/routers/events.py
+++ b/src/api/routers/events.py
@@ -17,10 +17,10 @@ router = APIRouter(prefix="/api/v1/events", tags=["events"])
 class EngagementEventRequest(BaseModel):
     customer_id: int = Field(..., gt=0)
     event_type: str = Field(..., pattern="^(email_open|website_visit)$")
-    metadata: dict | None = None
+    event_metadata: dict | None = None
 
 
-@router.post("/engagement")
+@router.post("/engagement", status_code=200)
 async def create_engagement_event(
     body: EngagementEventRequest,
     ctx: AuthContext = Depends(require_auth),
@@ -34,7 +34,7 @@ async def create_engagement_event(
         tenant_id=ctx.tenant_id,
         customer_id=body.customer_id,
         event_type=body.event_type,
-        metadata=body.metadata,
+        metadata=body.event_metadata,
     )
     result = await score_svc.calculate_score(
         customer_id=body.customer_id,
@@ -47,4 +47,5 @@ async def create_engagement_event(
             "score": result.score,
             "tier": result.tier_label,
         },
+        "message": "Engagement event recorded successfully",
     }
--- a/src/db/models/engagement.py
+++ b/src/db/models/engagement.py
@@ -1,0 +1,24 @@
+"""Engagement event ORM model — records email opens, website visits, etc."""
+
+from __future__ import annotations
+
+from datetime import datetime
+
+from sqlalchemy import DateTime, Integer, String, func
+from sqlalchemy.dialects.postgresql import JSONB
+from sqlalchemy.orm import Mapped, mapped_column
+
+from db.base import Base
+
+
+class EngagementEventModel(Base):
+    """Engagement event entity mapped to the ``engagement_events`` table."""
+
+    __tablename__ = "engagement_events"
+
+    id: Mapped[int] = mapped_column(Integer, primary_key=True, autoincrement=True)
+    tenant_id: Mapped[int] = mapped_column(Integer, nullable=False, index=True)
+    customer_id: Mapped[int] = mapped_column(Integer, nullable=False, index=True)
+    event_type: Mapped[str] = mapped_column(String(50), nullable=False)
+    event_metadata: Mapped[dict | None] = mapped_column(JSONB, nullable=True, default=dict)
+    created_at: Mapped[datetime] = mapped_column(DateTime(timezone=True), server_default=func.now(), nullable=False)
--- a/src/db/models/engagement.py
+++ b/src/db/models/engagement.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 
 from datetime import datetime
 
-from sqlalchemy import DateTime, Integer, String, func
+from sqlalchemy import DateTime, ForeignKey, Integer, String, func
 from sqlalchemy.dialects.postgresql import JSONB
 from sqlalchemy.orm import Mapped, mapped_column
 
@@ -17,8 +17,12 @@ class EngagementEventModel(Base):
     __tablename__ = "engagement_events"
 
     id: Mapped[int] = mapped_column(Integer, primary_key=True, autoincrement=True)
-    tenant_id: Mapped[int] = mapped_column(Integer, nullable=False, index=True)
-    customer_id: Mapped[int] = mapped_column(Integer, nullable=False, index=True)
+    tenant_id: Mapped[int] = mapped_column(
+        Integer, ForeignKey("tenants.id", ondelete="CASCADE"), nullable=False, index=True
+    )
+    customer_id: Mapped[int] = mapped_column(
+        Integer, ForeignKey("customers.id", ondelete="CASCADE"), nullable=False, index=True
+    )
     event_type: Mapped[str] = mapped_column(String(50), nullable=False)
     event_metadata: Mapped[dict | None] = mapped_column(JSONB, nullable=True, default=dict)
     created_at: Mapped[datetime] = mapped_column(DateTime(timezone=True), server_default=func.now(), nullable=False)
--- a/src/services/customer_service.py
+++ b/src/services/customer_service.py
@@ -45,6 +45,15 @@ class CustomerService:
 
         return customer
 
+    # Maps the public lead_tier filter values (hot/warm/cold) to the stored
+    # ScoreTier letter values (A/B/C) that the customer.tier column carries.
+    # Tier "D" customers are below the cold threshold and excluded by design.
+    LEAD_TIER_TO_STORED: dict[str, str] = {
+        "hot": "A",
+        "warm": "B",
+        "cold": "C",
+    }
+
     async def list_customers(
         self,
         tenant_id: int,
@@ -52,14 +61,23 @@ class CustomerService:
         page_size: int = 20,
         status: str | None = None,
         owner_id: int | None = None,
+        lead_tier: str | None = None,
+        order_by_score: bool = False,
     ) -> tuple[list[Any], int]:
         """List customers for tenant with optional filters."""
+        stored_tier: str | None = None
+        if lead_tier is not None:
+            if lead_tier not in self.LEAD_TIER_TO_STORED:
+                raise ValidationException("lead_tier must be one of: hot, warm, cold")
+            stored_tier = self.LEAD_TIER_TO_STORED[lead_tier]
         return await self.repository.list_customers(
             tenant_id=tenant_id,
             page=page,
             page_size=page_size,
             status=status,
             owner_id=owner_id,
+            lead_tier=stored_tier,
+            order_by_score=order_by_score,
         )
 
     async def get_customer(self, customer_id: int, tenant_id: int) -> Any:
--- a/src/services/event_service.py
+++ b/src/services/event_service.py
@@ -4,16 +4,20 @@ from __future__ import annotations
 
 from typing import Any
 
+from sqlalchemy import and_, select
 from sqlalchemy.ext.asyncio import AsyncSession
 
+from db.models.customer import CustomerModel
 from db.models.engagement import EngagementEventModel
-from pkg.errors.app_exceptions import ValidationException
+from pkg.errors.app_exceptions import NotFoundException, ValidationException
 
 VALID_EVENT_TYPES = {"email_open", "website_visit"}
 
 
 class EventService:
     """Business logic for engagement events — persists EngagementEventModel rows."""
+
+    VALID_EVENT_TYPES = VALID_EVENT_TYPES
 
     def __init__(self, session: AsyncSession) -> None:
         self.session = session
@@ -33,8 +37,20 @@ class EventService:
         commit the session themselves if they need the row visible to
         subsequent reads.
         """
-        if event_type not in VALID_EVENT_TYPES:
-            raise ValidationException(f"event_type must be one of: {', '.join(sorted(VALID_EVENT_TYPES))}")
+        if event_type not in self.VALID_EVENT_TYPES:
+            raise ValidationException(f"event_type must be one of: {', '.join(sorted(self.VALID_EVENT_TYPES))}")
+
+        customer_exists = await self.session.execute(
+            select(CustomerModel.id).where(
+                and_(
+                    CustomerModel.id == customer_id,
+                    CustomerModel.tenant_id == tenant_id,
+                )
+            )
+        )
+        if customer_exists.scalar_one_or_none() is None:
+            raise NotFoundException("Customer")
+
         event = EngagementEventModel(
             tenant_id=tenant_id,
             customer_id=customer_id,
--- a/src/services/event_service.py
+++ b/src/services/event_service.py
@@ -1,0 +1,47 @@
+"""EventService — records customer engagement events."""
+
+from __future__ import annotations
+
+from typing import Any
+
+from sqlalchemy.ext.asyncio import AsyncSession
+
+from db.models.engagement import EngagementEventModel
+from pkg.errors.app_exceptions import ValidationException
+
+VALID_EVENT_TYPES = {"email_open", "website_visit"}
+
+
+class EventService:
+    """Business logic for engagement events — persists EngagementEventModel rows."""
+
+    def __init__(self, session: AsyncSession) -> None:
+        self.session = session
+
+    async def record_engagement_event(
+        self,
+        tenant_id: int,
+        customer_id: int,
+        event_type: str,
+        metadata: dict[str, Any] | None = None,
+    ) -> EngagementEventModel:
+        """Insert a new engagement event and return the refreshed ORM object.
+
+        Flushes so the auto-generated id is populated, but does not commit —
+        the router-owned transaction boundary (Rule 121) commits on normal
+        exit. Callers that invoke this service outside a router context must
+        commit the session themselves if they need the row visible to
+        subsequent reads.
+        """
+        if event_type not in VALID_EVENT_TYPES:
+            raise ValidationException(f"event_type must be one of: {', '.join(sorted(VALID_EVENT_TYPES))}")
+        event = EngagementEventModel(
+            tenant_id=tenant_id,
+            customer_id=customer_id,
+            event_type=event_type,
+            event_metadata=metadata or {},
+        )
+        self.session.add(event)
+        await self.session.flush()
+        await self.session.refresh(event)
+        return event
--- a/tests/integration/test_engagement_webhook_integration.py
+++ b/tests/integration/test_engagement_webhook_integration.py
@@ -7,7 +7,7 @@ Run against a real PostgreSQL database (Supabase via DATABASE_URL env var):
 from __future__ import annotations
 
 import uuid
-from datetime import datetime, timezone
+from datetime import datetime, timedelta, timezone
 
 import pytest
 from sqlalchemy import select
@@ -54,8 +54,6 @@ async def _seed_customer(
 
 async def _seed_tenant(async_session: AsyncSession, tenant_id: int) -> None:
     """Insert a tenant row so FK references are satisfied. Idempotent — no-op if already present."""
-    from sqlalchemy import select
-
     result = await async_session.execute(select(TenantModel).where(TenantModel.id == tenant_id))
     if result.scalar_one_or_none() is not None:
         return
@@ -97,13 +95,14 @@ class TestEngagementWebhookIntegration:
             metadata=metadata,
         )
         # EventService uses flush() (Rule 121 — routers own commit).
-        # Commit here so the re-read below sees the row.
-        await async_session.commit()
+        # Flush here so the re-read below sees the row.
+        await async_session.flush()
 
         assert event.id is not None
         assert event.tenant_id == tenant_id
         assert event.customer_id == customer_id
         assert event.event_type == "email_open"
+        # event_metadata is the ORM attribute name (column: engagement_events.event_metadata)
         assert event.event_metadata == metadata
         assert event.created_at is not None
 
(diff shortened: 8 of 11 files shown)


src/api/routers/events.py
```diff
@@ -0,0 +1,50 @@
...
+from services.event_service import EventService
+from services.score_service import ScoreService
+
+router = APIRouter(prefix="/api/v1/events", tags=["events"])
+
+
+class EngagementEventRequest(BaseModel):
+    customer_id: int = Field(..., gt=0)
+    event_type: str = Field(..., pattern="^(email_open|website_visit)$")
+    metadata: dict | None = None
```
yihua-zhuo: **CRITICAL** (critical)

**Issue:** Pydantic `metadata` is a reserved attribute on `BaseModel` (it is `model.metadata` class-level config); using it as a field name causes a `UserWarning` and may break serialization/alias logic. This should be renamed to `event_metadata`.

**Suggestion:** Rename the field: `event_metadata: dict | None = None` — and update the service call on line 37 to `metadata=body.event_metadata`.

src/db/models/engagement.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class EngagementEventModel(Base):
+    """Engagement event entity mapped to the ``engagement_events`` table."""
+
+    __tablename__ = "engagement_events"
+
+    id: Mapped[int] = mapped_column(Integer, primary_key=True, autoincrement=True)
+    tenant_id: Mapped[int] = mapped_column(Integer, nullable=False, index=True)
+    customer_id: Mapped[int] = mapped_column(Integer, nullable=False, index=True)
```
yihua-zhuo: **CRITICAL** (critical)

**Issue:** customer_id has no foreign key to customers.id — orphan rows possible and referential integrity is not enforced.

**Suggestion:** Add ForeignKey('customers.id', ondelete='CASCADE') and consider a composite UniqueConstraint or composite index on (tenant_id, customer_id) for query patterns.

tests/integration/test_engagement_webhook_integration.py
```diff
@@ -0,0 +1,322 @@
...
+        metadata = {"campaign": "spring", "source": "newsletter"}
+        event = await svc.record_engagement_event(
+            tenant_id=tenant_id,
+            customer_id=customer_id,
+            event_type="email_open",
+            metadata=metadata,
+        )
+        # EventService uses flush() (Rule 121 — routers own commit).
+        # Commit here so the re-read below sees the row.
+        await async_session.commit()
```
yihua-zhuo: **CRITICAL** (critical)

**Issue:** Integration test calls `await async_session.commit()` — Rule 124 says integration tests must use real DB fixtures, and Rule 121 says the transaction boundary lives at the router layer; the test itself is not a router and should not commit.

**Suggestion:** Remove the explicit commits; if cross-statement visibility is needed, call `await async_session.flush()` (which the service already does internally) and rely on the fixture's teardown.

tests/integration/test_engagement_webhook_integration.py
```diff
@@ -0,0 +1,322 @@
...
+            tenant_id=tenant_id,
+            customer_id=customer_id,
+            event_type="email_open",
+        )
+        # EventService uses flush() (Rule 121 — routers own commit).
+        # Commit here so ScoreService's select sees the new score_factors-free
+        # customer row (the seed above already flushed, so this commit is
+        # a no-op for that row, but the service call's flush needs commit
+        # for the new event row — the cross-test read in test 3 needs it).
+        await async_session.commit()
```
yihua-zhuo: **CRITICAL** (critical)

**Issue:** Same issue — `await async_session.commit()` is called in a test to force cross-test visibility, violating Rule 121 and Rule 124's spirit that tests exercise actual transaction semantics.

**Suggestion:** Use a dedicated `db_schema` / per-transaction pattern or merge the two operations into a single transactional fixture call rather than committing manually.

tests/integration/test_engagement_webhook_integration.py
```diff
@@ -0,0 +1,322 @@
...
+        from services.event_service import EventService
+
+        svc = EventService(async_session)
+        await svc.record_engagement_event(
+            tenant_id=tenant_id,
+            customer_id=customer_id_a,
+            event_type="email_open",
+        )
+        # Commit so the event row is visible to the cross-tenant score check.
+        await async_session.commit()
```
yihua-zhuo: **CRITICAL** (critical)

**Issue:** Third instance of `await async_session.commit()` in the integration test, again violating Rule 121 — services must not commit and tests should not paper over that boundary.

**Suggestion:** Use `flush()` for visibility, or restructure the test so the cross-tenant check runs within the same flushed transaction.

tests/unit/test_events_webhook.py
```diff
@@ -0,0 +1,194 @@
...
+        from services.event_service import EventService
+
+        svc = EventService(MagicMock())
+        with pytest.raises(ValidationException, match="event_type must be one of"):
+            await svc.record_engagement_event(tenant_id=1, customer_id=1, event_type="bogus")
+
+    @pytest.mark.asyncio
+    async def test_valid_event_types_accepted(self):
+        """All event types in VALID_EVENT_TYPES are accepted by the service."""
+        for event_type in VALID_EVENT_TYPES:
```
yihua-zhuo, line 208: **CRITICAL** (critical)

**Issue:** `assert event_type in {"email_open", "website_visit"}` — the test iterates over `VALID_EVENT_TYPES` imported from the service and asserts each is in a hardcoded two-element set. If `VALID_EVENT_TYPES` grows beyond two entries, this assertion silently passes for the first two and fails only for the rest, and the intent (the set *is* the full allowlist) is lost.

**Suggestion:** Replace with `assert set(VALID_EVENT_TYPES) == {"email_open", "website_visit"}` (exact equality) or `assert all(et in {"email_open", "website_visit"} for et in VALID_EVENT_TYPES)` with a comment that the allowlist must be exactly these two.

.plans/issue-586.md
```diff
@@ -0,0 +1,79 @@
...
+2. `/home/runner/work/agent-job/agent-job/docs/dev-plan/_template-deep.md`
+3. `/home/runner/work/agent-job/agent-job/docs/dev-plan/10-customers/0586-add-lead-tier-filter-and-auto-rank-to-customer-list-endpoint.md`
+
+## Affected Files
+- `src/api/routers/customers.py` — add `lead_tier` and `order_by_score` query params to `list_customers`; thread them to `CustomerService.list_customers`
+- `src/services/customer_service.py` — extend `list_customers` signature with `lead_tier` and `order_by_score`; validate `lead_tier` against allowed set; pass through to `CustomerRepository.list_customers`
+- `src/db/repositories/customer.py` — extend `list_customers` signature with `lead_tier` and `order_by_score`; add `CustomerModel.tier` equality filter and conditional `ORDER BY score DESC` (with `COALESCE(score, 0)` to handle NULL)
+- `src/db/models/engagement.py` — new file; define `EngagementEventModel` (id, tenant_id, customer_id, event_type, event_metadata JSONB, created_at)
+- `src/services/event_service.py` — new file; define `EventService` with `record_engagement_event(tenant_id, customer_id, event_type, metadata) -> EngagementEventModel`
```
yihua-zhuo, line 21: **WARNING** (suggestion)

**Issue:** Plan says `record_engagement_event` does 'adds it, commits, refreshes, and returns' — but Rule 121 prohibits commit in the service layer; the implementation correctly uses flush() but the plan text contradicts the rule and the code.

**Suggestion:** Update step 6 to say 'adds it, flushes, refreshes, and returns' to match Rule 121 and the actual implementation.

.plans/issue-586.md
```diff
@@ -0,0 +1,79 @@
...
+- `src/api/routers/events.py` — new file; `APIRouter(prefix="/api/v1/events", tags=["events"])` with `POST /engagement` endpoint that calls `EventService.record_engagement_event` then `ScoreService.calculate_score`, returns `{"success": True, "data": {"customer_id": N, "score": M, "tier": "A"}}`
+- `alembic/versions/<id>_add_engagement_events.py` — new migration; `CREATE TABLE engagement_events` with `tenant_id` index and `event_metadata` JSONB; manually verify `sa.JSONB()` and `sa.DateTime(timezone=True)`
```
yihua-zhuo, line 23: **WARNING** (suggestion)

**Issue:** Plan describes `EngagementEventRequest` with `metadata: dict | None = None` — using a Pydantic-reserved field name `metadata` is a known anti-pattern that conflicts with `BaseModel.metadata`; should be `event_metadata` or aliased.

**Suggestion:** Rename the field to `event_metadata: dict | None = None` in the plan (and verify the router uses the same name).

.plans/issue-586.md
```diff
@@ -0,0 +1,79 @@
...
+- The engagement event row is persisted with the correct `tenant_id`, `customer_id`, `event_type`, and `event_metadata`
+- Cross-tenant isolation holds: a webhook POST for `tenant_id=A` is not visible or actionable from `tenant_id=B`
+- All unit and integration tests pass; `ruff check` and `ruff format --check` are clean; migration applies and reverts cleanly in a three-cycle
+
+## Risks / Open Questions
+- **Tier value mapping**: The dev-plan body says `lead_tier` accepts `hot/warm/cold`, but the existing `CustomerModel.tier` column stores the `ScoreTier` enum values (`A`/`B`/`C`/`D`). The plan adds a `{"hot", "warm", "cold"}` validation set in the service layer, but the actual SQL filter operates on the stored letter tier. If callers expect hot/warm/cold to map to A/B/C, the mapping must be defined at the service or router boundary — not in the SQL query. This decision should be confirmed with the issue owner before implementation, as it may require updating the filter to translate input values to the stored enum.
```
yihua-zhuo, line 77: **INFO** (suggestion)

**Issue:** Plan raises an open question about the `hot/warm/cold` to `A/B/C` mapping but does not commit to the resolution; the implementation already maps it in the service, yet the plan's API description on line 6 still says the filter 'returns only customers whose tier column matches' which is misleading for callers who pass 'hot' but get tier='A' rows.

**Suggestion:** Resolve the open question in the plan: confirm the mapping is intentional, and update the acceptance criteria to reflect that `lead_tier=hot` filters on stored `tier='A'`.

.plans/issue-586.md
```diff
@@ -0,0 +1,79 @@
...
+- **Webhook auth model**: The dev-plan Step 3 uses `Depends(require_auth)`, which expects a user JWT. External systems (email trackers, website analytics) typically authenticate via a shared secret or API key, not user credentials. The plan follows the dev-plan's auth choice, but a production webhook would more likely use a header-based secret check. This should be confirmed with the issue owner if the webhook is intended for external system integration.
+- **ScoreService persistence**: `ScoreService.calculate_score` (in `src/services/score_service.py`) is currently read-only — it computes a `ScoreResult` from the persisted `score_factors` JSON column but does not write the computed `score`/`tier` back to `CustomerModel`. The webhook calls `calculate_score` and then must persist the result. The plan assumes the webhook handler (or a follow-up) updates `customer.score` and `customer.tier` after calling `calculate_score`; if `ScoreService` is extended to persist as part of #585, this becomes simpler. This should be verified against the #585 implementation before starting work.
```
yihua-zhuo, line 79: **INFO** (suggestion)

**Issue:** Plan flags `ScoreService.calculate_score` as not persisting results as a risk, but the acceptance criteria say the webhook 'invokes calculate_score exactly once' and returns the computed score — the integration tests confirm the score is returned in the response but the plan does not verify that the customer's stored `score`/`tier` columns are updated, leaving a gap between the response and DB state.

**Suggestion:** Add an explicit acceptance check that the customer's persisted `score` and `tier` columns are updated after the webhook fires, or document why they are not.

.plans/issue-586.md
```diff
@@ -0,0 +1,79 @@
...
+
+5. **Generate and fix Alembic migration**: run `alembic revision --autogenerate -m "add engagement_events table"` against the `alembic_dev` DB; manually correct `sa.JSON()` → `sa.JSONB()` on the `event_metadata` column; verify `sa.DateTime(timezone=True)` on `created_at`; run `alembic upgrade head && alembic downgrade -1 && alembic upgrade head` three-cycle; run a second `autogenerate` to confirm empty drift, then delete the drift-check file.
+
+6. **Create `EventService`** in `src/services/event_service.py`: `__init__(self, session: AsyncSession)` with no default; `record_engagement_event(tenant_id, customer_id, event_type, metadata=None) -> EngagementEventModel` that constructs an `EngagementEventModel` instance, adds it, commits, refreshes, and returns. Validate `event_type` is in `{"email_open", "website_visit"}` and raise `ValidationException` otherwise.
```
yihua-zhuo, line 40: **INFO** (suggestion)

**Issue:** Plan step 6 says the service 'validates event_type is in {email_open, website_visit} and raises ValidationException' — this is redundant with the Pydantic `pattern` constraint on the request schema, and the duplicate check means the validation message may differ depending on which layer catches it.

**Suggestion:** Pick one validation layer. The Pydantic pattern already returns 422; remove the service-level check or drop the Pydantic pattern and rely solely on the service.

alembic/versions/add_engagement_events_table.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def upgrade() -> None:
+    op.create_table(
+        "engagement_events",
+        sa.Column("id", sa.Integer(), autoincrement=True, nullable=False),
+        sa.Column("tenant_id", sa.Integer(), nullable=False),
+        sa.Column("customer_id", sa.Integer(), nullable=False),
+        sa.Column("event_type", sa.String(length=50), nullable=False),
+        sa.Column("event_metadata", postgresql.JSONB(astext_type=sa.Text()), nullable=True),
```
yihua-zhuo: **INFO** (suggestion)

**Issue:** `event_metadata` column is nullable=True with no default; the model sets `default=dict`, so DB-level NULL is possible for rows inserted before the model default was applied or by other paths — this is a minor schema/impedance mismatch.

**Suggestion:** Add `server_default=sa.text("'{}'")` to align the DB default with the model default, preventing NULLs.

alembic/versions/add_engagement_events_table.py
```diff
@@ -0,0 +1,59 @@
...
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = "add_engagement_events"
+down_revision: str | None = "e8ad92ca62b3"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
```
yihua-zhuo, line 26: **INFO** (suggestion)

**Issue:** Migration does not create a composite index on `(tenant_id, customer_id)` or `(tenant_id, event_type)` which are the likely query patterns for tenant-scoped engagement lookups and scoring aggregations.

**Suggestion:** Add `op.create_index(..., ['tenant_id', 'customer_id'])` and/or `(tenant_id, event_type)` for common lookup patterns.

alembic/versions/add_engagement_events_table.py
```diff
@@ -0,0 +1,59 @@
...
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = "add_engagement_events"
```
yihua-zhuo: **INFO** (suggestion)

**Issue:** Revision ID `add_engagement_events` is not a standard Alembic hex hash — not a bug, but inconsistent with all other migrations in the project that use hex revision IDs (e.g., `e8ad92ca62b3`, `127938d15761`).

**Suggestion:** Use a standard hex revision ID like `a1b2c3d4e5f6` to match the project convention.

src/api/routers/customers.py
```diff
@@ -216,6 +216,8 @@ async def list_customers(
     page_size: int = Query(20, ge=1, le=100),
     status: str | None = None,
     owner_id: int | None = Query(None, ge=0),
+    lead_tier: str | None = Query(None, description="Filter by lead tier: hot, warm, cold"),
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** `lead_tier` query param has no validation pattern — invalid values (e.g. `lead_tier=foo`) are accepted at the router level and only rejected inside `CustomerService.list_customers` as a `ValidationException` (422). The plan implies a pattern constraint matching `{hot, warm, cold}` should be applied at the boundary.

**Suggestion:** Add `pattern="^(hot|warm|cold)$"` to the `lead_tier: str | None = Query(...)` parameter to reject invalid values at the FastAPI boundary with 422, consistent with `CUSTOMER_STATUS_PATTERN` already used for `status`.

src/api/routers/events.py
```diff
@@ -0,0 +1,50 @@
...
+
+from db.connection import get_db
+from internal.middleware.fastapi_auth import AuthContext, require_auth
+from services.event_service import EventService
+from services.score_service import ScoreService
+
+router = APIRouter(prefix="/api/v1/events", tags=["events"])
+
+
+class EngagementEventRequest(BaseModel):
```
yihua-zhuo, line 17: **WARNING** (suggestion)

**Issue:** Pydantic `pattern="^(email_open|website_visit)$"` is duplicated by `EventService.VALID_EVENT_TYPES` validation — two layers of validation for the same constraint can diverge and produce inconsistent error messages.

**Suggestion:** Remove the Pydantic `pattern` and let `EventService.record_engagement_event` be the sole validator (returning a 422 via `ValidationException`), or remove the service-level check and rely solely on the Pydantic pattern. Pick one.

src/api/routers/events.py
```diff
@@ -0,0 +1,50 @@
...
+    customer_id: int = Field(..., gt=0)
+    event_type: str = Field(..., pattern="^(email_open|website_visit)$")
+    metadata: dict | None = None
+
+
+@router.post("/engagement")
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** No `status_code` is declared on the `@router.post` — the plan and test fixtures expect 200, which is the default, but the plan's response shape is non-standard (raw dict instead of the `ApiResponse` envelope used by other routers), making the endpoint inconsistent with the rest of the API surface.

**Suggestion:** Either use a Pydantic response model and `ApiResponse` envelope to match the project pattern, or document why this webhook endpoint diverges from the standard envelope.

src/api/routers/events.py
```diff
@@ -0,0 +1,50 @@
...
+async def create_engagement_event(
+    body: EngagementEventRequest,
+    ctx: AuthContext = Depends(require_auth),
+    session: AsyncSession = Depends(get_db),
```
yihua-zhuo, line 27: **INFO** (suggestion)

**Issue:** `Depends(require_auth)` uses user JWT auth for what is described as a webhook for external systems (email trackers, website analytics) — the plan itself flags this as a concern (line 78 of the plan). External webhooks typically use shared-secret or API-key header auth, not user credentials.

**Suggestion:** Consider an alternative auth dependency (e.g., a header-based shared-secret check) appropriate for machine-to-machine webhook calls, or document the decision to use user JWT.

src/api/routers/events.py
```diff
@@ -0,0 +1,50 @@
...
+):
+    """Record an engagement event and trigger a score recalculation for the customer."""
+    event_svc = EventService(session)
+    score_svc = ScoreService(session)
+
+    await event_svc.record_engagement_event(
```
yihua-zhuo, line 33: **INFO** (suggestion)

**Issue:** The webhook calls `event_svc.record_engagement_event` then `score_svc.calculate_score` but never persists the recalculated `score`/`tier` back to the `customer` row — `ScoreService.calculate_score` is read-only (it returns a `ScoreResult` but does not write to the DB), so the customer's stored `score` and `tier` columns remain stale after the webhook fires.

**Suggestion:** After calling `calculate_score`, update the customer row: `customer.score = result.score; customer.tier = result.tier.value` and let the router commit (Rule 121).

src/db/models/engagement.py
```diff
@@ -0,0 +1,24 @@
...
+from db.base import Base
+
+
+class EngagementEventModel(Base):
+    """Engagement event entity mapped to the ``engagement_events`` table."""
+
+    __tablename__ = "engagement_events"
+
+    id: Mapped[int] = mapped_column(Integer, primary_key=True, autoincrement=True)
+    tenant_id: Mapped[int] = mapped_column(Integer, nullable=False, index=True)
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** tenant_id is a bare int column without a ForeignKey to tenants.id, violating rule 138 which requires a proper FK for tenant-owned tables.

**Suggestion:** Add ForeignKey('tenants.id', ondelete='CASCADE') to enforce referential integrity at the DB level.

src/db/models/engagement.py
```diff
@@ -0,0 +1,24 @@
...
+    customer_id: Mapped[int] = mapped_column(Integer, nullable=False, index=True)
+    event_type: Mapped[str] = mapped_column(String(50), nullable=False)
+    event_metadata: Mapped[dict | None] = mapped_column(JSONB, nullable=True, default=dict)
```
yihua-zhuo, line 27: **INFO** (suggestion)

**Issue:** event_type is a free-form String(50) with no validation at the model level; only EventService validates values at runtime.

**Suggestion:** Consider an Enum type or a CHECK constraint in a migration to prevent invalid values from being inserted by other code paths.

src/services/event_service.py
```diff
@@ -0,0 +1,47 @@
...
+from __future__ import annotations
+
+from typing import Any
+
+from sqlalchemy.ext.asyncio import AsyncSession
+
+from db.models.engagement import EngagementEventModel
+from pkg.errors.app_exceptions import ValidationException
+
+VALID_EVENT_TYPES = {"email_open", "website_visit"}
```
yihua-zhuo, line 14: **INFO** (suggestion)

**Issue:** VALID_EVENT_TYPES is a module-level set, but EventService is the only consumer — consider making it a class-level constant for namespacing and to support subclass overrides.

**Suggestion:** Move VALID_EVENT_TYPES inside the class as a class attribute, or keep it module-level with a comment that it is intentionally shared.

src/services/event_service.py
```diff
@@ -0,0 +1,47 @@
...
+        """Insert a new engagement event and return the refreshed ORM object.
+
+        Flushes so the auto-generated id is populated, but does not commit —
+        the router-owned transaction boundary (Rule 121) commits on normal
+        exit. Callers that invoke this service outside a router context must
+        commit the session themselves if they need the row visible to
+        subsequent reads.
+        """
+        if event_type not in VALID_EVENT_TYPES:
+            raise ValidationException(f"event_type must be one of: {', '.join(sorted(VALID_EVENT_TYPES))}")
```
yihua-zhuo: **INFO** (suggestion)

**Issue:** No tenant ownership verification: the service trusts the caller to pass a valid tenant_id and customer_id combination, but does not verify the customer belongs to that tenant.

**Suggestion:** Add a check (or rely on a FK constraint at the DB level) that customer_id belongs to the given tenant_id, or document that this is intentionally the caller's responsibility.

src/services/event_service.py
```diff
@@ -0,0 +1,47 @@
...
+        event = EngagementEventModel(
+            tenant_id=tenant_id,
+            customer_id=customer_id,
+            event_type=event_type,
+            event_metadata=metadata or {},
+        )
+        self.session.add(event)
```
yihua-zhuo, line 60: **INFO** (suggestion)

**Issue:** metadata=metadata or {} creates a new empty dict per call — fine for correctness, but if callers frequently pass None this allocation could be avoided by using a shared sentinel.

**Suggestion:** No change needed unless profiling shows this is a hot path; the current approach is readable and safe.

tests/integration/test_engagement_webhook_integration.py
```diff
@@ -0,0 +1,322 @@
...
+    async def test_lead_tier_invalid_raises_validation_in_service(
+        self, db_schema, tenant_id, async_session
+    ):
+        """Service-layer validation rejects unknown lead_tier values with ValidationException."""
+        await _seed_tenant(async_session, tenant_id)
+        await _seed_customer(async_session, tenant_id)
+
+        from pkg.errors.app_exceptions import ValidationException
+
+        svc = CustomerService(CustomerRepository(async_session))
```
yihua-zhuo, line 326: **WARNING** (suggestion)

**Issue:** `CustomerService(CustomerRepository(async_session))` — Rule 127 says services should return domain objects and routers serialize; the service is constructed in the test with a real repository, but the test exercises the service's own validation path, which is fine, yet the manual repo construction here may diverge from how the router wires it.

**Suggestion:** Add a comment explaining why the service is constructed in the test with an inline repo (or use a small factory fixture) so future maintainers don't think this is the production wiring.

tests/integration/test_engagement_webhook_integration.py
```diff
@@ -0,0 +1,322 @@
...
+        score_factors=score_factors,
+    )
+    async_session.add(customer)
+    await async_session.flush()
+    return customer.id
+
+
+async def _seed_tenant(async_session: AsyncSession, tenant_id: int) -> None:
+    """Insert a tenant row so FK references are satisfied. Idempotent — no-op if already present."""
+    from sqlalchemy import select
```
yihua-zhuo: **INFO** (suggestion)

**Issue:** Duplicate import: `from sqlalchemy import select` is already imported at the top of the file (line 13) and re-imported inside `_seed_tenant` on line 57.

**Suggestion:** Remove the redundant `from sqlalchemy import select` on line 57 — the top-level import is sufficient.

tests/integration/test_engagement_webhook_integration.py
```diff
@@ -0,0 +1,322 @@
...
+        # Highest first, NULLs treated as 0 (last in DESC order)
+        ids_in_order = [c.id for c in items]
+        assert ids_in_order.index(high_id) < ids_in_order.index(mid_id)
+        assert ids_in_order.index(mid_id) < ids_in_order.index(low_id)
+
+    async def test_default_ordering_unchanged_without_order_by_score(
+        self, db_schema, tenant_id, async_session
+    ):
+        """Without order_by_score the default created_at DESC ordering is preserved."""
+        from datetime import datetime, timedelta, timezone
```
yihua-zhuo: **INFO** (suggestion)

**Issue:** `from datetime import datetime, timedelta, timezone` is re-imported inside `test_default_ordering_unchanged_without_order_by_score` even though `datetime, timezone` are already imported at the module level (line 10).

**Suggestion:** Remove the redundant import; the module-level import on line 10 plus `from datetime import timedelta` (or add `timedelta` to the top-level import) is sufficient.

tests/integration/test_engagement_webhook_integration.py
```diff
@@ -0,0 +1,322 @@
+"""Integration tests for the engagement webhook and lead-tier filter / score sort.
+
+Run against a real PostgreSQL database (Supabase via DATABASE_URL env var):
+    DATABASE_URL="postgresql+asyncpg://..." pytest tests/integration/test_engagement_webhook_integration.py -v
+"""
+
+from __future__ import annotations
+
+import uuid
+from datetime import datetime, timezone
```
yihua-zhuo: **INFO** (suggestion)

**Issue:** The `timezone` import from line 10 is used only inside the test method that re-imports `datetime` locally — if the test-local import were cleaned up, this import would still be needed, so it is fine. No action required, but the redundancy above creates confusion about which `datetime` is in scope.

**Suggestion:** Consolidate all `datetime` imports to the top of the file.

tests/integration/test_engagement_webhook_integration.py
```diff
@@ -0,0 +1,322 @@
...
+#  Engagement webhook — service-level integration
+# ──────────────────────────────────────────────────────────────────────────────────────
+
+
+@pytest.mark.integration
+@pytest.mark.asyncio
+class TestEngagementWebhookIntegration:
+    """POST /events/engagement persists rows and triggers score recalculation."""
+
+    async def test_engagement_event_persisted_with_correct_fields(
```
yihua-zhuo, line 80: **INFO** (suggestion)

**Issue:** The assertion `assert event.event_metadata == metadata` relies on the ORM attribute name `event_metadata` — the model defines this as `event_metadata` (confirmed), but the test does not document the column name, which could break silently if the model is renamed.

**Suggestion:** Add a brief comment naming the source column (`EngagementEventModel.event_metadata`) so future refactors catch the dependency.

tests/integration/test_engagement_webhook_integration.py
```diff
@@ -0,0 +1,322 @@
...
+        result = await score_svc.calculate_score(
+            customer_id=customer_id, tenant_id=tenant_id
+        )
+
+        assert result.score >= 80
+        assert result.tier.value == "A"
+        assert result.tier_label == "A"
+
+    async def test_cross_tenant_isolation_event_not_visible(
+        self, db_schema, tenant_id, tenant_id_2, async_session
```
yihua-zhuo, line 156: **WARNING** (suggestion)

**Issue:** Cross-tenant test seeds only one customer for tenant A and does not seed a customer belonging to tenant B — the test passes because `ScoreService.calculate_score` raises `NotFoundException` when the customer isn't found in tenant B's scope, but the test name implies an event-visibility check (Rule 126: cross-tenant negative tests are required).

**Suggestion:** Also seed a customer owned by tenant_id_2 and assert that querying events for tenant A does not return tenant B's rows, to fully cover Rule 126's cross-tenant negative test requirement.

tests/unit/test_events_webhook.py
```diff
@@ -0,0 +1,194 @@
...
+
+        response = client.post(
+            "/api/v1/events/engagement",
+            json={"customer_id": 1, "event_type": "email_open"},
+        )
+
+        assert response.status_code == 200
+        assert call_order == ["record", "calc"]
+
+    def test_invalid_event_type_returns_422(self, app_with_overrides):
```
yihua-zhuo, line 145: **WARNING** (suggestion)

**Issue:** Test `test_invalid_event_type_returns_422` asserts Pydantic validation returns 422 — Rule 92 says proper HTTP status codes must be returned and Rule 93 says request/response schemas must be validated. The test verifies the status code but does not assert that the response body contains a structured error envelope (`success: False, message: ...`).

**Suggestion:** Add `body = response.json()` and assert `body["success"] is False` and that `body` contains a `message` or `detail` field, matching the global error envelope shape.

tests/unit/test_events_webhook.py
```diff
@@ -0,0 +1,194 @@
...
+
+        response = client.post(
+            "/api/v1/events/engagement",
+            json={"event_type": "email_open"},
+        )
+
+        assert response.status_code == 422, response.text
+        mock_event.record_engagement_event.assert_not_called()
+
+    def test_negative_customer_id_returns_422(self, app_with_overrides):
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** `test_negative_customer_id_returns_422` only tests `customer_id=0` — the docstring says `<= 0` but the boundary case `customer_id=-1` is not tested. Rule 64 says edge cases must be tested.

**Suggestion:** Parametrize over `[0, -1, -999]` or add a second test case for `customer_id=-1` to fully cover the `<= 0` boundary.

tests/unit/test_events_webhook.py
```diff
@@ -0,0 +1,194 @@
...
+@pytest.fixture
+def app_with_overrides(monkeypatch, mock_db_session):
+    """Build a FastAPI app with the events router and override auth/db deps."""
+    mock_event = MagicMock()
+    mock_event.record_engagement_event = AsyncMock()
+
+    mock_score = MagicMock()
+    mock_score.calculate_score = AsyncMock(return_value=_make_score_result())
+
+    monkeypatch.setattr("api.routers.events.EventService", lambda session: mock_event)
```
yihua-zhuo, line 53: **INFO** (suggestion)

**Issue:** `monkeypatch.setattr("api.routers.events.EventService", lambda session: mock_event)` — the test patches the service classes at the router module level. This works but bypasses the service constructor's `AsyncSession` type contract. Rule 135 prefers composable mock state; this is a thin lambda that is acceptable for router-level tests.

**Suggestion:** Add a comment that this is a router-level delegation test and that the service's own SQL is covered by integration tests, so the lambda is intentional.

tests/unit/test_events_webhook.py
```diff
@@ -0,0 +1,194 @@
...
+    app.dependency_overrides[get_db] = lambda: mock_db_session
+
+    @app.exception_handler(AppException)
+    async def app_exception_handler(request: Request, exc: AppException) -> JSONResponse:
+        return JSONResponse(
+            status_code=exc.status_code,
+            content={"success": False, "message": exc.detail, "code": exc.code},
+        )
+
+    client = TestClient(app, raise_server_exceptions=False)
```
yihua-zhuo, line 68: **INFO** (suggestion)

**Issue:** `TestClient(app, raise_server_exceptions=False)` swallows server exceptions silently — this can mask 500 errors that should surface during test development. Rule 45 says failure paths must be tested.

**Suggestion:** Add at least one test that explicitly triggers a 500 (e.g., service raises an unexpected exception) and asserts the global handler returns a structured 500, to verify that `raise_server_exceptions=False` doesn't hide real bugs.

tests/unit/test_events_webhook.py
```diff
@@ -0,0 +1,194 @@
...
+
+    monkeypatch.setattr("api.routers.events.EventService", lambda session: mock_event)
+    monkeypatch.setattr("api.routers.events.ScoreService", lambda session: mock_score)
+
+    app = FastAPI()
+    app.include_router(router)
+    app.dependency_overrides[require_auth] = lambda: _make_auth_ctx()
+    app.dependency_overrides[get_db] = lambda: mock_db_session
+
+    @app.exception_handler(AppException)
```
yihua-zhuo, line 61: **INFO** (suggestion)

**Issue:** The `app_exception_handler` is registered locally on the test app, but the production app's global handler (in `src/main.py`) may return a different envelope shape. If the handler diverges, these tests will pass but production will differ.

**Suggestion:** Import the production handler from `src/main.py` and reuse it in the test app instead of redefining it, so the test exercises the actual error envelope.

tests/unit/test_events_webhook.py
```diff
@@ -0,0 +1,194 @@
+"""Unit tests for src/api/routers/events.py — engagement webhook endpoint tests."""
```
yihua-zhuo: **INFO** (suggestion)

**Issue:** The test file's module docstring says 'engagement webhook endpoint tests' — the file also contains `TestEventServiceValidation` which tests the standalone service, not the router endpoint. The docstring scope is narrower than the file contents.

**Suggestion:** Widen the docstring to 'Tests for the events webhook router and EventService validation path.'

tests/unit/test_events_webhook.py
```diff
@@ -0,0 +1,194 @@
...
+
+class TestEventServiceValidation:
+    """Unit tests for the standalone EventService.record_engagement_event validation path."""
+
+    @pytest.mark.asyncio
+    async def test_record_engagement_event_invalid_type_raises(self):
+        """EventService raises ValidationException for event_type not in the allowlist."""
+        from services.event_service import EventService
+
+        svc = EventService(MagicMock())
```
yihua-zhuo, line 197: **INFO** (suggestion)

**Issue:** `test_record_engagement_event_invalid_type_raises` constructs `EventService(MagicMock())` — the service's `__init__` types session as `AsyncSession` (no default), but a raw `MagicMock` is passed. Rule 27 says type hints should be meaningful, and Rule 66 says mocking should not be excessive. A raw `MagicMock` for the session is acceptable here since the validation path doesn't touch the DB, but it should be documented.

**Suggestion:** Add a comment: '# validation path does not touch the DB; MagicMock is safe here.'

yihua-zhuo:

**Status:** `fail`
**Inline comments posted:** 29

**Summary:** Reviewed 11 file(s) between origin/master and origin/auto/issue-586.

`src/db/repositories/customer.py` L349: 
**CRITICAL** (critical)

**Issue:** `bulk_recycle` records `previous_owner_id` from `lead.owner_id` AFTER the per-row UPDATE and `refresh()`, so the value is always the new owner (0), never the previous owner. This silently corrupts the recycle history.

**Suggestion:** Capture the original owner before the UPDATE: `prev_owner = lead.owner_id`, then use `prev_owner` in the history dict. E.g. `previous_owner_id=prev_owner`. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`src/api/routers/customers.py` L226: 
**WARNING** (suggestion)

**Issue:** No tenant_id is passed to the enrichment_status helper in the else branch of the list query path — but the helper does receive ctx.tenant_id on line 237, so this is fine. However, the list_customers query itself has no explicit tenant_id filter visible at the router level (it is in the service/repository layer).

**Suggestion:** No change needed at router level — tenant_id is threaded to the service which is responsible for the WHERE clause per Rule 126. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`src/db/repositories/customer.py` L328: 
**WARNING** (suggestion)

**Issue:** N+1 query pattern: each lead gets its own UPDATE + flush + refresh round-trip. For a large batch this is O(N) DB calls (Rule 85).

**Suggestion:** Perform one batch UPDATE with `CASE WHEN id=X THEN recycle_count+1 ...`, or build history entries from the initial SELECT's in-memory values, then a single flush. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`src/db/repositories/customer.py` L176: 
**INFO** (suggestion)

**Issue:** `import logging` is placed inside `count_by_status` instead of at module top (Rule 15/21 — avoid unnecessary local imports).

**Suggestion:** Move `import logging` to the top of the file with the other imports. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`src/db/repositories/customer.py` L211: 
**INFO** (suggestion)

**Issue:** Magic number `100` as the search result cap in `search_customers` (Rule 19 — extract named constants).

**Suggestion:** Define `MAX_SEARCH_RESULTS = 100` at module or class level and use it: `.limit(MAX_SEARCH_RESULTS)`. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`src/services/customer_service.py` L23: 
**WARNING** (suggestion)

**Issue:** Return type annotations use `Any` (e.g. `-> Any`, `-> Any | None`) on every public method, violating Rule 27 (meaningful type hints) and Rule 127 (services return domain objects).

**Suggestion:** Replace with concrete types: `-> CustomerModel`, `-> tuple[list[CustomerModel], int]`, `-> dict[CustomerStatus, int]`, etc. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`src/services/customer_service.py` L32: 
**INFO** (suggestion)

**Issue:** Single-letter variable name `d` is cryptic (Rule 11 — meaningful names).

**Suggestion:** Rename to `data_dict` or `payload`. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`src/services/customer_service.py` L22: 
**WARNING** (suggestion)

**Issue:** String-quoted forward reference `"LeadRoutingService | None"` suggests a circular-import hazard (Rule 3 — no circular imports).

**Suggestion:** Break the cycle by extracting a `LeadRouter` Protocol in a shared module, or inject the dependency via a factory rather than importing `LeadRoutingService` directly into the service module. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`tests/unit/test_customer_service.py` L3: 
**WARNING** (suggestion)

**Issue:** Rule 135: `_make_mock_customer_repo` uses raw `MagicMock`/`AsyncMock` to build a repository mock. The project has a composable handler pattern (`make_mock_session` + domain handlers) that should be preferred for service tests; the file even imports `make_mock_session` but only uses it in one test class.

**Suggestion:** Replace `_make_mock_customer_repo` with a domain handler from `tests/unit/domain_handlers/customer.py` or create a file-local handler for the service's repo interface. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`tests/unit/test_customer_service.py` L14: 
**WARNING** (suggestion)

**Issue:** Rule 134/133: `MagicMock()` repository mock has no tenant isolation semantics — any call to `create`, `list_customers`, etc. succeeds regardless of `tenant_id`. This hides cross-tenant bugs and violates the unit-mock-tenant-isolation guideline.

**Suggestion:** Use a domain handler that checks `params['tenant_id']` on every read/write, matching the SQL bind parameters used by the production code. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`tests/unit/test_customer_service.py` L150: 
**INFO** (suggestion)

**Issue:** Rule 135: `mock_db_session` fixture is included in the `test_create_customer_accepts_dict` and `test_create_customer_accepts_dto` test signatures but never used — these tests only need the mock repo, not a mock session.

**Suggestion:** Remove the `mock_db_session` parameter from tests that only need `mock_customer_repo`. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`tests/unit/test_customer_service.py` L302: 
**INFO** (suggestion)

**Issue:** Rule 135: `test_count_by_status_basic` creates a separate raw `session = MagicMock()` and `mock_result` then *ignores* them, overwriting `mock_customer_repo.count_by_status` with a return value. The unused mock session adds noise and confusion.

**Suggestion:** Remove the unused `session`/`mock_result` setup; use only the `mock_customer_repo` fixture. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`tests/unit/test_customer_service.py` L69: 
**INFO** (suggestion)

**Issue:** Rule 136: The `test_all_fields` test passes `status="customer"` as a raw string instead of `CustomerStatus.CUSTOMER`. This doesn't test enum validation — a Pydantic DTO accepting any string is potentially a validation gap.

**Suggestion:** Add a test asserting that an invalid status value (e.g. `"invalid_status"`) is rejected by the DTO. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`tests/unit/test_customer_service.py` L128: 
**INFO** (suggestion)

**Issue:** Rule 136: `from_dict` only tests for `name` and `email` missing — it should also assert that extra/unknown fields are rejected or ignored per the declared schema.

**Suggestion:** Add a test passing an unknown key to `from_dict` and asserting the behavior matches the schema's `model_config`. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_

alembic/versions/add_engagement_events_table.py
```diff
@@ -0,0 +1,88 @@
+"""add_engagement_events_table
```
yihua-zhuo, line 1: **CRITICAL** (critical)

**Issue:** Migration filename does not follow the project convention of `<revision_id>_<description>.py` (e.g. `e8ad92ca62b3_add_lead_score_columns.py`). All other files in alembic/versions/ use a hex revision prefix; this one uses `add_engagement_events_table.py` which makes it unsearchable and easy to collide on.

**Suggestion:** Rename to `a1b2c3d4e5f6_add_engagement_events_table.py` to match the existing naming pattern.

.plans/issue-586.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Implementation Steps
+
+1. **Extend `CustomerRepository.list_customers`** in `src/db/repositories/customer.py` (L38-64): add `lead_tier: str | None = None` and `order_by_score: bool = False` parameters; in the conditions list, add `CustomerModel.tier == lead_tier` when `lead_tier` is not None; for ordering, use `func.coalesce(CustomerModel.score, 0).desc()` when `order_by_score` is True, otherwise retain the existing `CustomerModel.created_at.desc()`.
+
+2. **Extend `CustomerService.list_customers`** in `src/services/customer_service.py` (L48-63): add the same two parameters; validate `lead_tier` against the set `{"hot", "warm", "cold"}` and raise `ValidationException` for unknown values; forward both parameters to `self.repository.list_customers`.
+
+3. **Update `GET /customers/` in `src/api/routers/customers.py`** (L213-244): add `lead_tier: str | None = Query(None, description="Filter by lead tier: hot, warm, cold")` and `order_by_score: bool = Query(False, description="Auto-rank by score descending")` parameters; pass both to `service.list_customers`. No changes to the enrichment enrichment loop or response envelope.
```
yihua-zhuo, line 34: **WARNING** (suggestion)

**Issue:** Step 3 says 'lead_tier' uses pattern='^(hot|warm|cold)$' but Risk #77 flags a mismatch with the stored ScoreTier enum values (A/B/C/D); the plan never resolves this gap between the dev-plan and the accepted tier format.

**Suggestion:** Add an explicit decision on tier value mapping (hot→A, warm→B, cold→C) and update Step 3 and the service layer description accordingly.

.plans/issue-586.md
```diff
@@ -0,0 +1,79 @@
...
+
+10. **Create integration tests** in `tests/integration/test_engagement_webhook_integration.py`: seed a customer in a tenant, POST to `/api/v1/events/engagement` with `event_type="email_open"`, then query the customer row and assert `score` and `tier` are updated. Add a cross-tenant test: POST with `tenant_id=A` customer, query from `tenant_id=B` → should not see the event. Add a `lead_tier` filter test: seed two customers with different tier values, GET `/api/v1/customers/?lead_tier=hot` returns only the hot one. Add an `order_by_score` test: seed three customers with scores 10/50/90, GET `/api/v1/customers/?order_by_score=true` returns them in descending order.
+
+11. **Run the full verification pipeline**: `ruff check src/ && ruff format --check src/` must exit 0; `PYTHONPATH=src pytest tests/unit/ -v` must pass; `PYTHONPATH=src pytest tests/integration/test_engagement_webhook_integration.py -v` must pass; `alembic upgrade head && alembic downgrade -1 && alembic upgrade head` must complete without error.
```
yihua-zhuo, line 50: **WARNING** (suggestion)

**Issue:** Acceptance criterion §6 E2E curl examples reference an endpoint shape that may not match implementation: the plan says the response uses 'tier' key but ScoreResult.tier_label is an enum-derived string.

**Suggestion:** Pin the exact response field names (score, tier) and show the actual JSON produced by ScoreService.calculate_score via ScoreResponse.

.plans/issue-586.md
```diff
@@ -0,0 +1,79 @@
...
+  - `test_engagement_webhook_integration.py` — new file; full-DB tests for webhook trigger → score update, cross-tenant isolation, `lead_tier` filter accuracy, and `order_by_score` sort order
+- Dev-plan verification:
+  - §6 acceptance #1: `ruff check src/api/routers/customers.py src/api/routers/events.py src/services/customer_service.py src/services/event_service.py` → 0 errors
+  - §6 acceptance #2: `PYTHONPATH=src pytest tests/unit/test_customer_service.py -v` → 4+ new cases passed
+  - §6 acceptance #3: `PYTHONPATH=src pytest tests/unit/test_events_webhook.py -v` → all passed
+  - §6 acceptance #4: `PYTHONPATH=src pytest tests/integration/test_engagement_webhook_integration.py -v` → all passed
+  - §6 acceptance #5: `alembic upgrade head && alembic downgrade -1 && alembic upgrade head` → three exit 0 (covers Step 5 migration)
+  - §6 E2E webhook: `curl -X POST /api/v1/events/engagement` with valid body → returns `{"success": true, "data": {"customer_id": N, "score": M, "tier": "..."}}`
+  - §6 E2E list: `curl /api/v1/customers/?lead_tier=hot&order_by_score=true` → returns filtered, score-sorted list
+
```
yihua-zhuo, line 66: **INFO** (suggestion)

**Issue:** Plan lists only 4 pass-through unit test cases for list_customers, but Rule 140 requires the plan to be checked for stale test counts; no negative or NULL-tier case is enumerated.

**Suggestion:** Add a test case for lead_tier=None (no filter) and order_by_score=False (default ordering preserved).

alembic/versions/add_engagement_events_table.py
```diff
@@ -0,0 +1,88 @@
...
+revision: str = "a1b2c3d4e5f6"
+down_revision: str | None = "e8ad92ca62b3"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    op.create_table(
+        "engagement_events",
+        sa.Column("id", sa.Integer(), autoincrement=True, nullable=False),
```
yihua-zhuo, line 29: **INFO** (suggestion)

**Issue:** tenant_id column has a ForeignKey but no index declared on the column itself — the separate ix_engagement_events_tenant_id index on line 57 compensates, but the column definition lacks `index=True` which is inconsistent with the customer_id column definition that also relies on a separate index.

**Suggestion:** This is acceptable as-is since explicit indexes are created below, but document the intent or add `index=True` for consistency with ORM model definitions.

alembic/versions/add_engagement_events_table.py
```diff
@@ -0,0 +1,88 @@
...
+    )
+    op.create_index(
+        "ix_engagement_events_tenant_event_type",
+        "engagement_events",
+        ["tenant_id", "event_type"],
+        unique=False,
+    )
+
+
+def downgrade() -> None:
```
yihua-zhuo, line 83: **INFO** (suggestion)

**Issue:** downgrade() drops indexes and table in correct reverse order — reversible migration passes Rule 129 review.

src/api/routers/customers.py
```diff
@@ -216,6 +216,8 @@ async def list_customers(
     page_size: int = Query(20, ge=1, le=100),
     status: str | None = None,
     owner_id: int | None = Query(None, ge=0),
+    lead_tier: str | None = Query(None, pattern="^(hot|warm|cold)$", description="Filter by lead tier: hot, warm, cold"),
```
yihua-zhuo, line 219: **WARNING** (suggestion)

**Issue:** lead_tier query parameter uses pattern '^(hot|warm|cold)$' but the actual CustomerModel.tier column stores ScoreTier enum letter values (A/B/C/D) per the migration e8ad92ca62b3 — the filter will silently return zero results because no row will ever match 'hot'. The plan's Risk #77 flagged this gap and it was not resolved.

**Suggestion:** Either change the pattern to '^(A|B|C|D)$' or add a mapping layer in the service that translates hot→A, warm→B, cold→C before querying.

yihua-zhuo, line 219: **INFO** (suggestion)

**Issue:** lead_tier parameter description says 'Filter by lead tier: hot, warm, cold' — if the actual stored values are A/B/C/D, this documentation is misleading for API consumers.

**Suggestion:** Align the description with the accepted values, or document the hot/warm/cold→A/B/C/D mapping explicitly.

src/api/routers/events.py
```diff
@@ -0,0 +1,51 @@
...
+router = APIRouter(prefix="/api/v1/events", tags=["events"])
+
+
+class EngagementEventRequest(BaseModel):
+    customer_id: int = Field(..., gt=0)
+    event_type: str = Field(..., pattern="^(email_open|website_visit)$")
+    event_metadata: dict | None = None
+
+
+@router.post("/engagement", status_code=200)
```
yihua-zhuo, line 23: **WARNING** (suggestion)

**Issue:** No idempotency key or dedup window — a webhook retry from an upstream system (email tracker, analytics) will create duplicate engagement_events rows and trigger duplicate score recalculations, violating Rule 96 (idempotency for webhooks).

**Suggestion:** Accept an optional idempotency_key field and check for existing events with the same key before inserting, or use the plan's metadata field for this purpose.

yihua-zhuo, line 23: **WARNING** (suggestion)

**Issue:** Webhook authenticates via Depends(require_auth) (user JWT) but the endpoint is a server-to-server webhook for external systems (email trackers, website analytics) — they cannot present user credentials. Rule 56/99 require appropriate auth for the endpoint's actual use case.

**Suggestion:** Use a separate dependency like a shared-secret header check (e.g., X-Webhook-Secret) for the engagement endpoint, or document why user-JWT is acceptable for this integration.

src/db/models/engagement.py
```diff
@@ -0,0 +1,28 @@
...
+    id: Mapped[int] = mapped_column(Integer, primary_key=True, autoincrement=True)
+    tenant_id: Mapped[int] = mapped_column(
+        Integer, ForeignKey("tenants.id", ondelete="CASCADE"), nullable=False, index=True
+    )
+    customer_id: Mapped[int] = mapped_column(
+        Integer, ForeignKey("customers.id", ondelete="CASCADE"), nullable=False, index=True
+    )
+    event_type: Mapped[str] = mapped_column(String(50), nullable=False)
+    event_metadata: Mapped[dict | None] = mapped_column(JSONB, nullable=True, default=dict)
+    created_at: Mapped[datetime] = mapped_column(DateTime(timezone=True), server_default=func.now(), nullable=False)
```
yihua-zhuo, line 28: **INFO** (suggestion)

**Issue:** Model has no `updated_at` field; if append-only by design, this should be documented as required by Rule 139 (lifecycle consistency).

**Suggestion:** Add a comment above `created_at`: `# Append-only: engagement events are never modified after insert.`

src/db/models/engagement.py
```diff
@@ -0,0 +1,28 @@
...
+from db.base import Base
+
+
+class EngagementEventModel(Base):
+    """Engagement event entity mapped to the ``engagement_events`` table."""
+
+    __tablename__ = "engagement_events"
+
+    id: Mapped[int] = mapped_column(Integer, primary_key=True, autoincrement=True)
+    tenant_id: Mapped[int] = mapped_column(
```
yihua-zhuo, line 20: **INFO** (suggestion)

**Issue:** Separate single-column indexes on `tenant_id` and `customer_id` instead of a composite index for the common `(tenant_id, customer_id)` access pattern (Rule 84).

**Suggestion:** Replace individual `index=True` flags with a composite `Index('ix_engagement_tenant_customer', 'tenant_id', 'customer_id')` in `__table_args__`.

src/services/customer_service.py
```diff
@@ -45,21 +45,39 @@ async def create_customer(
 
         return customer
 
+    # Maps the public lead_tier filter values (hot/warm/cold) to the stored
+    # ScoreTier letter values (A/B/C) that the customer.tier column carries.
+    # Tier "D" customers are below the cold threshold and excluded by design.
+    LEAD_TIER_TO_STORED: dict[str, str] = {
```
yihua-zhuo, line 51: **INFO** (suggestion)

**Issue:** `LEAD_TIER_TO_STORED` is defined between `create_customer` and `list_customers` methods, breaking class-level organization and single-responsibility layout (Rule 4/15).

**Suggestion:** Move it next to `VALID_STATUSES` at the top of the class, with a single line of context.

src/services/event_service.py
```diff
@@ -0,0 +1,63 @@
...
+from db.models.engagement import EngagementEventModel
+from pkg.errors.app_exceptions import NotFoundException, ValidationException
+
+VALID_EVENT_TYPES = {"email_open", "website_visit"}
+
+
+class EventService:
+    """Business logic for engagement events — persists EngagementEventModel rows."""
+
+    VALID_EVENT_TYPES = VALID_EVENT_TYPES
```
yihua-zhuo, line 20: **INFO** (suggestion)

**Issue:** `VALID_EVENT_TYPES = VALID_EVENT_TYPES` inside the class re-binds the module-level constant pointlessly (Rule 15 — unnecessary cleverness).

**Suggestion:** Remove the class-level re-assignment and reference the module constant directly, or define the set once inside the class.

src/services/event_service.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def __init__(self, session: AsyncSession) -> None:
+        self.session = session
+
+    async def record_engagement_event(
+        self,
+        tenant_id: int,
+        customer_id: int,
+        event_type: str,
+        metadata: dict[str, Any] | None = None,
```
yihua-zhuo, line 30: **INFO** (suggestion)

**Issue:** Parameter name `metadata` shadows the `importlib.metadata` standard-library module (Rule 14 — misleading naming).

**Suggestion:** Rename to `event_metadata` or `payload` to match the ORM column name and avoid stdlib shadowing.

src/services/event_service.py
```diff
@@ -0,0 +1,63 @@
...
+        Flushes so the auto-generated id is populated, but does not commit —
+        the router-owned transaction boundary (Rule 121) commits on normal
+        exit. Callers that invoke this service outside a router context must
+        commit the session themselves if they need the row visible to
+        subsequent reads.
+        """
+        if event_type not in self.VALID_EVENT_TYPES:
+            raise ValidationException(f"event_type must be one of: {', '.join(sorted(self.VALID_EVENT_TYPES))}")
+
+        customer_exists = await self.session.execute(
```
yihua-zhuo, line 43: **INFO** (suggestion)

**Issue:** Extra SELECT to confirm customer existence before every INSERT adds a round-trip (Rule 85 — N+1 avoidance). Acceptable for clean 404 semantics but worth a `select 1` with `EXISTS` or a foreign-key catch if volume is high.

**Suggestion:** If performance matters, use `select(1).where(...).exists()` and wrap the INSERT in a try/except for `IntegrityError` as a fallback path.

tests/integration/test_engagement_webhook_integration.py
```diff
@@ -0,0 +1,342 @@
...
+        score=score,
+        tier=tier,
+        score_factors=score_factors,
+    )
+    async_session.add(customer)
+    await async_session.flush()
+    return customer.id
+
+
+async def _seed_tenant(async_session: AsyncSession, tenant_id: int) -> None:
```
yihua-zhuo, line 55: **WARNING** (suggestion)

**Issue:** Rule 125/149: Local `_seed_tenant` helper duplicates a domain-owned `_seed_tenant` fixture already provided by `tests/integration/conftest.py` (line 254). The file-local version should be removed in favor of the shared fixture, and if an async variant is needed it belongs in `tests/integration/domain_fixtures/<domain>.py`.

**Suggestion:** Remove the local `_seed_tenant` function and inject the `_seed_tenant` fixture from `tests/integration/conftest.py` via the test method parameters.

tests/integration/test_engagement_webhook_integration.py
```diff
@@ -0,0 +1,342 @@
...
+from services.customer_service import CustomerService
+from db.repositories.customer import CustomerRepository
+
+
+# ──────────────────────────────────────────────────────────────────────────────────────
+#  Helpers
+# ──────────────────────────────────────────────────────────────────────────────────────
+
+
+async def _seed_customer(
```
yihua-zhuo, line 28: **WARNING** (suggestion)

**Issue:** Rule 125/149: `_seed_customer` helper is a domain-specific seed that belongs in `tests/integration/domain_fixtures/customer.py` (or equivalent domain-owned module), not inline in the test file.

**Suggestion:** Move `_seed_customer` into a domain-owned fixture module; the conftest already establishes this pattern.

tests/integration/test_engagement_webhook_integration.py
```diff
@@ -0,0 +1,342 @@
...
+    """POST /events/engagement persists rows and triggers score recalculation."""
+
+    async def test_engagement_event_persisted_with_correct_fields(
+        self, db_schema, tenant_id, async_session
+    ):
+        """EventService writes a row with the expected tenant_id, customer_id, event_type, and metadata."""
+        await _seed_tenant(async_session, tenant_id)
+        customer_id = await _seed_customer(async_session, tenant_id)
+
+        from services.event_service import EventService
```
yihua-zhuo, line 87: **INFO** (suggestion)

**Issue:** Rule 146: `from services.event_service import EventService` inside a test function is an implicit convention that hides import-time side effects and can violate import-safety if the module opens network or starts tasks. Move the import to module top-level.

**Suggestion:** Hoist `from services.event_service import EventService` and the score_service import to the top-level imports block.

tests/integration/test_engagement_webhook_integration.py
```diff
@@ -0,0 +1,342 @@
...
+        metadata = {"campaign": "spring", "source": "newsletter"}
+        event = await svc.record_engagement_event(
+            tenant_id=tenant_id,
+            customer_id=customer_id,
+            event_type="email_open",
+            metadata=metadata,
+        )
+        # EventService uses flush() (Rule 121 — routers own commit).
+        # Flush here so the re-read below sees the row.
+        await async_session.flush()
```
yihua-zhuo, line 99: **INFO** (suggestion)

**Issue:** Rule 121: Manual `await async_session.flush()` in the test is a workaround for the service using flush; the test itself shouldn't need to flush if the service's flush already makes data visible within the transaction. The comment on line 97-98 acknowledges the rule, but the test could be simplified by re-reading after the service's own flush.

**Suggestion:** Rely on the service's internal flush for re-reads; remove the explicit flush call or add a comment explaining why the test needs it.

tests/integration/test_engagement_webhook_integration.py
```diff
@@ -0,0 +1,342 @@
...
+        instance) so the test exercises the service's own validation path
+        without going through the FastAPI dependency injection machinery.
+        """
+        await _seed_tenant(async_session, tenant_id)
+        await _seed_customer(async_session, tenant_id)
+
+        from pkg.errors.app_exceptions import ValidationException
+
+        svc = CustomerService(CustomerRepository(async_session))
+        with pytest.raises(ValidationException, match="lead_tier must be one of"):
```
yihua-zhuo, line 327: **INFO** (suggestion)

**Issue:** Rule 135: The service-level test constructs `CustomerService(CustomerRepository(async_session))` inline. The shared `_seed_tenant` / seed helpers pattern isn't being used; consider extracting a domain fixture for this service construction to avoid duplication across test files.

**Suggestion:** Add a `customer_service` fixture in `tests/integration/domain_fixtures/customer.py` that wires the service with the real repo and session.

tests/unit/test_events_webhook.py
```diff
@@ -0,0 +1,223 @@
...
+    result = MagicMock()
+    result.score = score
+    result.tier = tier
+    result.tier_label = tier.value
+    return result
+
+
+@pytest.fixture
+def mock_db_session():
+    return MagicMock()
```
yihua-zhuo, line 35: **INFO** (suggestion)

**Issue:** Rule 135: The local `mock_db_session` fixture returns a bare `MagicMock()`. Since this is a router-level test using `app.dependency_overrides`, the `mock_db_session` is only consumed by the dependency override — the raw `MagicMock` is acceptable here but should be documented as such.

**Suggestion:** Add a comment noting the session is a stand-in for the dependency override and is not exercised by the test.

tests/unit/test_events_webhook.py
```diff
@@ -0,0 +1,223 @@
...
+    integration tests.
+    """
+    mock_event = MagicMock()
+    mock_event.record_engagement_event = AsyncMock()
+
+    mock_score = MagicMock()
+    mock_score.calculate_score = AsyncMock(return_value=_make_score_result())
+
+    monkeypatch.setattr("api.routers.events.EventService", lambda session: mock_event)
+    monkeypatch.setattr("api.routers.events.ScoreService", lambda session: mock_score)
```
yihua-zhuo, line 54: **WARNING** (suggestion)

**Issue:** Rule 127: The test patches `api.routers.events.EventService` and `ScoreService` at the router module level. This is acceptable for router-wiring tests, but the comment on lines 42-46 says 'services' own SQL is covered by integration tests — the test file should not also duplicate validation logic in `TestEventServiceValidation` (lines 188-224) since that's service-layer behavior, not router behavior, and should live in `tests/unit/test_event_service.py`.

**Suggestion:** Move `TestEventServiceValidation` to a dedicated `tests/unit/test_event_service.py`; keep this file focused on router wiring.

tests/unit/test_events_webhook.py
```diff
@@ -0,0 +1,223 @@
...
+        """POST /engagement with an invalid event_type returns 422 (Pydantic validation)."""
+        client, mock_event, mock_score = app_with_overrides
+
+        response = client.post(
+            "/api/v1/events/engagement",
+            json={"customer_id": 1, "event_type": "not_a_real_event"},
+        )
+
+        assert response.status_code == 422, response.text
+        # Pydantic 422 returns FastAPI's default validation error shape with "detail"
```
yihua-zhuo, line 155: **INFO** (suggestion)

**Issue:** Rule 144: The 422 validation test asserts `response.status_code == 422` and checks for `"detail"` or `"message"` keys, but doesn't verify that the response body contains the *specific* validation error (e.g. the field name that failed). Weak validation assertions are a known gap.

**Suggestion:** Assert that the response body includes the specific field name (e.g. `"event_type"` or `"customer_id"`) in the validation error detail.

tests/unit/test_events_webhook.py
```diff
@@ -0,0 +1,223 @@
+"""Unit tests for src/api/routers/events.py — tests for the events webhook router and EventService validation path."""
+
+from __future__ import annotations
+
+from unittest.mock import AsyncMock, MagicMock
+
+import pytest
+from fastapi import FastAPI
+from fastapi.testclient import TestClient
```
yihua-zhuo, line 9: **INFO** (suggestion)

**Issue:** Rule 146: `from fastapi.testclient import TestClient` uses Starlette's synchronous `TestClient` which runs the async app in a separate thread. This is fine for simple tests, but blocks the event loop — for async-heavy endpoints, `httpx.AsyncClient` with `ASGITransport` would be more appropriate. Not critical for this file's tests since the service is fully mocked.

**Suggestion:** Consider using `httpx.AsyncClient` with `ASGITransport(app=app)` for consistency with the rest of the async-first test suite.

PR #806, opened 2026-06-06
Auto: implement issue #573

### LINKED ISSUE #573: Implement ChurnPredictionService with scoring logic
Subtask of #51
Depends on #572

Create `src/services/churn_prediction_service.py`. Accept `session: AsyncSession` in constructor (no default). Add `calculate_score(customer_id, tenant_id)` that reads customer attributes (login_frequency, purchase_recency, support_ticket_count, engagement_score) and computes a 0-100 score using a weighted formula. Compute tier from score thresholds (high>70, medium 40-70, low<40), extract top 3 risk factors with weights, and generate recommended_actions list. Raise `NotFoundException` if customer not found. Return a domain object (not dict). Write unit tests using MockState and make_customer_handler.

Dev-Plan: docs/dev-plan/60-analytics/0573-implement-churnpredictionservice-with-scoring-logic.md

yihua-zhuo:
Closes #573

Plan: `.plans/issue-573.md`

1. **Plan** — Claude wrote a detailed implementation plan to `.plans/issue-573.md`.

Please review before merging.

## Summary by CodeRabbit

* **Documentation**
  * Added planning documentation for churn prediction implementation.

* **New Features**
  * Added churn prediction capability that calculates customer risk scores on a 0–100 scale.
  * Provides risk tier classifications (high/medium/low) and identifies top 3 risk factors.
  * Generates recommended retention actions based on customer behavior metrics.

### DIFF
--- a/.plans/issue-573.md
+++ b/.plans/issue-573.md
@@ -1,7 +1,7 @@
 # Implementation Plan — Issue #573
 
 ## Goal
-Create a new `ChurnPredictionService` class in `src/services/churn_prediction_service.py` that computes a real-time 0-100 churn score from four customer dimensions (login_frequency, purchase_recency, support_ticket_count, engagement_score) using a weighted formula. The service returns a domain object (`ChurnPrediction` dataclass) with score, tier, top-3 risk factors, and recommended actions — no DB writes. This fills a gap left by the existing DB-backed `churn_prediction.py`, which will be complemented by this rule-based entry point. The API router and downstream features (#672, #673, #674) depend on this service being available.
+Create a new `ChurnPredictionService` class in `src/services/churn_prediction_service.py` that computes a real-time 0-100 churn score from four customer dimensions (login_frequency, purchase_recency, support_ticket_count, engagement_score) using a weighted formula. The service returns a domain object (`ChurnPrediction` dataclass) with score, tier, top-3 risk factors, and recommended actions — no DB writes (read-only queries against activity, opportunity, and ticket tables). This fills a gap left by the existing DB-backed `churn_prediction.py`, which will be complemented by this rule-based entry point. The API router and downstream features (#672, #673, #674) depend on this service being available.
 
 ## Source Contract
 Dev-plan target: `/home/runner/work/agent-job/agent-job/docs/dev-plan/60-analytics/0573-implement-churnpredictionservice-with-scoring-logic.md`
@@ -64,7 +64,7 @@ Reading order followed:
    - **Test 4 — Tier boundary with seeded metrics**: Use a `mock_db_session` with specific `state.activities` entries to exercise different score ranges and verify `tier` computation logic (note: opportunity/ticket mock data comes from hardcoded handlers, so assertions should validate the tier mapping contract is wired correctly, not exact scores).
 
 ## Test Plan
-- **Unit tests in `tests/unit/`**: `test_churn_prediction_service.py` — covers happy path (customer exists, all dimensions return data), error path (customer not found → `NotFoundException`), and structural validation (returns `ChurnPrediction` dataclass with correct fields, score in 0-100, valid tier, 3 risk factors, non-empty recommendations).
+- **Unit tests in `tests/unit/`**: `test_churn_prediction_service.py` — covers happy path (customer exists, all dimensions return data), error path (customer not found → `NotFoundException`; wrong tenant also raises `NotFoundException`, added during implementation to verify the tenant predicate in `_fetch_raw_metrics`), and structural validation (returns `ChurnPrediction` dataclass with correct fields, score in 0-100, valid tier, 3 risk factors, non-empty recommendations).
 - **Integration tests in `tests/integration/`**: None — the dev-plan §1.3 explicitly excludes DB writes, and all DB reads are exercised through unit-test mocks. No integration test needed.
 - **Dev-plan verification** (target-board §6):
   - `ruff check src/services/churn_prediction_service.py` → 0 errors (Step 1–4 completion gate)
--- a/.plans/issue-573.md
+++ b/.plans/issue-573.md
@@ -1,5 +1,3 @@
-I now have all the information needed. The dev-plan's test approach using `make_count_handler` will work for activity counts (via `make_activity_handler`), but for tickets and opportunities the mocks return hardcoded values. I need to account for this in the test plan. The test assertions should focus on structural correctness (score in 0-100, valid tier, 3 risk factors, list of actions) rather than precise numeric values.
-
 # Implementation Plan — Issue #573
 
 ## Goal
@@ -31,22 +29,22 @@ Reading order followed:
    - **login_frequency**: `SELECT count(ActivityModel.id) WHERE tenant_id=? AND customer_id=? AND created_at >= now - 30d`. Fallback to 0 if no activity.
    - **purchase_recency**: `SELECT max(OpportunityModel.created_at) WHERE tenant_id=? AND customer_id=? AND stage='won'`. Compute `days_since`; normalize tz-naive to UTC. Fallback to 90 days if no won opportunity.
    - **support_ticket_count**: `SELECT count(TicketModel.id) WHERE tenant_id=? AND customer_id=? AND status IN ('open','pending')`. Fallback to 0.
-   - **engagement_score_raw**: Reuse login_count as proxy (activity count, to be normalized later).
+   - **engagement_score_raw**: `SELECT count(DISTINCT ActivityModel.type) WHERE tenant_id=? AND customer_id=? AND created_at >= now - 30d`. Counts distinct activity *types* (call, email, meeting, ...) — a separate signal from login_frequency so the two sub-scores carry independent information.
    - Return `dict` with keys: `login_frequency`, `purchase_recency_days`, `support_ticket_count`, `engagement_score_raw`.
 
 3. **Implement `_normalize_score`, `_compute_tier`, and `_build_top_3_factors` private methods.**
-   - `_normalize_score(name, raw) -> float` maps each dimension's raw value to 0-100:
-     - `login_frequency`: `min(raw / 10 * 100, 100)` — 10+ logins = full score.
-     - `purchase_recency`: `max(0.0, 100.0 - raw)` — 0 days = full score, 90+ days = 0.
-     - `support_ticket_count`: `min(raw / 5 * 100, 100)` — 5+ tickets = full score (this is "health score" — low ticket count = healthy).
-     - `engagement_score`: `min(raw / 30 * 100, 100)` — 30+ activities = full score.
+   - `_normalize_score(name, raw) -> float` maps each dimension's raw value to a true 0-100 health sub-score (higher = healthier, so the weighted sum is a direct sum with no double-inversions):
+     - `login_frequency`: `min(raw / 10 * 100, 100)` — 10+ logins = full health.
+     - `purchase_recency`: `max(0.0, 100.0 - raw)` — 0 days = full health, 90+ days = 0.
+     - `support_ticket_count`: `max(0.0, 100.0 - min(raw / 5 * 100, 100))` — 0 tickets = full health, 5+ tickets = 0.
+     - `engagement_score`: `min(raw / 5 * 100, 100)` — 5+ distinct activity types = full health.
    - `_compute_tier(score) -> str`: `>= 70` → `"high"`, `>= 40` → `"medium"`, `< 40` → `"low"`.
    - `_build_top_3_factors(name_to_score) -> list[ChurnRiskFactor]`: sort dimensions by sub-score descending, return top 3 as `ChurnRiskFactor(name, weight, score, description)` objects.
 
 4. **Implement `async calculate_score(self, customer_id: int, tenant_id: int) -> ChurnPrediction`.**
    - Call `_fetch_raw_metrics` to get raw values.
-   - Normalize each dimension to 0-100 sub-score via `_normalize_score`.
-   - Compute weighted total: `login_score * 0.25 + purchase_score * 0.25 + (100 - support_score) * 0.25 + engagement_score * 0.25`. (Support score is inverted: more tickets → higher churn contribution.)
+   - Normalize each dimension to 0-100 sub-score via `_normalize_score` (each branch returns a true health score — high = healthy, so the weighted sum is direct).
+   - Compute weighted total as `sum(sub * self.WEIGHTS[k] for k, sub in name_to_score.items())` — the WEIGHTS constant is the single source of truth, no hardcoded 0.25 literals.
    - Clamp to [0, 100] with `min(score, 100.0)`, round to 2 decimal places.
    - Determine `tier` via `_compute_tier`.
    - Build `top_3_risk_factors` via `_build_top_3_factors`.
--- a/.plans/issue-573.md
+++ b/.plans/issue-573.md
@@ -1,0 +1,88 @@
+I now have all the information needed. The dev-plan's test approach using `make_count_handler` will work for activity counts (via `make_activity_handler`), but for tickets and opportunities the mocks return hardcoded values. I need to account for this in the test plan. The test assertions should focus on structural correctness (score in 0-100, valid tier, 3 risk factors, list of actions) rather than precise numeric values.
+
+# Implementation Plan — Issue #573
+
+## Goal
+Create a new `ChurnPredictionService` class in `src/services/churn_prediction_service.py` that computes a real-time 0-100 churn score from four customer dimensions (login_frequency, purchase_recency, support_ticket_count, engagement_score) using a weighted formula. The service returns a domain object (`ChurnPrediction` dataclass) with score, tier, top-3 risk factors, and recommended actions — no DB writes. This fills a gap left by the existing DB-backed `churn_prediction.py`, which will be complemented by this rule-based entry point. The API router and downstream features (#672, #673, #674) depend on this service being available.
+
+## Source Contract
+Dev-plan target: `/home/runner/work/agent-job/agent-job/docs/dev-plan/60-analytics/0573-implement-churnpredictionservice-with-scoring-logic.md`
+Template depth: `deep`
+Reading order followed:
+1. `/home/runner/work/agent-job/agent-job/docs/dev-plan/README.md`
+2. `/home/runner/work/agent-job/agent-job/docs/dev-plan/_template-deep.md`
+3. `/home/runner/work/agent-job/agent-job/docs/dev-plan/60-analytics/0573-implement-churnpredictionservice-with-scoring-logic.md`
+
+## Affected Files
+- `src/services/churn_prediction_service.py` — **new file**; defines `ChurnRiskFactor` and `ChurnPrediction` dataclasses + `ChurnPredictionService` class with `calculate_score` method
+- `tests/unit/test_churn_prediction_service.py` — **new file**; unit tests with `MockState` + `make_customer_handler` + `make_count_handler`
+
+## Implementation Steps
+
+1. **Create `src/services/churn_prediction_service.py` with imports, dataclasses, and class skeleton.**
+   - File: `src/services/churn_prediction_service.py`
+   - Imports: `dataclass` from `dataclasses`, `UTC`/`datetime`/`timedelta` from `datetime`, `and_`/`func`/`select` from `sqlalchemy`, `AsyncSession` from `sqlalchemy.ext.asyncio`, `CustomerModel` from `db.models.customer`, `ActivityModel` from `db.models.activity`, `OpportunityModel` from `db.models.opportunity`, `TicketModel` from `db.models.ticket`, `NotFoundException` from `pkg.errors.app_exceptions`.
+   - Define `@dataclass class ChurnRiskFactor` with fields `name: str`, `weight: float`, `score: float`, `description: str`.
+   - Define `@dataclass class ChurnPrediction` with fields `customer_id: int`, `score: float`, `tier: str`, `top_3_risk_factors: list[ChurnRiskFactor]`, `recommended_actions: list[str]`.
+   - Define `class ChurnPredictionService` with `WEIGHTS` class constant (4 dimensions × 0.25 each) and `__init__(self, session: AsyncSession)` that stores `self.session = session`. No default value for `session`.
+
+2. **Implement `async _fetch_raw_metrics(self, customer_id: int, tenant_id: int) -> dict`.**
+   - Query `CustomerModel` by `id` + `tenant_id`; if not found, `raise NotFoundException("Customer")`.
+   - **login_frequency**: `SELECT count(ActivityModel.id) WHERE tenant_id=? AND customer_id=? AND created_at >= now - 30d`. Fallback to 0 if no activity.
+   - **purchase_recency**: `SELECT max(OpportunityModel.created_at) WHERE tenant_id=? AND customer_id=? AND stage='won'`. Compute `days_since`; normalize tz-naive to UTC. Fallback to 90 days if no won opportunity.
+   - **support_ticket_count**: `SELECT count(TicketModel.id) WHERE tenant_id=? AND customer_id=? AND status IN ('open','pending')`. Fallback to 0.
+   - **engagement_score_raw**: Reuse login_count as proxy (activity count, to be normalized later).
+   - Return `dict` with keys: `login_frequency`, `purchase_recency_days`, `support_ticket_count`, `engagement_score_raw`.
+
+3. **Implement `_normalize_score`, `_compute_tier`, and `_build_top_3_factors` private methods.**
+   - `_normalize_score(name, raw) -> float` maps each dimension's raw value to 0-100:
+     - `login_frequency`: `min(raw / 10 * 100, 100)` — 10+ logins = full score.
+     - `purchase_recency`: `max(0.0, 100.0 - raw)` — 0 days = full score, 90+ days = 0.
+     - `support_ticket_count`: `min(raw / 5 * 100, 100)` — 5+ tickets = full score (this is "health score" — low ticket count = healthy).
+     - `engagement_score`: `min(raw / 30 * 100, 100)` — 30+ activities = full score.
+   - `_compute_tier(score) -> str`: `>= 70` → `"high"`, `>= 40` → `"medium"`, `< 40` → `"low"`.
+   - `_build_top_3_factors(name_to_score) -> list[ChurnRiskFactor]`: sort dimensions by sub-score descending, return top 3 as `ChurnRiskFactor(name, weight, score, description)` objects.
+
+4. **Implement `async calculate_score(self, customer_id: int, tenant_id: int) -> ChurnPrediction`.**
+   - Call `_fetch_raw_metrics` to get raw values.
+   - Normalize each dimension to 0-100 sub-score via `_normalize_score`.
+   - Compute weighted total: `login_score * 0.25 + purchase_score * 0.25 + (100 - support_score) * 0.25 + engagement_score * 0.25`. (Support score is inverted: more tickets → higher churn contribution.)
+   - Clamp to [0, 100] with `min(score, 100.0)`, round to 2 decimal places.
+   - Determine `tier` via `_compute_tier`.
+   - Build `top_3_risk_factors` via `_build_top_3_factors`.
+   - Generate `recommended_actions: list[str]` based on raw metrics:
+     - `support_ticket_count > 2` → `"优先处理客户工单，降低流失风险"`
+     - `purchase_recency_days > 60` → `"客户长期无购买记录，触发重新激活营销"`
+     - `login_frequency < 3` → `"客户登录频率低，建议发送个性化内容激活"`
+     - If none triggered → `"客户状态健康，维持常规维护"`
+   - Return `ChurnPrediction` dataclass (not dict, not `ApiResponse`).
+
+5. **Create `tests/unit/test_churn_prediction_service.py` with unit tests.**
+   - Imports: `pytest`, `ChurnPredictionService` + `ChurnPrediction` from `services.churn_prediction_service`, `NotFoundException` from `pkg.errors.app_exceptions`, `make_mock_session` + `MockState` + `make_customer_handler` + `make_count_handler` from `tests.unit.conftest`.
+   - `mock_db_session` fixture: `state = MockState(); state.customers[1] = {"id": 1, "tenant_id": 1, "name": "Test"}; return make_mock_session([make_customer_handler(state), make_count_handler(state)])`.
+   - **Test 1 — Happy path**: `customer_id=1, tenant_id=1` exists. Assert `isinstance(result, ChurnPrediction)`, `0.0 <= result.score <= 100.0`, `result.tier in ("high", "medium", "low")`, `len(result.top_3_risk_factors) == 3`, `isinstance(result.recommended_actions, list) and len(result.recommended_actions) >= 1`.
+   - **Test 2 — Customer not found**: `customer_id=9999, tenant_id=1`. Assert `pytest.raises(NotFoundException)`.
+   - **Test 3 — Returned object is dataclass, not dict**: Assert `result` is `ChurnPrediction` (not `dict`, not `ApiResponse`). Assert all fields are accessible as attributes.
+   - **Test 4 — Tier boundary with seeded metrics**: Use a `mock_db_session` with specific `state.activities` entries to exercise different score ranges and verify `tier` computation logic (note: opportunity/ticket mock data comes from hardcoded handlers, so assertions should validate the tier mapping contract is wired correctly, not exact scores).
+
+## Test Plan
+- **Unit tests in `tests/unit/`**: `test_churn_prediction_service.py` — covers happy path (customer exists, all dimensions return data), error path (customer not found → `NotFoundException`), and structural validation (returns `ChurnPrediction` dataclass with correct fields, score in 0-100, valid tier, 3 risk factors, non-empty recommendations).
+- **Integration tests in `tests/integration/`**: None — the dev-plan §1.3 explicitly excludes DB writes, and all DB reads are exercised through unit-test mocks. No integration test needed.
+- **Dev-plan verification** (target-board §6):
+  - `ruff check src/services/churn_prediction_service.py` → 0 errors (Step 1–4 completion gate)
+  - `ruff format --check src/services/churn_prediction_service.py` → exit 0 (Step 4 completion gate)
+  - `PYTHONPATH=src python -c "from services.churn_prediction_service import ChurnPredictionService, ChurnPrediction; print('ok')"` → `ok` (import sanity check)
+  - `PYTHONPATH=src pytest tests/unit/test_churn_prediction_service.py -v` → `3+ passed` (Step 5 completion gate)
+
+## Acceptance Criteria
+- `src/services/churn_prediction_service.py` exists and defines `ChurnPredictionService`, `ChurnPrediction`, and `ChurnRiskFactor`.
+- `ChurnPredictionService.__init__` accepts `session: AsyncSession` with no default value.
+- `calculate_score(customer_id, tenant_id)` returns a `ChurnPrediction` dataclass (not dict, not `ApiResponse`).
+- The returned object has `score` in `[0.0, 100.0]`, `tier` in `{"high", "medium", "low"}`, exactly 3 `ChurnRiskFactor` items in `top_3_risk_factors`, and a non-empty `list[str]` `recommended_actions`.
+- Tier thresholds: `score >= 70` → `"high"`, `score >= 40` → `"medium"`, `score < 40` → `"low"`.
+- `NotFoundException` is raised when the customer does not exist for the given `tenant_id`.
+- All SQL queries filter by `tenant_id`.
+- `ruff check src/services/churn_prediction_service.py` → 0 errors.
+- `ruff format --check src/services/churn_prediction_service.py` → exit 0.
+- `PYTHONPATH=src pytest tests/unit/test_churn_prediction_service.py -v` → all tests pass (≥ 3 passed).
+- `PYTHONPATH=src python -c "from services.churn_prediction_service import ChurnPredictionService, ChurnPrediction; print('ok')"` → `ok`.
--- a/src/services/churn_prediction_service.py
+++ b/src/services/churn_prediction_service.py
@@ -1,0 +1,195 @@
+"""ChurnPredictionService — rule-based real-time churn scoring (0-100).
+
+Computes a weighted churn score from four customer dimensions and returns
+a ``ChurnPrediction`` dataclass. No DB writes — all queries are read-only.
+"""
+
+from dataclasses import dataclass, field
+from datetime import UTC, datetime, timedelta
+
+from sqlalchemy import and_, func, select
+from sqlalchemy.ext.asyncio import AsyncSession
+
+from db.models.activity import ActivityModel
+from db.models.customer import CustomerModel
+from db.models.opportunity import OpportunityModel
+from db.models.ticket import TicketModel
+from pkg.errors.app_exceptions import NotFoundException
+
+
+@dataclass
+class ChurnRiskFactor:
+    """Single churn risk dimension contributing to the overall score."""
+
+    name: str
+    weight: float
+    score: float
+    description: str
+
+
+@dataclass
+class ChurnPrediction:
+    """Result of a real-time churn scoring calculation."""
+
+    customer_id: int
+    score: float
+    tier: str
+    top_3_risk_factors: list[ChurnRiskFactor] = field(default_factory=list)
+    recommended_actions: list[str] = field(default_factory=list)
+
+
+class ChurnPredictionService:
+    """Rule-based churn scoring — 4-dimension weighted engine, no DB writes."""
+
+    WEIGHTS: dict[str, float] = {
+        "login_frequency": 0.25,
+        "purchase_recency": 0.25,
+        "support_ticket_count": 0.25,
+        "engagement_score": 0.25,
+    }
+
+    _FACTOR_DESCRIPTIONS: dict[str, str] = {
+        "login_frequency": "login frequency in the last 30 days",
+        "purchase_recency": "days since the last won opportunity",
+        "support_ticket_count": "open/pending support ticket burden",
+        "engagement_score": "overall activity-based engagement",
+    }
+
+    def __init__(self, session: AsyncSession):
+        self.session = session
+
+    async def _fetch_raw_metrics(self, customer_id: int, tenant_id: int) -> dict:
+        """Read raw per-dimension counts/dates for a single customer."""
+        result = await self.session.execute(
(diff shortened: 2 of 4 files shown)


tests/unit/test_churn_prediction_service.py
```diff
@@ -0,0 +1,140 @@
...
+    def empty_service(self):
+        """Service with no SQL handlers — every SELECT returns an empty result."""
+        session = make_mock_session([])
+        return ChurnPredictionService(session)
+
+    async def test_raises_not_found(self, empty_service):
+        with pytest.raises(NotFoundException):
+            await empty_service.calculate_score(customer_id=9999, tenant_id=1)
+
+    async def test_wrong_tenant_raises_not_found(self, empty_service):
```
yihua-zhuo: **CRITICAL** (critical)

**Issue:** test_wrong_tenant_raises_not_found uses the same empty_service (no handlers) as test_raises_not_found, so both tests exercise the identical 'no customer in mock state' path — neither proves that the SQL predicate actually filters by tenant_id (Rule 126 cross-tenant negative test required, Rule 134 mock must preserve tenant semantics).

**Suggestion:** Seed a customer with tenant_id=1 in the mock state, then call calculate_score(customer_id=1, tenant_id=999) using a session that has the customer handler wired up — this forces the tenant predicate to be the reason the row is excluded.

.plans/issue-573.md
```diff
@@ -0,0 +1,88 @@
+I now have all the information needed. The dev-plan's test approach using `make_count_handler` will work for activity counts (via `make_activity_handler`), but for tickets and opportunities the mocks return hardcoded values. I need to account for this in the test plan. The test assertions should focus on structural correctness (score in 0-100, valid tier, 3 risk factors, list of actions) rather than precise numeric values.
```
yihua-zhuo: **INFO** (suggestion)

**Issue:** Internal first-person planning narration leaked into the plan document ('I now have all the information needed...').

**Suggestion:** Remove line 1; it is stale internal deliberation that belongs in conversation context, not a plan file (Rule 140).

.plans/issue-573.md
```diff
@@ -0,0 +1,88 @@
...
+
+## Implementation Steps
+
+1. **Create `src/services/churn_prediction_service.py` with imports, dataclasses, and class skeleton.**
+   - File: `src/services/churn_prediction_service.py`
+   - Imports: `dataclass` from `dataclasses`, `UTC`/`datetime`/`timedelta` from `datetime`, `and_`/`func`/`select` from `sqlalchemy`, `AsyncSession` from `sqlalchemy.ext.asyncio`, `CustomerModel` from `db.models.customer`, `ActivityModel` from `db.models.activity`, `OpportunityModel` from `db.models.opportunity`, `TicketModel` from `db.models.ticket`, `NotFoundException` from `pkg.errors.app_exceptions`.
+   - Define `@dataclass class ChurnRiskFactor` with fields `name: str`, `weight: float`, `score: float`, `description: str`.
+   - Define `@dataclass class ChurnPrediction` with fields `customer_id: int`, `score: float`, `tier: str`, `top_3_risk_factors: list[ChurnRiskFactor]`, `recommended_actions: list[str]`.
+   - Define `class ChurnPredictionService` with `WEIGHTS` class constant (4 dimensions × 0.25 each) and `__init__(self, session: AsyncSession)` that stores `self.session = session`. No default value for `session`.
```
yihua-zhuo, line 25: **WARNING** (suggestion)

**Issue:** Plan defines a WEIGHTS class constant (Step 1) but Step 4's formula uses hardcoded 0.25 literals, so WEIGHTS is never referenced — a dead-code pattern propagated to the implementation.

**Suggestion:** In Step 4 write the formula in terms of the WEIGHTS dict, e.g. 'score = sum(sub * WEIGHTS[k] for k, sub in ...)', so the constant is the single source of truth.

.plans/issue-573.md
```diff
@@ -0,0 +1,88 @@
...
+
+2. **Implement `async _fetch_raw_metrics(self, customer_id: int, tenant_id: int) -> dict`.**
+   - Query `CustomerModel` by `id` + `tenant_id`; if not found, `raise NotFoundException("Customer")`.
+   - **login_frequency**: `SELECT count(ActivityModel.id) WHERE tenant_id=? AND customer_id=? AND created_at >= now - 30d`. Fallback to 0 if no activity.
+   - **purchase_recency**: `SELECT max(OpportunityModel.created_at) WHERE tenant_id=? AND customer_id=? AND stage='won'`. Compute `days_since`; normalize tz-naive to UTC. Fallback to 90 days if no won opportunity.
+   - **support_ticket_count**: `SELECT count(TicketModel.id) WHERE tenant_id=? AND customer_id=? AND status IN ('open','pending')`. Fallback to 0.
+   - **engagement_score_raw**: Reuse login_count as proxy (activity count, to be normalized later).
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** engagement_score_raw reuses login_frequency, making two of the four 'dimensions' perfectly correlated and reducing the scoring engine to 3 effective inputs.

**Suggestion:** Use a distinct signal for engagement (e.g. diversity of activity types, or a separate engagement table) so all four weights carry independent information (Rule 111).

src/services/churn_prediction_service.py
```diff
@@ -0,0 +1,195 @@
...
+    score: float
+    tier: str
+    top_3_risk_factors: list[ChurnRiskFactor] = field(default_factory=list)
+    recommended_actions: list[str] = field(default_factory=list)
+
+
+class ChurnPredictionService:
+    """Rule-based churn scoring — 4-dimension weighted engine, no DB writes."""
+
+    WEIGHTS: dict[str, float] = {
```
yihua-zhuo, line 45: **WARNING** (suggestion)

**Issue:** WEIGHTS class constant is defined but never referenced anywhere; calculate_score hardcodes 0.25 in four places instead (Rule 17 dead code, Rule 110 duplication).

**Suggestion:** Replace the hardcoded 0.25 values in the weighted total with self.WEIGHTS references, e.g. login_score * self.WEIGHTS['login_frequency'], or loop over the dict.

src/services/churn_prediction_service.py
```diff
@@ -0,0 +1,195 @@
...
+
+        Higher sub-score = healthier / lower churn risk for that dimension.
+        Support tickets and purchase recency are inverted: more tickets or
+        more days = lower health.
+        """
+        if name == "login_frequency":
+            return min(raw / 10 * 100, 100)
+        if name == "purchase_recency":
+            return max(0.0, 100.0 - raw)
+        if name == "support_ticket_count":
```
yihua-zhuo, line 143: **WARNING** (suggestion)

**Issue:** _normalize_score for 'support_ticket_count' returns min(raw/5*100, 100), which scores MORE tickets as HIGHER health — directly contradicting the function docstring 'Higher sub-score = healthier / lower churn risk' (Rule 27 incomplete/incorrect typing contract, Rule 14 misleading).

**Suggestion:** Make the helper return a consistent health score: support_ticket_count should be max(0.0, 100.0 - min(raw/5*100, 100)) so the docstring and return value agree, and remove the double-inversion at line 168.

src/services/churn_prediction_service.py
```diff
@@ -0,0 +1,195 @@
...
+    async def calculate_score(self, customer_id: int, tenant_id: int) -> ChurnPrediction:
+        """Compute churn score, tier, top factors, and recommended actions."""
+        raw = await self._fetch_raw_metrics(customer_id, tenant_id)
+
+        login_score = self._normalize_score("login_frequency", raw["login_frequency"])
+        purchase_score = self._normalize_score("purchase_recency", raw["purchase_recency_days"])
+        support_score = self._normalize_score("support_ticket_count", raw["support_ticket_count"])
+        engagement_score = self._normalize_score("engagement_score", raw["engagement_score_raw"])
+
+        raw_total = login_score * 0.25 + purchase_score * 0.25 + (100 - support_score) * 0.25 + engagement_score * 0.25
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** Support score is double-inverted: normalized to 'high = many tickets' then subtracted from 100 — the (100 - support_score) expression at the call site exists only to undo the helper's wrong semantics (Rule 15 unnecessary cleverness, Rule 120 simplicity).

**Suggestion:** Fix _normalize_score to return a true health score, then drop the (100 - support_score) wrapper so the formula reads as a plain weighted sum.

src/services/churn_prediction_service.py
```diff
@@ -0,0 +1,195 @@
...
+                )
+            )
+        )
+        support_ticket_count = int(ticket_result.scalar() or 0)
+
+        return {
+            "login_frequency": login_frequency,
+            "purchase_recency_days": purchase_recency_days,
+            "support_ticket_count": support_ticket_count,
+            "engagement_score_raw": login_frequency,
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** engagement_score_raw is aliased to login_frequency, so login_frequency and engagement_score are perfectly correlated — the 4-dimension engine degenerates to 3 effective inputs (Rule 111 domain model fidelity).

**Suggestion:** Compute engagement from a distinct signal (e.g. count distinct activity types, or a separate engagement_metric column) instead of reusing the login count.

src/services/churn_prediction_service.py
```diff
@@ -0,0 +1,195 @@
...
+        "login_frequency": "login frequency in the last 30 days",
+        "purchase_recency": "days since the last won opportunity",
+        "support_ticket_count": "open/pending support ticket burden",
+        "engagement_score": "overall activity-based engagement",
+    }
+
+    def __init__(self, session: AsyncSession):
+        self.session = session
+
+    async def _fetch_raw_metrics(self, customer_id: int, tenant_id: int) -> dict:
```
yihua-zhuo, line 62: **INFO** (suggestion)

**Issue:** _fetch_raw_metrics issues 4 sequential awaits (customer lookup + 3 aggregate queries); these are independent and could be issued concurrently (Rule 38 async efficiency, Rule 39 batching).

**Suggestion:** Use asyncio.gather for the three aggregate queries after the customer existence check, or combine into fewer queries if the schema allows.

tests/unit/test_churn_prediction_service.py
```diff
@@ -0,0 +1,140 @@
...
+        assert len(result.recommended_actions) >= 1
+
+
+class TestCustomerNotFound:
+    """Error path — raise NotFoundException when customer does not exist."""
+
+    @pytest.fixture
+    def empty_service(self):
+        """Service with no SQL handlers — every SELECT returns an empty result."""
+        session = make_mock_session([])
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** The empty_service fixture builds make_mock_session([]) with zero handlers, so every SELECT silently returns empty — this is a crude ad-hoc mock that bypasses the composable handler pattern recommended by the guidelines (Rule 135 prefer composable mock state).

**Suggestion:** Use make_mock_session([make_customer_handler(state)]) with a state that has no matching customer, so the test exercises the real handler path rather than a bare empty session.

tests/unit/test_churn_prediction_service.py
```diff
@@ -0,0 +1,140 @@
...
+    def test_medium_tier(self):
+        assert ChurnPredictionService._compute_tier(40.0) == "medium"
+        assert ChurnPredictionService._compute_tier(69.99) == "medium"
+
+    def test_low_tier(self):
+        assert ChurnPredictionService._compute_tier(0.0) == "low"
+        assert ChurnPredictionService._compute_tier(39.99) == "low"
+
+
+class TestNormalizeScore:
```
yihua-zhuo: **INFO** (suggestion)

**Issue:** TestNormalizeScore and TestTierBoundaries test private static methods in isolation; while not harmful, this duplicates coverage already exercised by the public calculate_score path and adds maintenance surface (Rule 66 mocking/testing proportionality).

**Suggestion:** Consider removing these classes and instead asserting tier/score values in seeded end-to-end tests (e.g. seed state.activities to drive known sub-scores), which would also strengthen the tenant-isolation test.

yihua-zhuo:

**Status:** `pass`
**Inline comments posted:** 23

**Summary:** Reviewed 4 file(s) between origin/master and origin/auto/issue-573.

src/services/churn_prediction_service.py
```diff
@@ -0,0 +1,204 @@
...
+            select(CustomerModel).where(and_(CustomerModel.id == customer_id, CustomerModel.tenant_id == tenant_id))
+        )
+        customer = result.scalar_one_or_none()
+        if customer is None:
+            raise NotFoundException("Customer")
+
+        now = datetime.now(UTC)
+        login_window = now - timedelta(days=30)
+
+        login_result, engagement_result, purchase_result, ticket_result = await asyncio.gather(
```
yihua-zhuo: **CRITICAL** (critical)

**Issue:** Rule 131: four concurrent `session.execute` calls via `asyncio.gather` against the same AsyncSession can interleave and corrupt the session's transactional state, violating async-safety guarantees.

**Suggestion:** Execute the four queries sequentially with individual `await` statements — concurrent execution on a single session is not supported by SQLAlchemy's async API.

tests/unit/test_churn_prediction_service.py
```diff
@@ -0,0 +1,154 @@
...
+        "updated_at": None,
+        "score": None,
+        "tier": None,
+        "score_factors": None,
+        "top_factors": None,
+        "recommendations": None,
+    }
+
+
+def _make_tenant_aware_customer_handler(state: MockState):
```
yihua-zhuo: **CRITICAL** (critical)

**Issue:** Rule 149 / Rule 125: this test file defines its own customer handler (`_make_tenant_aware_customer_handler`) and monkey-patches around the shared `make_customer_handler` — domain-specific mock logic should live in `tests/unit/domain_handlers/`, not in the test file itself.

**Suggestion:** Create `tests/unit/domain_handlers/churn.py` with `get_handlers(state)` exposing the tenant-aware customer handler, and import it here.

.plans/issue-573.md
```diff
@@ -0,0 +1,86 @@
+# Implementation Plan — Issue #573
+
+## Goal
+Create a new `ChurnPredictionService` class in `src/services/churn_prediction_service.py` that computes a real-time 0-100 churn score from four customer dimensions (login_frequency, purchase_recency, support_ticket_count, engagement_score) using a weighted formula. The service returns a domain object (`ChurnPrediction` dataclass) with score, tier, top-3 risk factors, and recommended actions — no DB writes. This fills a gap left by the existing DB-backed `churn_prediction.py`, which will be complemented by this rule-based entry point. The API router and downstream features (#672, #673, #674) depend on this service being available.
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** Plan states the service is "no DB writes" but the implementation actually executes four read queries (login, engagement, purchase, tickets) — misleading scope description.

**Suggestion:** Clarify the plan to say "no DB writes (read-only queries against activity, opportunity, and ticket tables)".

.plans/issue-573.md
```diff
@@ -0,0 +1,86 @@
...
+   - `mock_db_session` fixture: `state = MockState(); state.customers[1] = {"id": 1, "tenant_id": 1, "name": "Test"}; return make_mock_session([make_customer_handler(state), make_count_handler(state)])`.
+   - **Test 1 — Happy path**: `customer_id=1, tenant_id=1` exists. Assert `isinstance(result, ChurnPrediction)`, `0.0 <= result.score <= 100.0`, `result.tier in ("high", "medium", "low")`, `len(result.top_3_risk_factors) == 3`, `isinstance(result.recommended_actions, list) and len(result.recommended_actions) >= 1`.
+   - **Test 2 — Customer not found**: `customer_id=9999, tenant_id=1`. Assert `pytest.raises(NotFoundException)`.
+   - **Test 3 — Returned object is dataclass, not dict**: Assert `result` is `ChurnPrediction` (not `dict`, not `ApiResponse`). Assert all fields are accessible as attributes.
+   - **Test 4 — Tier boundary with seeded metrics**: Use a `mock_db_session` with specific `state.activities` entries to exercise different score ranges and verify `tier` computation logic (note: opportunity/ticket mock data comes from hardcoded handlers, so assertions should validate the tier mapping contract is wired correctly, not exact scores).
```
yihua-zhuo, line 64: **WARNING** (suggestion)

**Issue:** Test 4 references 'hardcoded handlers' for opportunity/ticket data but the test file contains only customer + count handlers, with no evidence the activity/opportunity/ticket queries are actually exercised — plan acceptance and implementation are out of sync.

**Suggestion:** Add domain handlers for activity, opportunity, and ticket queries, or remove Test 4 from the plan since the test file does not implement it.

.plans/issue-573.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Test Plan
+- **Unit tests in `tests/unit/`**: `test_churn_prediction_service.py` — covers happy path (customer exists, all dimensions return data), error path (customer not found → `NotFoundException`), and structural validation (returns `ChurnPrediction` dataclass with correct fields, score in 0-100, valid tier, 3 risk factors, non-empty recommendations).
```
yihua-zhuo: **INFO** (suggestion)

**Issue:** Test plan claims 'all DB reads are exercised through unit-test mocks' but only customer + count handlers are wired; the four SQL queries against activity/opportunity/ticket tables have no mock support in the test file.

**Suggestion:** Add handlers for activity login count, engagement distinct-type count, opportunity max date, and ticket count — or document that those paths are only verified via the integration DB in a future PR.

src/services/churn_prediction_service.py
```diff
@@ -0,0 +1,204 @@
+"""ChurnPredictionService — rule-based real-time churn scoring (0-100).
+
+Computes a weighted churn score from four customer dimensions and returns
+a ``ChurnPrediction`` dataclass. No DB writes — all queries are read-only.
+"""
+
+import asyncio
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** `import asyncio` is used only for `asyncio.gather` on four separate `session.execute` calls against the same connection — running them concurrently via a shared session can exhaust the connection pool and is unsafe with SQLAlchemy async sessions that use a single connection per session.

**Suggestion:** Replace `asyncio.gather` with sequential `await self.session.execute(...)` calls — a single AsyncSession is not designed for concurrent statement execution.

src/services/churn_prediction_service.py
```diff
@@ -0,0 +1,204 @@
...
+        )
+
+        login_frequency = int(login_result.scalar() or 0)
+        engagement_diversity = int(engagement_result.scalar() or 0)
+
+        last_purchase = purchase_result.scalar()
+        if last_purchase is not None:
+            if last_purchase.tzinfo is None:
+                last_purchase = last_purchase.replace(tzinfo=UTC)
+            purchase_recency_days = max(0, (now - last_purchase).days)
```
yihua-zhuo: **INFO** (suggestion)

**Issue:** `purchase_recency_days` uses integer `.days` from a `timedelta`, which truncates fractional days — a 36-hour-old purchase is reported as 1 day, potentially misclassifying recency.

**Suggestion:** Use `(now - last_purchase).total_seconds() / 86400.0` for sub-day precision or document the integer-truncation behavior.

src/services/churn_prediction_service.py
```diff
@@ -0,0 +1,204 @@
...
+        """Compute churn score, tier, top factors, and recommended actions."""
+        raw = await self._fetch_raw_metrics(customer_id, tenant_id)
+
+        name_to_score = {
+            "login_frequency": self._normalize_score("login_frequency", raw["login_frequency"]),
+            "purchase_recency": self._normalize_score("purchase_recency", raw["purchase_recency_days"]),
+            "support_ticket_count": self._normalize_score("support_ticket_count", raw["support_ticket_count"]),
+            "engagement_score": self._normalize_score("engagement_score", raw["engagement_score_raw"]),
+        }
+        raw_total = sum(score * self.WEIGHTS[name] for name, score in name_to_score.items())
```
yihua-zhuo, line 182: **INFO** (suggestion)

**Issue:** WEIGHTS are all 0.25 and the plan labels this 'no double-inversions', but `_build_top_3_factors` sorts by sub-score descending (healthiest first) and returns those as 'risk factors' — the top-3 are the healthiest dimensions, not the top risks, which is semantically misleading.

**Suggestion:** Sort ascending by sub-score to return the three dimensions with the *lowest* health (highest risk), or rename the field to `top_3_health_factors`.

src/services/churn_prediction_service.py
```diff
@@ -0,0 +1,204 @@
...
+        score = round(min(raw_total, 100.0), 2)
```
yihua-zhuo: **INFO** (suggestion)

**Issue:** Score is clamped with `min(raw_total, 100.0)` but not with a lower bound — if WEIGHTS are ever changed so their sum exceeds 1.0, scores could exceed 100 silently.

**Suggestion:** Use `max(0.0, min(raw_total, 100.0))` for full clamping, or add a WEIGHTS-sum assertion at import time.

tests/unit/test_churn_prediction_service.py
```diff
@@ -0,0 +1,154 @@
+"""Unit tests for ChurnPredictionService — rule-based real-time scoring."""
+
+from __future__ import annotations
+
+import pytest
+
+from pkg.errors.app_exceptions import NotFoundException
+from services.churn_prediction_service import ChurnPrediction, ChurnPredictionService, ChurnRiskFactor
+from tests.unit.conftest import (
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** Imports `MockResult` and `MockRow` from `tests.unit.conftest` but these are internal mock infrastructure — tests outside `tests/unit/` should not import conftest internals, and even within unit tests these are implementation details.

**Suggestion:** Move `_make_tenant_aware_customer_handler` into a domain-owned file under `tests/unit/domain_handlers/churn.py` per Rule 149 and Rule 125.

tests/unit/test_churn_prediction_service.py
```diff
@@ -0,0 +1,154 @@
...
+    """Customer handler that filters by tenant_id in the SQL predicate.
+
+    The shared ``make_customer_handler`` only matches ``from customers where id`` and ignores
+    tenant_id, which would mask cross-tenant isolation bugs. This handler inspects the
+    tenant_id param and only returns the customer when it matches the seeded record.
+    """
+
+    def handler(sql_text, params):
+        normalized = " ".join(sql_text.split())
+        if "from customers where" in normalized and "customers.id" in normalized:
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** Handler inspects raw SQL text (`'from customers where'`, `'customers.id'`) instead of structured query metadata — fragile to SQL reformatting and violates the spirit of Rule 133 (bind-parameter validation), which requires inspecting actual bind keys.

**Suggestion:** Use a query-type dispatcher (e.g. detect by table name only) and read bind params by name; avoid substring-matching on rendered SQL.

tests/unit/test_churn_prediction_service.py
```diff
@@ -0,0 +1,154 @@
...
+            # SQLAlchemy bind params are suffixed (_1, _2, ...). Accept either form.
+            customer_id = params.get("id") if "id" in params else params.get("id_1")
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** Accepts both `params.get('id')` and `params.get('id_1')` to handle SQLAlchemy's bind suffixing, but the production code uses `CustomerModel.id == customer_id` — the test's fallback logic signals the test is coupling to SQLAlchemy's internal parameter naming, not the service contract.

**Suggestion:** Inspect all numeric suffix variants in a loop, or add a comment explaining this is a SQLAlchemy internal-name workaround, and add a unit test that asserts the service uses the correct parameter name.

tests/unit/test_churn_prediction_service.py
```diff
@@ -0,0 +1,154 @@
...
+
+    return handler
+
+
+@pytest.fixture
+def mock_db_session():
+    state = MockState()
+    _seed_test_customer(state, customer_id=1, tenant_id=1)
+    return make_mock_session(
+        [_make_tenant_aware_customer_handler(state), make_customer_handler(state), make_count_handler(state)],
```
yihua-zhuo: **INFO** (suggestion)

**Issue:** The fixture includes both `_make_tenant_aware_customer_handler` *and* `make_customer_handler` — the second one (shared) ignores tenant_id, which means the wrong-tenant test depends on handler ordering. This is fragile and undocumented.

**Suggestion:** Use only the tenant-aware handler in `mock_db_session`; the shared handler is not needed for this service's tests.

tests/unit/test_churn_prediction_service.py
```diff
@@ -0,0 +1,154 @@
...
+            [_make_tenant_aware_customer_handler(state), make_count_handler(state)],
+            state=state,
+        )
+        return ChurnPredictionService(session)
+
+    async def test_raises_not_found_when_customer_missing(self, tenant_isolated_service):
+        with pytest.raises(NotFoundException):
+            await tenant_isolated_service.calculate_score(customer_id=9999, tenant_id=1)
+
+    async def test_wrong_tenant_raises_not_found(self, tenant_isolated_service):
```
yihua-zhuo, line 103: **INFO** (suggestion)

**Issue:** `test_wrong_tenant_raises_not_found` is a good cross-tenant negative test (Rule 126) but the plan does not list it — plan and test file are out of sync.

**Suggestion:** Add this test to the plan's Test Plan section or note in the plan that it was added during implementation.

.plans/issue-573.md
```diff
@@ -0,0 +1,86 @@
...
+   - **Test 2 — Customer not found**: `customer_id=9999, tenant_id=1`. Assert `pytest.raises(NotFoundException)`.
+   - **Test 3 — Returned object is dataclass, not dict**: Assert `result` is `ChurnPrediction` (not `dict`, not `ApiResponse`). Assert all fields are accessible as attributes.
+   - **Test 4 — Tier boundary with seeded metrics**: Use a `mock_db_session` with specific `state.activities` entries to exercise different score ranges and verify `tier` computation logic (note: opportunity/ticket mock data comes from hardcoded handlers, so assertions should validate the tier mapping contract is wired correctly, not exact scores).
+
+## Test Plan
+- **Unit tests in `tests/unit/`**: `test_churn_prediction_service.py` — covers happy path (customer exists, all dimensions return data), error path (customer not found → `NotFoundException`; wrong tenant also raises `NotFoundException`, added during implementation to verify the tenant predicate in `_fetch_raw_metrics`), and structural validation (returns `ChurnPrediction` dataclass with correct fields, score in 0-100, valid tier, 3 risk factors, non-empty recommendations).
```
yihua-zhuo, line 67: **INFO** (suggestion)

**Issue:** Plan says '3+ passed' as the test gate, but the actual test file implements 8 test methods, which is good — however the plan description of Test 4 (tier boundary) is inconsistent with the implementation that has no such test.

**Suggestion:** Reconcile the plan's Test 4 description with the implemented tests, or implement the missing tier-boundary test.

.plans/issue-573.md
```diff
@@ -0,0 +1,86 @@
...
+Dev-plan target: `/home/runner/work/agent-job/agent-job/docs/dev-plan/60-analytics/0573-implement-churnpredictionservice-with-scoring-logic.md`
+Template depth: `deep`
+Reading order followed:
+1. `/home/runner/work/agent-job/agent-job/docs/dev-plan/README.md`
+2. `/home/runner/work/agent-job/agent-job/docs/dev-plan/_template-deep.md`
+3. `/home/runner/work/agent-job/agent-job/docs/dev-plan/60-analytics/0573-implement-churnpredictionservice-with-scoring-logic.md`
+
+## Affected Files
+- `src/services/churn_prediction_service.py` — **new file**; defines `ChurnRiskFactor` and `ChurnPrediction` dataclasses + `ChurnPredictionService` class with `calculate_score` method
+- `tests/unit/test_churn_prediction_service.py` — **new file**; unit tests with `MockState` + `make_customer_handler` + `make_count_handler`
```
yihua-zhuo, line 16: **INFO** (suggestion)

**Issue:** Plan says tests use `MockState` + `make_customer_handler` + `make_count_handler`, but the implementation uses a custom `make_tenant_aware_customer_handler` from `tests/unit/domain_handlers/churn.py` because the shared `make_customer_handler` ignores tenant_id — this is a significant deviation from the plan that should be documented.

**Suggestion:** Update the plan's 'Affected Files' and test plan to mention `make_tenant_aware_customer_handler` and the `tests/unit/domain_handlers/churn.py` module.

src/services/churn_prediction_service.py
```diff
@@ -0,0 +1,203 @@
+"""ChurnPredictionService — rule-based real-time churn scoring (0-100).
+
+Computes a weighted churn score from four customer dimensions and returns
+a ``ChurnPrediction`` dataclass. No DB writes (read-only queries against
+activity, opportunity, and ticket tables).
+"""
+
+from dataclasses import dataclass, field
```
yihua-zhuo, line 8: **INFO** (suggestion)

**Issue:** Imports `field` from `dataclasses` is only used for default_factory on list fields — this is correct but the module-level docstring doesn't mention the dependency on the `counts.py` mock handler returning fixed values.

**Suggestion:** No code change needed; the docstring is adequate.

src/services/churn_prediction_service.py
```diff
@@ -0,0 +1,203 @@
...
+    tier: str
+    top_3_risk_factors: list[ChurnRiskFactor] = field(default_factory=list)
+    recommended_actions: list[str] = field(default_factory=list)
+
+
+class ChurnPredictionService:
+    """Rule-based churn scoring — 4-dimension weighted engine, no DB writes."""
+
+    WEIGHTS: dict[str, float] = {
+        "login_frequency": 0.25,
```
yihua-zhuo, line 46: **INFO** (suggestion)

**Issue:** WEIGHTS dict assigns 0.25 to all four dimensions — this makes the formula a simple average. The plan justifies this as 'no hardcoded 0.25 literals' but since all weights are identical, the 'weighted' formula reduces to an unweighted mean, which may not match the intent of a 'weighted formula' described in the goal.

**Suggestion:** Consider whether genuinely different weights would better reflect business reality, or document explicitly that equal weighting is intentional.

src/services/churn_prediction_service.py
```diff
@@ -0,0 +1,203 @@
...
+                    OpportunityModel.stage == "won",
+                )
+            )
+        )
+        ticket_result = await self.session.execute(
+            select(func.count(TicketModel.id)).where(
+                and_(
+                    TicketModel.tenant_id == tenant_id,
+                    TicketModel.customer_id == customer_id,
+                    TicketModel.status.in_(("open", "pending")),
```
yihua-zhuo, line 106: **INFO** (suggestion)

**Issue:** Uses `TicketModel.status.in_(("open", "pending"))` with a tuple — this is correct for asyncpg/async SQLAlchemy but plan says `status IN ('open','pending')` — the implementation matches, no issue.

src/services/churn_prediction_service.py
```diff
@@ -0,0 +1,203 @@
...
+    @staticmethod
+    def _normalize_score(name: str, raw: float) -> float:
+        """Map a raw dimension value to a 0-100 health sub-score.
+
+        Higher sub-score = healthier / lower churn risk for that dimension.
+        Each branch returns a true health score, so the caller can sum them
+        directly without double-inverting.
+        """
+        if name == "login_frequency":
+            return min(raw / 10 * 100, 100)
```
yihua-zhuo, line 140: **INFO** (suggestion)

**Issue:** Formula `min(raw / 10 * 100, 100)` uses Python operator precedence where `/` and `*` have equal precedence and evaluate left-to-right, so this is `min((raw/10)*100, 100)` — correct but could be clearer as `min(raw / 10.0 * 100.0, 100.0)` for explicitness.

**Suggestion:** No change needed; the formula is correct.

src/services/churn_prediction_service.py
```diff
@@ -0,0 +1,203 @@
...
+    def _compute_tier(score: float) -> str:
+        if score >= 70:
+            return "high"
+        if score >= 40:
+            return "medium"
+        return "low"
+
+    def _build_top_3_factors(self, name_to_score: dict[str, float]) -> list[ChurnRiskFactor]:
+        """Return the three dimensions with the highest churn risk (lowest health first)."""
+        name_to_risk = {name: 100.0 - score for name, score in name_to_score.items()}
```
yihua-zhuo, line 159: **INFO** (suggestion)

**Issue:** `_build_top_3_factors` docstring says 'lowest health first' but sorts by `name_to_risk` descending (highest risk first) — the naming is correct, the docstring is accurate.

src/services/churn_prediction_service.py
```diff
@@ -0,0 +1,203 @@
...
+        raw = await self._fetch_raw_metrics(customer_id, tenant_id)
+
+        name_to_score = {
+            "login_frequency": self._normalize_score("login_frequency", raw["login_frequency"]),
+            "purchase_recency": self._normalize_score("purchase_recency", raw["purchase_recency_days"]),
+            "support_ticket_count": self._normalize_score("support_ticket_count", raw["support_ticket_count"]),
+            "engagement_score": self._normalize_score("engagement_score", raw["engagement_score_raw"]),
+        }
+        raw_total = sum(score * self.WEIGHTS[name] for name, score in name_to_score.items())
+        score = round(max(0.0, min(raw_total, 100.0)), 2)
```
yihua-zhuo, line 183: **INFO** (suggestion)

**Issue:** Clamps to [0, 100] with `max(0.0, min(raw_total, 100.0))` — correct. However, since all WEIGHTS sum to 1.0 and each normalized score is already in [0, 100], the clamp is defensive but unnecessary in practice.

**Suggestion:** No change needed; defensive clamping is reasonable.

src/services/churn_prediction_service.py
yihua-zhuo, line 140: **WARNING** (suggestion)

**Issue:** All four dimensions have identical weight 0.25, making the 'weighted formula' a simple average — the plan's step 3 says 'no hardcoded 0.25 literals' and uses WEIGHTS as 'single source of truth', but since all values are the same, the weighting provides no differentiation. This may be intentional but is worth flagging.

**Suggestion:** If differentiation is desired, consider: login_frequency=0.30, purchase_recency=0.35, support_ticket_count=0.15, engagement_score=0.20. If equal weighting is intentional, add a comment explaining why.

tests/unit/domain_handlers/churn.py
```diff
@@ -0,0 +1,65 @@
+"""Churn prediction domain SQL mock handlers.
+
+The tenant-aware customer handler in this module is the recommended fixture
+for ChurnPredictionService tests because the shared ``make_customer_handler``
+ignores the ``tenant_id`` bind parameter — using it would silently pass
+cross-tenant queries and mask isolation bugs.
```
yihua-zhuo, line 6: **INFO** (suggestion)

**Issue:** Module docstring correctly explains why this handler exists separately from the shared `make_customer_handler` — good documentation of the design decision.

tests/unit/domain_handlers/churn.py
```diff
@@ -0,0 +1,65 @@
...
+"""
+
+from __future__ import annotations
+
+from tests.unit.conftest import MockResult, MockRow, MockState
+
+ORDER = 10
+
+
+def _bind_value(params: dict, base_name: str):
```
yihua-zhuo, line 16: **INFO** (suggestion)

**Issue:** `_bind_value` helper handles SQLAlchemy's auto-generated suffix variants (e.g. `id` vs `id_1`) — this is a pragmatic solution but creates a test that passes regardless of which placeholder SQLAlchemy chose, which could mask a real SQL change.

**Suggestion:** No change needed for now; the docstring acknowledges this trade-off explicitly.

tests/unit/domain_handlers/churn.py
```diff
@@ -0,0 +1,65 @@
...
+
+    Unlike the shared ``make_customer_handler`` (which matches ``from customers
+    where id`` and ignores tenant_id), this handler inspects the tenant_id
+    bind param and only returns the customer when it matches the seeded
+    record. Wrong-tenant queries therefore return an empty result set,
+    allowing the service to raise NotFoundException at its tenant predicate.
+    """
+
+    def handler(sql_text, params):
+        normalized = " ".join(sql_text.split())
```
yihua-zhuo, line 44: **INFO** (suggestion)

**Issue:** Uses string matching on normalized SQL (`'from customers' in normalized`) — this is consistent with the pattern used by `customers.py` domain handler and acceptable for mock handlers.

tests/unit/domain_handlers/churn.py
```diff
@@ -0,0 +1,65 @@
...
+        if "from customers" not in normalized:
+            return None
+        if "where" not in normalized:
+            return None
+        customer_id = _bind_value(params, "id")
+        tenant_id = _bind_value(params, "tenant_id")
+        if customer_id not in state.customers:
+            return MockResult([])
+        rec = state.customers[customer_id]
+        if tenant_id is None or rec.get("tenant_id") == tenant_id:
```
yihua-zhuo, line 54: **INFO** (suggestion)

**Issue:** Returns the customer when `tenant_id is None` (i.e. when the query has no tenant filter) — this is a deliberate design choice that allows the handler to work with queries that don't include tenant_id, but it means the mock won't catch a service that forgets the tenant filter on the customer query itself (the service _does_ include it, so this is fine in practice).

**Suggestion:** No change needed; the service under test does include tenant_id on the customer query.

tests/unit/test_churn_prediction_service.py
```diff
@@ -0,0 +1,123 @@
+"""Unit tests for ChurnPredictionService — rule-based real-time scoring."""
+
+from __future__ import annotations
+
+import pytest
+
+from pkg.errors.app_exceptions import NotFoundException
```
yihua-zhuo, line 7: **INFO** (suggestion)

**Issue:** Imports `make_count_handler` from `tests.unit.conftest` — this works because conftest.py auto-re-exports `__all__` names from domain handler modules at lines 219-221, but it's an indirect coupling that could break if the re-export mechanism changes.

**Suggestion:** Consider importing directly from `tests.unit.domain_handlers.counts` to make the dependency explicit.

tests/unit/test_churn_prediction_service.py
```diff
@@ -0,0 +1,123 @@
...
+from services.churn_prediction_service import ChurnPrediction, ChurnPredictionService, ChurnRiskFactor
+from tests.unit.conftest import MockState, make_count_handler, make_mock_session
```
yihua-zhuo, line 9: **INFO** (suggestion)

**Issue:** `ChurnRiskFactor` is imported but never used in test assertions beyond `isinstance` — acceptable but the import could be removed if not needed.

**Suggestion:** No change needed; the import is used in `test_top_3_risk_factors`.

tests/unit/test_churn_prediction_service.py
```diff
@@ -0,0 +1,123 @@
...
+from tests.unit.domain_handlers.churn import make_tenant_aware_customer_handler
```
yihua-zhuo, line 10: **INFO** (suggestion)

**Issue:** Importing from `tests.unit.domain_handlers.churn` — correct path, follows the domain-handler pattern.

tests/unit/test_churn_prediction_service.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+def _seed_test_customer(state: MockState, customer_id: int = 1, tenant_id: int = 1) -> None:
```
yihua-zhuo, line 13: **INFO** (suggestion)

**Issue:** `_seed_test_customer` creates a full customer record with many fields that are irrelevant to the churn service (e.g. `phone`, `company`, `tags`, `score`, `tier`) — this is defensive but adds noise.

**Suggestion:** No change needed; the extra fields are harmless and mirror the actual CustomerModel schema.

tests/unit/test_churn_prediction_service.py
```diff
@@ -0,0 +1,123 @@
...
+
+@pytest.fixture
+def service(mock_db_session):
+    return ChurnPredictionService(mock_db_session)
+
+
+class TestCalculateScore:
+    """Happy path — customer exists, all dimensions return data, result is a ChurnPrediction."""
+
+    async def test_returns_dataclass(self, service):
```
yihua-zhuo, line 52: **INFO** (suggestion)

**Issue:** Test `test_returns_dataclass` and `test_not_dict` overlap — both check the return type. The plan describes these as separate tests, which is fine for explicit contract documentation.

**Suggestion:** No change needed; explicit contract tests are acceptable.

tests/unit/test_churn_prediction_service.py
```diff
@@ -0,0 +1,123 @@
...
+
+    async def test_score_in_range(self, service):
+        result = await service.calculate_score(customer_id=1, tenant_id=1)
+        assert 0.0 <= result.score <= 100.0
+
+    async def test_tier_valid(self, service):
+        result = await service.calculate_score(customer_id=1, tenant_id=1)
+        assert result.tier in ("high", "medium", "low")
+
+    async def test_top_3_risk_factors(self, service):
```
yihua-zhuo, line 64: **INFO** (suggestion)

**Issue:** Test plan mentions Test 4 (tier boundary with seeded metrics) but no test in this file seeds `state.activities` or exercises different tier ranges — all happy-path tests use the same default mock data.

**Suggestion:** Add a test that seeds different activity/ticket counts to verify tier boundary logic (e.g. all dimensions healthy → 'high', all unhealthy → 'low').

tests/unit/test_churn_prediction_service.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+class TestCalculateScore:
+    """Happy path — customer exists, all dimensions return data, result is a ChurnPrediction."""
+
+    async def test_returns_dataclass(self, service):
+        result = await service.calculate_score(customer_id=1, tenant_id=1)
+        assert isinstance(result, ChurnPrediction)
+
+    async def test_score_in_range(self, service):
```
yihua-zhuo, line 56: **WARNING** (suggestion)

**Issue:** Tests assert score is in [0, 100] and tier is valid, but with the fixed `make_count_handler` returning 7 for all count queries, the score is deterministic (45.0, tier 'medium') — tests are not actually exercising the normalization logic, they're just checking the return type contract.

**Suggestion:** Add assertions on specific score values (e.g. `assert result.score == 45.0`) or seed custom handler responses to exercise different normalization paths.

tests/unit/test_churn_prediction_service.py
```diff
@@ -0,0 +1,123 @@
...
+        result = await service.calculate_score(customer_id=1, tenant_id=1)
+        assert 0.0 <= result.score <= 100.0
+
+    async def test_tier_valid(self, service):
```
yihua-zhuo, line 60: **INFO** (suggestion)

**Issue:** Tier assertion only checks `tier in ('high', 'medium', 'low')` — with fixed mock data, tier will always be 'medium', so this test never exercises the 'high' or 'low' paths.

**Suggestion:** Add tier-specific tests: seed state to get score ≥70 (assert 'high'), score ≥40 (assert 'medium'), score <40 (assert 'low').

PR #807, opened 2026-06-06
Auto: implement issue #599

### LINKED ISSUE #599: Wire up GET /sales/opportunities/{id}/recommendations endpoint
Subtask of #46
Depends on #598

Add a new router file `src/api/routers/recommendations.py` with `GET /sales/opportunities/{id}/recommendations`. Use `require_auth` dependency, inject `AsyncSession` via `Depends(get_db)`. Call `RecommendationService(session).get_recommendations(id, tenant_id)`, serialize result with `.to_dict()`, wrap in `ApiResponse`. Register router in `src/main.py`. Unit test: 200 on existing opportunity, 404 on missing.

Dev-Plan: docs/dev-plan/20-sales/0599-wire-up-get-sales-opportunities-id-recommendations-endpoint.md

yihua-zhuo:
Closes #599

Plan: `.plans/issue-599.md`

1. **Plan** — Claude wrote a detailed implementation plan to `.plans/issue-599.md`.

Please review before merging.

## Summary by CodeRabbit

* **New Features**
  * Added a protected endpoint that returns personalized sales recommendations (conversion probability, similar opportunities, next-best action).

* **Bug Fixes**
  * Unified error handling to provide consistent JSON error responses across the API.

* **Tests**
  * Added comprehensive unit and integration tests covering success, not-found, internal errors, missing tenant/auth, and integration wiring.

* **Documentation**
  * Added a development plan describing acceptance criteria and testing/linting steps.

### DIFF
--- a/.plans/issue-599.md
+++ b/.plans/issue-599.md
@@ -1,0 +1,61 @@
+I have all the context needed. Here is the plan:
+
+# Implementation Plan — Issue #599
+
+## Goal
+Wire up the `GET /api/v1/sales/opportunities/{id}/recommendations` HTTP endpoint by creating a new router that calls the existing `SalesRecommendationService`, adding an async `get_recommendations` method to that service, and registering the router (which `api/__init__.py`'s `iter_routers()` discovers automatically).
+
+## Source Contract
+Dev-plan target: `/home/runner/work/agent-job/agent-job/docs/dev-plan/20-sales/0599-wire-up-get-sales-opportunities-id-recommendations-endpoint.md`
+Template depth: `deep`
+Reading order followed:
+1. `/home/runner/work/agent-job/agent-job/docs/dev-plan/README.md`
+2. `/home/runner/work/agent-job/agent-job/docs/dev-plan/_template-deep.md`
+3. `/home/runner/work/agent-job/agent-job/docs/dev-plan/20-sales/0599-wire-up-get-sales-opportunities-id-recommendations-endpoint.md`
+
+## Affected Files
+- `src/services/sales_recommendation.py` — change `__init__` to accept `AsyncSession | None = None`; add `async get_recommendations(opportunity_id, tenant_id) -> dict` method that composes `predict_conversion_probability`, `get_similar_customers`, and `get_next_best_action`
+- `src/api/routers/recommendations.py` — new file: define `recommendations_router` with `GET /opportunities/{opp_id}/recommendations`
+- `tests/unit/test_recommendations_router.py` — new file: 4 unit tests (200 success, 404, 500, 401 auth-required)
+
+## Implementation Steps
+
+1. **Modify `src/services/sales_recommendation.py`:**
+   - Add `from sqlalchemy.ext.asyncio import AsyncSession` to imports.
+   - Change `def __init__(self):` → `def __init__(self, session: AsyncSession | None = None):` and store `self.session = session` (CLAUDE.md requires `AsyncSession` type with no default, but this service is synchronous/mock-based and does not issue SQL — `AsyncSession | None = None` is the dev-plan's explicit choice for backward compatibility).
+   - Append `async def get_recommendations(self, opportunity_id: int, tenant_id: int) -> dict:` after line 270 (end of `predict_conversion_probability`). The method returns a dict with keys: `opportunity_id`, `conversion_probability`, `similar_opportunities` (list of `customer_id`/`current_tier`/`monthly_revenue`), `next_best_action` (action/target/reason/confidence). Calls existing synchronous helpers internally (no IO, safe in async context).
+
+2. **Create `src/api/routers/recommendations.py`:**
+   - Define `recommendations_router = APIRouter(prefix="/api/v1/sales", tags=["sales"])`.
+   - Register `@recommendations_router.get("/opportunities/{opp_id}/recommendations")` with `opp_id: int`, `ctx: AuthContext = Depends(require_auth)`, `session: AsyncSession = Depends(get_db)`.
+   - In the handler: instantiate `SalesRecommendationService(session)`, `await service.get_recommendations(opp_id, ctx.tenant_id)`, return `{"success": True, "data": data}`.
+   - No try/catch — `AppException` is handled by the global handler in `main.py:71-77`.
+   - The `iter_routers()` function (`src/api/__init__.py:39`) discovers routers by `name == "router" or name.endswith("_router")`, so `recommendations_router` is auto-registered in `main.py`'s `for router in iter_routers()` loop at `src/main.py:96`. No `main.py` edit needed.
+
+3. **Create `tests/unit/test_recommendations_router.py`:**
+   - Reuse the fixture pattern from `tests/unit/test_sales_router.py:51+`: create a `client_with_service` fixture that monkeypatches `api.routers.recommendations.SalesRecommendationService` to return a `MagicMock`, builds a standalone `FastAPI()`, includes the router, overrides `require_auth` and `get_db` dependencies, and returns a `TestClient`.
+   - Test 1 (`test_success_returns_200`): set `svc.get_recommendations = AsyncMock(return_value={...})`; assert 200 + `body["data"]["opportunity_id"] == 5`.
+   - Test 2 (`test_not_found_returns_404`): set side effect to `NotFoundException("Opportunity")`; assert 404.
+   - Test 3 (`test_internal_error_returns_500`): set side effect to `RuntimeError`; assert 500 (caught by generic handler at `main.py:87-93`).
+   - Test 4 (`test_missing_auth_returns_401`): do not override `require_auth`; assert 401.
+
+## Test Plan
+- Unit tests in `tests/unit/`: new `test_recommendations_router.py` covering 200/404/500/401 per dev-plan §5 Step 3
+- Integration tests in `tests/integration/`: none — the service uses mock/hashed data, no real DB queries; dev-plan §3.1 lists no integration test file
+- Dev-plan verification:
+  - `ruff check src/api/routers/recommendations.py src/services/sales_recommendation.py` → 0 errors (§6)
+  - `ruff format --check src/api/routers/recommendations.py src/services/sales_recommendation.py` → pass (§6)
+  - `PYTHONPATH=src pytest tests/unit/test_recommendations_router.py -v` → 4 passed (§6)
+  - `PYTHONPATH=src pytest tests/unit/test_sales_router.py -v` → all passed (regression check, §6)
+
+## Acceptance Criteria
+- `GET /api/v1/sales/opportunities/{id}/recommendations` returns 200 with `{"success": true, "data": {"opportunity_id": ..., "conversion_probability": ..., "similar_opportunities": [...], "next_best_action": {...}}}` for an existing opportunity
+- Returns 404 via `NotFoundException` for a missing opportunity (inherited from global handler)
+- Returns 401 when no auth token is present (handled by `require_auth` dependency)
+- `SalesRecommendationService.get_recommendations` is an `async` method accepting `(opportunity_id: int, tenant_id: int) -> dict`
+- `iter_routers()` auto-discovers `recommendations_router` without any `main.py` modification
+- All 4 unit tests pass; ruff lint and format checks pass
+
+## Risks / Open Questions
+- The dev-plan sets `__init__(self, session: AsyncSession | None = None)`, which violates CLAUDE.md's global rule "session=None is never allowed." This is a documented intentional choice in the dev-plan §4.4 known-pits because the service does not issue SQL (all data is hash-derived mocks). A follow-up could migrate to a real DB-backed model, which would require `session: AsyncSession` with no default.
+- `SalesRecommendationService.get_similar_customers` and `get_next_best_action` use `random` calls without a fixed seed; results will differ between test runs, so the test assertions must not depend on specific random values.
--- a/src/api/routers/recommendations.py
+++ b/src/api/routers/recommendations.py
@@ -1,0 +1,21 @@
+"""Recommendations router — GET /api/v1/sales/opportunities/{id}/recommendations."""
+
+from fastapi import APIRouter, Depends
+from sqlalchemy.ext.asyncio import AsyncSession
+
+from db.connection import get_db
+from internal.middleware.fastapi_auth import AuthContext, require_auth
+from services.sales_recommendation import SalesRecommendationService
+
+recommendations_router = APIRouter(prefix="/api/v1/sales", tags=["sales"])
+
+
+@recommendations_router.get("/opportunities/{opp_id}/recommendations")
+async def get_opportunity_recommendations(
+    opp_id: int,
+    ctx: AuthContext = Depends(require_auth),
+    session: AsyncSession = Depends(get_db),
+):
+    service = SalesRecommendationService(session)
+    data = await service.get_recommendations(opp_id, ctx.tenant_id)
+    return {"success": True, "data": data}
--- a/src/api/routers/recommendations.py
+++ b/src/api/routers/recommendations.py
@@ -1,6 +1,6 @@
-"""Recommendations router — GET /api/v1/sales/opportunities/{id}/recommendations."""
+"""Recommendations router — GET /api/v1/sales/opportunities/{opportunity_id}/recommendations."""
 
-from fastapi import APIRouter, Depends
+from fastapi import APIRouter, Depends, HTTPException
 from sqlalchemy.ext.asyncio import AsyncSession
 
 from db.connection import get_db
@@ -10,12 +10,31 @@ from services.sales_recommendation import SalesRecommendationService
 recommendations_router = APIRouter(prefix="/api/v1/sales", tags=["sales"])
 
 
-@recommendations_router.get("/opportunities/{opp_id}/recommendations")
+@recommendations_router.get("/opportunities/{opportunity_id}/recommendations")
 async def get_opportunity_recommendations(
-    opp_id: int,
+    opportunity_id: int,
     ctx: AuthContext = Depends(require_auth),
     session: AsyncSession = Depends(get_db),
 ):
-    service = SalesRecommendationService(session)
-    data = await service.get_recommendations(opp_id, ctx.tenant_id)
-    return {"success": True, "data": data}
+    if ctx.tenant_id is None:
+        raise HTTPException(status_code=400, detail="Tenant ID is required")
+    service = SalesRecommendationService()
+    result = await service.get_recommendations(opportunity_id, ctx.tenant_id)
+    next_action = result.next_best_action
+    return {
+        "success": True,
+        "data": {
+            "opportunity_id": result.opportunity_id,
+            "conversion_probability": result.conversion_probability,
+            "similar_opportunities": result.similar_opportunities,
+            "next_best_action": {
+                "action": next_action.action,
+                "target": next_action.target,
+                "reason": next_action.reason,
+                "confidence": next_action.confidence,
+            }
+            if next_action is not None
+            else None,
+        },
+        "message": "Recommendations fetched",
+    }
--- a/src/api/routers/recommendations.py
+++ b/src/api/routers/recommendations.py
@@ -1,10 +1,11 @@
 """Recommendations router — GET /api/v1/sales/opportunities/{opportunity_id}/recommendations."""
 
-from fastapi import APIRouter, Depends, HTTPException
+from fastapi import APIRouter, Depends
 from sqlalchemy.ext.asyncio import AsyncSession
 
 from db.connection import get_db
 from internal.middleware.fastapi_auth import AuthContext, require_auth
+from pkg.errors.app_exceptions import ValidationException
 from services.sales_recommendation import SalesRecommendationService
 
 recommendations_router = APIRouter(prefix="/api/v1/sales", tags=["sales"])
@@ -17,24 +18,11 @@ async def get_opportunity_recommendations(
     session: AsyncSession = Depends(get_db),
 ):
     if ctx.tenant_id is None:
-        raise HTTPException(status_code=400, detail="Tenant ID is required")
-    service = SalesRecommendationService()
+        raise ValidationException("Tenant ID is required")
+    service = SalesRecommendationService(session)
     result = await service.get_recommendations(opportunity_id, ctx.tenant_id)
-    next_action = result.next_best_action
     return {
         "success": True,
-        "data": {
-            "opportunity_id": result.opportunity_id,
-            "conversion_probability": result.conversion_probability,
-            "similar_opportunities": result.similar_opportunities,
-            "next_best_action": {
-                "action": next_action.action,
-                "target": next_action.target,
-                "reason": next_action.reason,
-                "confidence": next_action.confidence,
-            }
-            if next_action is not None
-            else None,
-        },
+        "data": result.to_dict(),
         "message": "Recommendations fetched",
     }
--- a/src/main.py
+++ b/src/main.py
@@ -34,6 +34,38 @@ async def lifespan(app: FastAPI):
     logger.info("app_shutdown")
 
 
+def register_exception_handlers(app: FastAPI) -> None:
+    """Register the global exception handlers.
+
+    Extracted from create_app so tests can apply the same handlers without
+    duplicating the JSON envelopes here.
+    """
+
+    @app.exception_handler(AppException)
+    async def app_exception_handler(request: Request, exc: AppException) -> JSONResponse:
+        logger.error("app_exception", code=exc.code, detail=exc.detail, path=request.url.path)
+        return JSONResponse(
+            status_code=exc.status_code,
+            content={"success": False, "message": exc.detail, "code": exc.code},
+        )
+
+    @app.exception_handler(HTTPException)
+    async def http_exception_handler(request: Request, exc: HTTPException) -> JSONResponse:
+        """Reshape HTTPException to ErrorEnvelope shape for Swagger consistency."""
+        return JSONResponse(
+            status_code=exc.status_code,
+            content={"success": False, "message": exc.detail},
+        )
+
+    @app.exception_handler(Exception)
+    async def generic_exception_handler(request: Request, exc: Exception) -> JSONResponse:
+        logger.error("unhandled_exception", type=type(exc).__name__, detail=str(exc))
+        return JSONResponse(
+            status_code=500,
+            content={"success": False, "message": "Internal server error", "code": "INTERNAL_ERROR"},
+        )
+
+
 def create_app() -> FastAPI:
     # Determine OpenAPI availability based on environment
     docs_url = "/docs" if settings.openapi_enabled else None
@@ -68,29 +100,7 @@ def create_app() -> FastAPI:
     )
 
     # ── Global exception handlers ─────────────────────────────────────────
-    @app.exception_handler(AppException)
-    async def app_exception_handler(request: Request, exc: AppException) -> JSONResponse:
-        logger.error("app_exception", code=exc.code, detail=exc.detail, path=request.url.path)
-        return JSONResponse(
-            status_code=exc.status_code,
-            content={"success": False, "message": exc.detail, "code": exc.code},
-        )
-
-    @app.exception_handler(HTTPException)
-    async def http_exception_handler(request: Request, exc: HTTPException) -> JSONResponse:
-        """Reshape HTTPException to ErrorEnvelope shape for Swagger consistency."""
-        return JSONResponse(
-            status_code=exc.status_code,
-            content={"success": False, "message": exc.detail},
-        )
-
-    @app.exception_handler(Exception)
-    async def generic_exception_handler(request: Request, exc: Exception) -> JSONResponse:
-        logger.error("unhandled_exception", type=type(exc).__name__, detail=str(exc))
-        return JSONResponse(
-            status_code=500,
-            content={"success": False, "message": "Internal server error", "code": "INTERNAL_ERROR"},
-        )
+    register_exception_handlers(app)
 
     # ── Routes ─────────────────────────────────────────────────────────────
     for router in iter_routers():
--- a/src/services/recommendation_service.py
+++ b/src/services/recommendation_service.py
@@ -28,7 +28,7 @@ class RecommendationService:
 
     def __init__(self, session: AsyncSession):
         self.session = session
-        self._sales_svc = SalesRecommendationService()
+        self._sales_svc = SalesRecommendationService(session)
 
     async def get_recommendations(self, opportunity_id: int, tenant_id: int) -> dict:
         key = _cache_key(opportunity_id, tenant_id)
--- a/src/services/recommendation_service.py
+++ b/src/services/recommendation_service.py
@@ -5,38 +5,55 @@ results in a module-level dict for 3600s.
 """
 
 import time
+from dataclasses import dataclass, field
 
 from sqlalchemy import select
 from sqlalchemy.ext.asyncio import AsyncSession
 
 from db.models.opportunity import OpportunityModel
 from pkg.errors.app_exceptions import NotFoundException
-from services.sales_recommendation import SalesRecommendationService
+from services.sales_recommendation import SalesActionRecommendation, SalesRecommendationService
 
 _CACHE_TTL = 3600.0  # seconds
 
 # Module-level singleton cache — survives across service instances within a process.
-_cache: dict[str, tuple[float, dict]] = {}
+_cache: dict[str, tuple[float, "CachedRecommendationResult"]] = {}
 
 
 def _cache_key(opportunity_id: int, tenant_id: int) -> str:
     return f"{opportunity_id}:{tenant_id}"
 
 
-class RecommendationService:
-    __slots__ = ("session", "_sales_svc")
+@dataclass
+class CachedRecommendationResult:
+    """Cached recommendation result returned by RecommendationService.get_recommendations."""
 
+    opportunity_id: int
+    conversion_probability: float
+    next_action: SalesActionRecommendation
+    similar_deals: list[dict] = field(default_factory=list)
+
+    def to_dict(self) -> dict:
+        return {
+            "opportunity_id": self.opportunity_id,
+            "conversion_probability": self.conversion_probability,
+            "next_action": self.next_action.to_dict(),
+            "similar_deals": self.similar_deals,
+        }
+
+
+class RecommendationService:
     def __init__(self, session: AsyncSession):
         self.session = session
         self._sales_svc = SalesRecommendationService(session)
 
-    async def get_recommendations(self, opportunity_id: int, tenant_id: int) -> dict:
+    async def get_recommendations(self, opportunity_id: int, tenant_id: int) -> CachedRecommendationResult:
         key = _cache_key(opportunity_id, tenant_id)
         now = time.time()
         if key in _cache:
-            ts, data = _cache[key]
+            ts, cached = _cache[key]
             if now - ts < _CACHE_TTL:
-                return data
+                return cached
         result = await self.session.execute(
             select(OpportunityModel).where(
                 OpportunityModel.id == opportunity_id,
@@ -46,12 +63,12 @@ class RecommendationService:
         opp = result.scalar_one_or_none()
         if opp is None:
             raise NotFoundException("Opportunity")
-        data = {
-            "opportunity_id": opportunity_id,
-            "conversion_probability": self._sales_svc.predict_conversion_probability(opportunity_id),
-            "next_action": self._sales_svc.get_next_best_action(tenant_id, opp.customer_id),
-            "similar_deals": [],
-        }
+        data = CachedRecommendationResult(
+            opportunity_id=opportunity_id,
+            conversion_probability=self._sales_svc.predict_conversion_probability(opportunity_id),
+            next_action=self._sales_svc.get_next_best_action(tenant_id, opp.customer_id),
+            similar_deals=[],
+        )
         _cache[key] = (now, data)
         return data
 
--- a/src/services/sales_recommendation.py
+++ b/src/services/sales_recommendation.py
@@ -7,9 +7,10 @@ import hashlib
 import random
 from dataclasses import dataclass, field
 
-from sqlalchemy import text
+from sqlalchemy import select
 from sqlalchemy.ext.asyncio import AsyncSession
 
+from db.models.opportunity import OpportunityModel
 from pkg.errors.app_exceptions import NotFoundException
 
 
@@ -51,6 +52,15 @@ class SimilarCustomer:
     monthly_revenue: int
     usage_rate: float
     satisfaction_score: float
+
+    def to_dict(self) -> dict:
+        return {
+            "customer_id": self.customer_id,
+            "current_tier": self.current_tier,
+            "monthly_revenue": self.monthly_revenue,
+            "usage_rate": self.usage_rate,
+            "satisfaction_score": self.satisfaction_score,
+        }
 
 
 @dataclass
@@ -126,8 +136,10 @@ class SalesRecommendationService:
     async def _resolve_customer_id(self, opportunity_id: int, tenant_id: int) -> int:
         """通过 DB 查询解析商机关联的 customer_id。如果商机不存在则抛出 NotFoundException。"""
         result = await self.session.execute(
-            text("SELECT customer_id FROM opportunities WHERE id = :id AND tenant_id = :tenant_id"),
-            {"id": opportunity_id, "tenant_id": tenant_id},
+            select(OpportunityModel.customer_id).where(
+                OpportunityModel.id == opportunity_id,
+                OpportunityModel.tenant_id == tenant_id,
+            )
         )
         row = result.first()
         if row is None:
--- a/src/services/sales_recommendation.py
+++ b/src/services/sales_recommendation.py
@@ -5,9 +5,7 @@
 
 import hashlib
 import random
-from dataclasses import dataclass
-
-from sqlalchemy.ext.asyncio import AsyncSession
+from dataclasses import dataclass, field
 
 
 @dataclass
@@ -42,6 +40,16 @@ class SimilarCustomer:
     satisfaction_score: float
 
 
+@dataclass
+class RecommendationResult:
+    """商机完整推荐结果"""
+
+    opportunity_id: int
+    conversion_probability: float
+    similar_opportunities: list[dict] = field(default_factory=list)
+    next_best_action: SalesActionRecommendation | None = None
+
+
 class SalesRecommendationService:
     """销售推荐服务"""
 
@@ -61,9 +69,9 @@ class SalesRecommendationService:
         "enterprise": ["premium"],
     }
 
-    def __init__(self, session: AsyncSession | None = None):
-        """初始化服务"""
-        self.session = session
+    def __init__(self, seed: int | None = None):
+        """初始化服务。`seed` 控制随机行为以保证测试可复现。"""
+        self._seed = seed
         self._customer_cache: dict[int, dict] = {}
 
     def _get_mock_customer_data(self, tenant_id: int, customer_id: int) -> dict:
@@ -272,15 +280,21 @@ class SalesRecommendationService:
 
         return round(min(max(conversion_prob, 0.0), 1.0), 2)
 
-    async def get_recommendations(self, opportunity_id: int, tenant_id: int) -> dict:
-        """获取商机完整推荐结果。"""
+    async def get_recommendations(self, opportunity_id: int, tenant_id: int) -> RecommendationResult:
+        """获取商机完整推荐结果。
+
+        接受 `opportunity_id` + `tenant_id`，查找关联客户并组装结果。
+        """
+        # NOTE: service is currently mock-based; once migrated to real DB-backed models
+        # the opportunity's `customer_id` must be resolved from the opportunity row.
+        customer_id = opportunity_id
         conversion_prob = self.predict_conversion_probability(opportunity_id)
-        similar = self.get_similar_customers(tenant_id, opportunity_id)
-        next_action = self.get_next_best_action(tenant_id, opportunity_id)
-        return {
-            "opportunity_id": opportunity_id,
-            "conversion_probability": conversion_prob,
-            "similar_opportunities": [
+        similar = self.get_similar_customers(tenant_id, customer_id)
+        next_action = self.get_next_best_action(tenant_id, customer_id)
+        return RecommendationResult(
+            opportunity_id=opportunity_id,
+            conversion_probability=conversion_prob,
+            similar_opportunities=[
                 {
                     "customer_id": s.customer_id,
                     "current_tier": s.current_tier,
@@ -288,10 +302,5 @@ class SalesRecommendationService:
                 }
                 for s in similar
             ],
-            "next_best_action": {
-                "action": next_action.action,
-                "target": next_action.target,
-                "reason": next_action.reason,
-                "confidence": next_action.confidence,
-            },
-        }
+            next_best_action=next_action,
+        )
--- a/src/services/sales_recommendation.py
+++ b/src/services/sales_recommendation.py
@@ -7,6 +7,8 @@ import hashlib
 import random
 from dataclasses import dataclass
 
+from sqlalchemy.ext.asyncio import AsyncSession
+
 
 @dataclass
 class ProductRecommendation:
@@ -59,8 +61,9 @@ class SalesRecommendationService:
         "enterprise": ["premium"],
     }
 
-    def __init__(self):
+    def __init__(self, session: AsyncSession | None = None):
         """初始化服务"""
+        self.session = session
         self._customer_cache: dict[int, dict] = {}
 
     def _get_mock_customer_data(self, tenant_id: int, customer_id: int) -> dict:
@@ -268,3 +271,27 @@ class SalesRecommendationService:
         conversion_prob = base_prob * market_sentiment * competition_factor
 
         return round(min(max(conversion_prob, 0.0), 1.0), 2)
+
+    async def get_recommendations(self, opportunity_id: int, tenant_id: int) -> dict:
+        """获取商机完整推荐结果。"""
+        conversion_prob = self.predict_conversion_probability(opportunity_id)
+        similar = self.get_similar_customers(tenant_id, opportunity_id)
+        next_action = self.get_next_best_action(tenant_id, opportunity_id)
+        return {
+            "opportunity_id": opportunity_id,
+            "conversion_probability": conversion_prob,
+            "similar_opportunities": [
+                {
+                    "customer_id": s.customer_id,
+                    "current_tier": s.current_tier,
+                    "monthly_revenue": s.monthly_revenue,
+                }
+                for s in similar
+            ],
+            "next_best_action": {
+                "action": next_action.action,
+                "target": next_action.target,
+                "reason": next_action.reason,
+                "confidence": next_action.confidence,
+            },
+        }
--- a/src/services/sales_recommendation.py
+++ b/src/services/sales_recommendation.py
@@ -7,6 +7,11 @@ import hashlib
 import random
 from dataclasses import dataclass, field
 
+from sqlalchemy import text
+from sqlalchemy.ext.asyncio import AsyncSession
+
+from pkg.errors.app_exceptions import NotFoundException
+
 
 @dataclass
 class ProductRecommendation:
@@ -27,6 +32,14 @@ class SalesActionRecommendation:
     target: str
     reason: str
     confidence: float
+
+    def to_dict(self) -> dict:
+        return {
+            "action": self.action,
+            "target": self.target,
+            "reason": self.reason,
+            "confidence": self.confidence,
+        }
 
 
 @dataclass
@@ -49,6 +62,14 @@ class RecommendationResult:
     similar_opportunities: list[dict] = field(default_factory=list)
     next_best_action: SalesActionRecommendation | None = None
 
+    def to_dict(self) -> dict:
+        return {
+            "opportunity_id": self.opportunity_id,
+            "conversion_probability": self.conversion_probability,
+            "similar_opportunities": self.similar_opportunities,
+            "next_best_action": self.next_best_action.to_dict() if self.next_best_action else None,
+        }
+
 
 class SalesRecommendationService:
     """销售推荐服务"""
@@ -69,15 +90,13 @@ class SalesRecommendationService:
         "enterprise": ["premium"],
     }
 
-    def __init__(self, seed: int | None = None):
+    def __init__(self, session: AsyncSession, seed: int | None = None):
         """初始化服务。`seed` 控制随机行为以保证测试可复现。"""
-        self._seed = seed
-        self._customer_cache: dict[int, dict] = {}
+        self.session = session
+        self._rng = random.Random(seed)  # noqa: S311 - non-cryptographic recommendation scoring
 
     def _get_mock_customer_data(self, tenant_id: int, customer_id: int) -> dict:
         """获取模拟客户数据"""
-        import hashlib
-
         seed = int(hashlib.sha256(f"{tenant_id}:{customer_id}".encode()).hexdigest()[:8], 16)
 
         tier_index = seed % 4
@@ -89,22 +108,31 @@ class SalesRecommendationService:
             "current_tier": current_tier,
             "usage_rate": (seed % 80 + 20) / 100.0,
             "monthly_revenue": self.PRODUCTS[current_tier]["price"],
-            "purchase_history": random.sample(tiers, (seed % 3) + 1),
-            "browsing_history": random.sample(tiers, (seed % 4) + 1),
+            "purchase_history": self._rng.sample(tiers, (seed % 3) + 1),
+            "browsing_history": self._rng.sample(tiers, (seed % 4) + 1),
             "satisfaction_score": (seed % 100) / 100.0,
         }
 
     def _get_similar_customers_by_tier(self, tenant_id: int, tier: str, limit: int = 5) -> list[int]:
         """基于层级获取相似客户ID"""
-        random.seed(42)  # 固定随机种子保证一致性
         tier_index = list(self.PRODUCTS.keys()).index(tier)
         similar = [
             i
             for i in range(1, 1000)
             if int(hashlib.sha256(f"{tenant_id}:{i}".encode()).hexdigest()[:8], 16) % 4 == tier_index
         ][:limit]
-        random.seed()  # 重置随机种子
         return similar
+
+    async def _resolve_customer_id(self, opportunity_id: int, tenant_id: int) -> int:
+        """通过 DB 查询解析商机关联的 customer_id。如果商机不存在则抛出 NotFoundException。"""
+        result = await self.session.execute(
+            text("SELECT customer_id FROM opportunities WHERE id = :id AND tenant_id = :tenant_id"),
+            {"id": opportunity_id, "tenant_id": tenant_id},
+        )
+        row = result.first()
+        if row is None:
+            raise NotFoundException("Opportunity")
+        return row[0]
 
     def get_next_best_action(self, tenant_id: int, customer_id: int) -> SalesActionRecommendation:
         """
@@ -141,7 +169,7 @@ class SalesRecommendationService:
             action=action,
             target=target,
             reason=reason,
-            confidence=round(random.uniform(0.6, 0.95), 2),  # noqa: S311 - non-security recommendation scoring
+            confidence=round(self._rng.uniform(0.6, 0.95), 2),  # noqa: S311 - non-security recommendation scoring
         )
 
     def recommend_cross_sell(self, tenant_id: int, customer_id: int) -> list[ProductRecommendation]:
@@ -262,8 +290,6 @@ class SalesRecommendationService:
         预测商机成交概率（增强版）
         结合：历史数据、市场情绪、竞争分析
         """
-        import hashlib
-
         seed = int(hashlib.sha256(str(opportunity_id).encode()).hexdigest()[:8], 16)
 
         # 基础概率 (历史数据)
@@ -283,11 +309,10 @@ class SalesRecommendationService:
     async def get_recommendations(self, opportunity_id: int, tenant_id: int) -> RecommendationResult:
         """获取商机完整推荐结果。
 
-        接受 `opportunity_id` + `tenant_id`，查找关联客户并组装结果。
+        接受 `opportunity_id` + `tenant_id`，通过 DB 查询解析关联的 customer_id，
+        然后组装转换概率、相似客户和下一步最佳行动。
         """
-        # NOTE: service is currently mock-based; once migrated to real DB-backed models
-        # the opportunity's `customer_id` must be resolved from the opportunity row.
-        customer_id = opportunity_id
+        customer_id = await self._resolve_customer_id(opportunity_id, tenant_id)
         conversion_prob = self.predict_conversion_probability(opportunity_id)
         similar = self.get_similar_customers(tenant_id, customer_id)
         next_action = self.get_next_best_action(tenant_id, customer_id)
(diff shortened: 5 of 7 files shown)


.plans/issue-599.md
```diff
@@ -0,0 +1,61 @@
...
+## Affected Files
+- `src/services/sales_recommendation.py` — change `__init__` to accept `AsyncSession | None = None`; add `async get_recommendations(opportunity_id, tenant_id) -> dict` method that composes `predict_conversion_probability`, `get_similar_customers`, and `get_next_best_action`
+- `src/api/routers/recommendations.py` — new file: define `recommendations_router` with `GET /opportunities/{opp_id}/recommendations`
+- `tests/unit/test_recommendations_router.py` — new file: 4 unit tests (200 success, 404, 500, 401 auth-required)
+
+## Implementation Steps
+
+1. **Modify `src/services/sales_recommendation.py`:**
+   - Add `from sqlalchemy.ext.asyncio import AsyncSession` to imports.
+   - Change `def __init__(self):` → `def __init__(self, session: AsyncSession | None = None):` and store `self.session = session` (CLAUDE.md requires `AsyncSession` type with no default, but this service is synchronous/mock-based and does not issue SQL — `AsyncSession | None = None` is the dev-plan's explicit choice for backward compatibility).
```
yihua-zhuo, line 25: **CRITICAL** (critical)

**Issue:** Plan documents `__init__(self, session: AsyncSession | None = None)` with a default of `None`, violating project rule 121 / CLAUDE.md 'session=None is never allowed'; the risk section acknowledges this but still prescribes the violation.

**Suggestion:** Type the parameter as `session: AsyncSession` with no default and require the router to pass the dependency-injected session; migrate the service to real DB-backed models or keep the service purely stateless (remove `self.session` and the parameter entirely).

src/services/sales_recommendation.py
```diff
@@ -59,8 +61,9 @@ class SalesRecommendationService:
         "enterprise": ["premium"],
     }
 
-    def __init__(self):
+    def __init__(self, session: AsyncSession | None = None):
```
yihua-zhuo: **CRITICAL** (critical)

**Issue:** `__init__(self, session: AsyncSession | None = None)` allows `session=None`, violating project rule 121 / CLAUDE.md 'session=None is never allowed'.

**Suggestion:** Change to `def __init__(self, session: AsyncSession):` with no default; remove the stored `self.session` since the service does not use it.

src/services/sales_recommendation.py
```diff
@@ -268,3 +271,27 @@ def predict_conversion_probability(self, opportunity_id: int) -> float:
         conversion_prob = base_prob * market_sentiment * competition_factor
 
         return round(min(max(conversion_prob, 0.0), 1.0), 2)
+
+    async def get_recommendations(self, opportunity_id: int, tenant_id: int) -> dict:
+        """获取商机完整推荐结果。"""
+        conversion_prob = self.predict_conversion_probability(opportunity_id)
+        similar = self.get_similar_customers(tenant_id, opportunity_id)
+        next_action = self.get_next_best_action(tenant_id, opportunity_id)
+        return {
```
yihua-zhuo: **CRITICAL** (critical)

**Issue:** `get_recommendations` returns a raw `dict` instead of a domain object or `(items, total)` tuple, violating project rule 127 ('Services return domain objects; routers serialize').

**Suggestion:** Define a `RecommendationResult` dataclass (or similar domain model) with `opportunity_id`, `conversion_probability`, `similar_opportunities`, `next_best_action` fields; return that from the service and let the router call `.to_dict()` or manually serialize.

.plans/issue-599.md
yihua-zhuo, line 25: **WARNING** (suggestion)

**Issue:** Plan stores `self.session` in `__init__` but the service never issues SQL; the stored session is dead state that misleads future maintainers into thinking the service is DB-backed.

**Suggestion:** If the service remains mock-only, drop the `session` parameter from `__init__` entirely — the router already gets a session but can pass `None` or the service constructor can take no arguments.

.plans/issue-599.md
```diff
@@ -0,0 +1,61 @@
...
+- Returns 404 via `NotFoundException` for a missing opportunity (inherited from global handler)
+- Returns 401 when no auth token is present (handled by `require_auth` dependency)
+- `SalesRecommendationService.get_recommendations` is an `async` method accepting `(opportunity_id: int, tenant_id: int) -> dict`
+- `iter_routers()` auto-discovers `recommendations_router` without any `main.py` modification
+- All 4 unit tests pass; ruff lint and format checks pass
+
+## Risks / Open Questions
+- The dev-plan sets `__init__(self, session: AsyncSession | None = None)`, which violates CLAUDE.md's global rule "session=None is never allowed." This is a documented intentional choice in the dev-plan §4.4 known-pits because the service does not issue SQL (all data is hash-derived mocks). A follow-up could migrate to a real DB-backed model, which would require `session: AsyncSession` with no default.
+- `SalesRecommendationService.get_similar_customers` and `get_next_best_action` use `random` calls without a fixed seed; results will differ between test runs, so the test assertions must not depend on specific random values.
```
yihua-zhuo, line 61: **INFO** (suggestion)

**Issue:** Plan notes that `random` is used without a fixed seed in `get_similar_customers` and `get_next_best_action`; this makes the service non-deterministic across calls within a single test run, violating rule 63 (tests are deterministic).

**Suggestion:** Add a seed parameter or inject a random generator so callers (and tests) can control determinism; or inject `seed` from the router/config.

.plans/issue-599.md
```diff
@@ -0,0 +1,61 @@
...
+   - Reuse the fixture pattern from `tests/unit/test_sales_router.py:51+`: create a `client_with_service` fixture that monkeypatches `api.routers.recommendations.SalesRecommendationService` to return a `MagicMock`, builds a standalone `FastAPI()`, includes the router, overrides `require_auth` and `get_db` dependencies, and returns a `TestClient`.
+   - Test 1 (`test_success_returns_200`): set `svc.get_recommendations = AsyncMock(return_value={...})`; assert 200 + `body["data"]["opportunity_id"] == 5`.
+   - Test 2 (`test_not_found_returns_404`): set side effect to `NotFoundException("Opportunity")`; assert 404.
+   - Test 3 (`test_internal_error_returns_500`): set side effect to `RuntimeError`; assert 500 (caught by generic handler at `main.py:87-93`).
+   - Test 4 (`test_missing_auth_returns_401`): do not override `require_auth`; assert 401.
+
+## Test Plan
+- Unit tests in `tests/unit/`: new `test_recommendations_router.py` covering 200/404/500/401 per dev-plan §5 Step 3
+- Integration tests in `tests/integration/`: none — the service uses mock/hashed data, no real DB queries; dev-plan §3.1 lists no integration test file
```
yihua-zhuo, line 44: **INFO** (suggestion)

**Issue:** Plan says no integration tests are needed because the service uses mock data; this is acceptable only if the service remains a pure mock — the plan should mark the follow-up migration (real DB) as a future work item with a clear trigger.

**Suggestion:** Add a note: 'When the service is migrated to real DB-backed models, add integration tests under tests/integration/ using async_session, db_schema, and tenant_id fixtures.'

src/api/routers/recommendations.py
```diff
@@ -0,0 +1,21 @@
...
+from sqlalchemy.ext.asyncio import AsyncSession
+
+from db.connection import get_db
+from internal.middleware.fastapi_auth import AuthContext, require_auth
+from services.sales_recommendation import SalesRecommendationService
+
+recommendations_router = APIRouter(prefix="/api/v1/sales", tags=["sales"])
+
+
+@recommendations_router.get("/opportunities/{opp_id}/recommendations")
```
yihua-zhuo: **INFO** (suggestion)

**Issue:** Path parameter is named `opp_id` while the service method parameter is `opportunity_id`; mixing abbreviations across layers reduces readability (rule 11, 12).

**Suggestion:** Rename the path parameter to `opportunity_id` to match the service method and the issue title; update the route path to `/opportunities/{opportunity_id}/recommendations`.

src/services/sales_recommendation.py
```diff
@@ -59,8 +61,9 @@ class SalesRecommendationService:
         "enterprise": ["premium"],
     }
 
-    def __init__(self):
+    def __init__(self, session: AsyncSession | None = None):
         """初始化服务"""
+        self.session = session
```
yihua-zhuo, line 105: **WARNING** (suggestion)

**Issue:** `self.session` is stored but never read anywhere in the class; dead state (rule 17 — dead code removed).

**Suggestion:** Remove the `self.session` assignment and the `session` parameter; if the service is truly stateless, do not accept a session at all.

src/services/sales_recommendation.py
```diff
@@ -268,3 +271,27 @@ def predict_conversion_probability(self, opportunity_id: int) -> float:
         conversion_prob = base_prob * market_sentiment * competition_factor
 
         return round(min(max(conversion_prob, 0.0), 1.0), 2)
+
+    async def get_recommendations(self, opportunity_id: int, tenant_id: int) -> dict:
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** `get_recommendations` is declared `async` but only calls synchronous methods (`predict_conversion_probability`, `get_similar_customers`, `get_next_best_action`); the `async` keyword is misleading and could hide future blocking IO (rule 131).

**Suggestion:** Either make the method synchronous if no IO is involved, or document why the async signature is needed for future compatibility; do not leave it as an empty async wrapper.

src/services/sales_recommendation.py
```diff
@@ -268,3 +271,27 @@ def predict_conversion_probability(self, opportunity_id: int) -> float:
...
+        """获取商机完整推荐结果。"""
+        conversion_prob = self.predict_conversion_probability(opportunity_id)
+        similar = self.get_similar_customers(tenant_id, opportunity_id)
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** `opportunity_id` is passed as `customer_id` to `get_similar_customers` and `get_next_best_action`; conflating two different domain entities in the mock logic is a domain-modeling concern (rule 111).

**Suggestion:** Pass an explicit `customer_id` (e.g., look up the opportunity's customer_id) or rename the mock functions to accept an `entity_id` and document the intentional conflation.

tests/unit/test_recommendations_router.py
```diff
@@ -0,0 +1,88 @@
...
+        assert body["data"]["opportunity_id"] == 5
+        assert body["data"]["conversion_probability"] == 0.72
+
+    def test_not_found_returns_404(self, client_with_service):
+        client, svc = client_with_service
+        svc.get_recommendations = AsyncMock(side_effect=NotFoundException("Opportunity"))
+        resp = client.get("/api/v1/sales/opportunities/9999/recommendations")
+        assert resp.status_code == 404
+
+    def test_internal_error_returns_500(self, client_with_service):
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** `test_internal_error_returns_500` creates its own `FastAPI()` app inside `client_with_service` without registering the generic exception handler from `main.py:87-93`; the 500 comes from Starlette's default behavior with `raise_server_exceptions=False`, not from the handler the plan cites.

**Suggestion:** Either register the project's generic exception handler in the test app, or assert that the response body matches the standard error envelope shape so the test verifies actual handler behavior, not just the status code.

tests/unit/test_recommendations_router.py
```diff
@@ -0,0 +1,88 @@
...
+        client, svc = client_with_service
+        svc.get_recommendations = AsyncMock(side_effect=RuntimeError("unexpected"))
+        resp = client.get("/api/v1/sales/opportunities/1/recommendations")
+        assert resp.status_code == 500
+
+    def test_missing_auth_returns_401(self):
```
yihua-zhuo: **INFO** (suggestion)

**Issue:** `test_missing_auth_returns_401` creates a separate `FastAPI()` app instead of reusing `client_with_service`; the monkeypatched `SalesRecommendationService` from the fixture does not apply, so this test isolates auth behavior — acceptable but means the fixture's name overpromises (rule 67).

**Suggestion:** Rename `client_with_service` to `mocked_service_client` or split into two fixtures: one with mocked service + overridden auth, and one with no auth override for the 401 test.

yihua-zhuo:

**Status:** `fail`
**Inline comments posted:** 21

**Summary:** Reviewed 7 file(s) between origin/master and origin/auto/issue-599.

`tests/integration/test_intelligence_integration.py` L122: 
**CRITICAL** (critical)

**Issue:** TestSalesRecommendationService tests do not exercise the database at all — they test pure hash/mock logic, which belongs in tests/unit/ (Rule 124, Rule 61).

**Suggestion:** Move these tests to tests/unit/test_sales_recommendation.py, or add DB-backed tests that seed an opportunity and call get_recommendations() end-to-end. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`src/main.py` L32: 
**INFO** (suggestion)

**Issue:** Bare `except Exception:` in the shutdown path swallows all errors; the `logger.warning(..., exc_info=True)` mitigates silent swallowing (Rule 46) but the broad except could mask real teardown bugs.

**Suggestion:** Narrow to `except httpx.HTTPError` or re-raise after logging, or at minimum log with a distinct event name so it's grep-able. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`src/main.py` L74: 
**INFO** (suggestion)

**Issue:** `CORS` is configured with `allow_origins=allowed_origins` (comma-separated from `cors_origins` setting) and `allow_credentials=True` with wildcard `allow_methods`/`allow_headers` — this is overly permissive if `cors_origins` is misconfigured to include a wildcard.

**Suggestion:** Validate `cors_origins` setting at startup to reject wildcard origins when credentials are enabled, and consider explicit method/header allow-lists. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`src/main.py` L109: 
**WARNING** (suggestion)

**Issue:** The `/` health endpoint always returns 200 without checking DB connectivity or LLM client health — not a proper readiness/liveness probe (Rule 74). Other health endpoints exist in `routers/health.py`.

**Suggestion:** Remove the bare `/` health check or rename to `/live` and add a separate `/ready` that verifies DB engine and httpx client. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`src/main.py` L85: 
**INFO** (suggestion)

**Issue:** JWT secret/algorithm are stored on `app.state` for dependency access — this couples config to FastAPI state and makes testing harder. Rule 7 suggests config should be separated from app wiring.

**Suggestion:** Access `settings.jwt_secret` directly in the auth dependency rather than threading through `app.state`. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`src/services/recommendation_service.py` L75: 
**INFO** (suggestion)

**Issue:** `invalidate_cache` is a `@staticmethod` that mutates module-level state — this is hard to test and creates hidden coupling (Rule 112). Consider injecting the cache.

**Suggestion:** Accept a `cache` parameter or use a class-level cache for testability. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`src/services/sales_recommendation.py` L19: 
**INFO** (suggestion)

**Issue:** Comments and docstrings are in Chinese while the rest of the codebase uses English — breaks consistency (Rule 20).

**Suggestion:** Translate docstrings and comments to English, e.g. `"""Product recommendation"""`. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`src/services/sales_recommendation.py` L88: 
**INFO** (suggestion)

**Issue:** PRODUCTS and CROSS_SELL_MAP are hardcoded class-level dicts (Rule 6, Rule 19) — magic constants should be extracted to config.

**Suggestion:** Move PRODUCTS and CROSS_SELL_MAP to a config module or a module-level constant, or load from settings. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`src/services/sales_recommendation.py` L200: 
**INFO** (suggestion)

**Issue:** Magic scoring weights 0.3, 0.2, 0.15 are used inline with no named constants (Rule 19).

**Suggestion:** Extract `_WEIGHT_BROWSING = 0.3`, `_WEIGHT_CROSS_SELL = 0.2`, `_WEIGHT_SIMILAR = 0.15` as module constants. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`src/services/sales_recommendation.py` L22: 
**WARNING** (suggestion)

**Issue:** ProductRecommendation dataclass lacks a `to_dict()` method while SimilarCustomer and SalesActionRecommendation both have one — inconsistent serialization contract (Rule 127, Rule 17).

**Suggestion:** Add a `to_dict()` method to ProductRecommendation matching the pattern of the other dataclasses. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`src/services/sales_recommendation.py` L300: 
**WARNING** (suggestion)

**Issue:** `predict_conversion_probability` accepts no `tenant_id` and is a pure hash-based mock — this is acceptable for testing but should be documented as a mock stub to avoid being mistaken for production logic.

**Suggestion:** Add a docstring note: `'# NOTE: mock implementation — replace with real model in production'`. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`src/services/sales_recommendation.py` L230: 
**INFO** (suggestion)

**Issue:** Max-recommendation limit 5 is a magic number (Rule 19).

**Suggestion:** Extract `_MAX_CROSS_SELL_RESULTS = 5` as a module constant. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`src/services/sales_recommendation.py` L158: 
**INFO** (suggestion)

**Issue:** Comments are what-comments (e.g. `# 决策逻辑`, `# 获取推荐产品`) rather than why-comments (Rule 16).

**Suggestion:** Remove comments that restate what the code does, or rewrite them to explain the business reasoning. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`tests/integration/test_intelligence_integration.py` L202: 
**WARNING** (suggestion)

**Issue:** TestSmartCategorizationService does not use the async_session fixture and tests a service with no DB dependency — it does not belong in tests/integration/ (Rule 124).

**Suggestion:** Move TestSmartCategorizationService to tests/unit/test_smart_categorization.py. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`tests/integration/test_intelligence_integration.py` L7: 
**WARNING** (suggestion)

**Issue:** Manual sys.path manipulation duplicates what PYTHONPATH=src should provide — bypasses project test configuration (Rule 71, project convention).

**Suggestion:** Remove the sys.path block and rely on the pytest configuration / PYTHONPATH=src as documented in CLAUDE.md. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`tests/integration/test_intelligence_integration.py` L21: 
**INFO** (suggestion)

**Issue:** `_seed_customer` is a cross-service seed helper defined inline — per Rule 125, domain fixtures should live in tests/integration/domain_fixtures/<domain>.py.

**Suggestion:** Move _seed_customer to tests/integration/domain_fixtures/customer.py and import it from there. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`tests/integration/test_intelligence_integration.py` L91: 
**INFO** (suggestion)

**Issue:** `test_get_churn_risk_factors` asserts `len(factors) == 6` and checks the exact set of factor names — this is an exact-match assertion that will break if a factor is added/renamed (Rule 63, Rule 49).

**Suggestion:** Use `assert len(factors) >= 1` and check that expected names are a subset, or document that 6 is a hard contract. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`tests/integration/test_intelligence_integration.py` L60: 
**INFO** (suggestion)

**Issue:** No not-found test for SalesRecommendationService.get_recommendations (e.g. non-existent opportunity_id) — failure path not tested (Rule 65).

**Suggestion:** Add a test that calls get_recommendations with a non-existent opportunity_id and asserts NotFoundException. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_

yihua-zhuo:
**Feature scope appears incomplete vs #599** ⚠️

The router file src/api/routers/recommendations.py does not exist, the router is not registered in src/main.py, and the requested unit test file tests/unit/test_recommendations_router.py does not exist.

**Missing:**
- src/api/routers/recommendations.py not created
- Router not registered in src/main.py
- tests/unit/test_recommendations_router.py not created (200/404 tests)

.plans/issue-599.md
```diff
@@ -0,0 +1,61 @@
...
+## Source Contract
+Dev-plan target: `/home/runner/work/agent-job/agent-job/docs/dev-plan/20-sales/0599-wire-up-get-sales-opportunities-id-recommendations-endpoint.md`
+Template depth: `deep`
+Reading order followed:
+1. `/home/runner/work/agent-job/agent-job/docs/dev-plan/README.md`
+2. `/home/runner/work/agent-job/agent-job/docs/dev-plan/_template-deep.md`
+3. `/home/runner/work/agent-job/agent-job/docs/dev-plan/20-sales/0599-wire-up-get-sales-opportunities-id-recommendations-endpoint.md`
+
+## Affected Files
+- `src/services/sales_recommendation.py` — change `__init__` to accept `AsyncSession | None = None`; add `async get_recommendations(opportunity_id, tenant_id) -> dict` method that composes `predict_conversion_probability`, `get_similar_customers`, and `get_next_best_action`
```
yihua-zhuo, line 17: **CRITICAL** (critical)

**Issue:** Plan is stale and contradicts the implementation: says `__init__` should accept `session: AsyncSession | None = None` and `get_recommendations` should return `-> dict`, but the actual code has `__init__(self, seed: int | None = None)` and `get_recommendations` returns `RecommendationResult` (Rule 140).

**Suggestion:** Update the Affected Files section to match the real implementation: `__init__` takes `seed`, and `get_recommendations` returns `RecommendationResult`.

src/api/routers/recommendations.py
```diff
@@ -0,0 +1,40 @@
...
+from services.sales_recommendation import SalesRecommendationService
+
+recommendations_router = APIRouter(prefix="/api/v1/sales", tags=["sales"])
+
+
+@recommendations_router.get("/opportunities/{opportunity_id}/recommendations")
+async def get_opportunity_recommendations(
+    opportunity_id: int,
+    ctx: AuthContext = Depends(require_auth),
+    session: AsyncSession = Depends(get_db),
```
yihua-zhuo, line 18: **CRITICAL** (critical)

**Issue:** `session: AsyncSession = Depends(get_db)` is injected but never passed to `SalesRecommendationService()` on line 21 — the dependency is dead code and the request-scoped transaction is created and discarded without use (Rule 121).

**Suggestion:** Remove the unused `session` parameter, or pass it to the service constructor so the DB session lifecycle is meaningful.

src/api/routers/recommendations.py
```diff
@@ -0,0 +1,40 @@
...
+):
+    if ctx.tenant_id is None:
```
yihua-zhuo, line 20: **CRITICAL** (critical)

**Issue:** Router raises raw `HTTPException(status_code=400)` for missing tenant_id, bypassing the `AppException` hierarchy and global handler — this produces a different response envelope than other endpoints (Rule 128).

**Suggestion:** Replace with `raise ValidationException("Tenant ID is required")` from `pkg.errors.app_exceptions` so the global handler shapes a consistent error response.

src/services/sales_recommendation.py
```diff
@@ -268,3 +279,28 @@ def predict_conversion_probability(self, opportunity_id: int) -> float:
...
         return round(min(max(conversion_prob, 0.0), 1.0), 2)
+
+    async def get_recommendations(self, opportunity_id: int, tenant_id: int) -> RecommendationResult:
+        """获取商机完整推荐结果。
+
+        接受 `opportunity_id` + `tenant_id`，查找关联客户并组装结果。
+        """
+        # NOTE: service is currently mock-based; once migrated to real DB-backed models
+        # the opportunity's `customer_id` must be resolved from the opportunity row.
+        customer_id = opportunity_id
```
yihua-zhuo: **CRITICAL** (critical)

**Issue:** `customer_id = opportunity_id` is a placeholder that treats the opportunity ID as a customer ID — `get_similar_customers` and `get_next_best_action` will return results for the wrong entity. The NOTE comment acknowledges this but the method is still callable in production (Rule 111).

**Suggestion:** Resolve the actual customer_id from the opportunity via a DB query (e.g., `SELECT customer_id FROM opportunities WHERE id = :id AND tenant_id = :tenant_id`), or raise `NotFoundException` if the opportunity doesn't exist.

.plans/issue-599.md
```diff
@@ -0,0 +1,61 @@
...
+- `src/api/routers/recommendations.py` — new file: define `recommendations_router` with `GET /opportunities/{opp_id}/recommendations`
+- `tests/unit/test_recommendations_router.py` — new file: 4 unit tests (200 success, 404, 500, 401 auth-required)
```
yihua-zhuo, line 19: **WARNING** (suggestion)

**Issue:** Plan says '4 unit tests (200 success, 404, 500, 401 auth-required)' but the actual test file has 5 tests — the missing one is `test_missing_tenant_returns_400` (Rule 140).

**Suggestion:** Update the test count to 5 and add the missing-tenant test to the description.

.plans/issue-599.md
```diff
@@ -0,0 +1,61 @@
...
+   - Append `async def get_recommendations(self, opportunity_id: int, tenant_id: int) -> dict:` after line 270 (end of `predict_conversion_probability`). The method returns a dict with keys: `opportunity_id`, `conversion_probability`, `similar_opportunities` (list of `customer_id`/`current_tier`/`monthly_revenue`), `next_best_action` (action/target/reason/confidence). Calls existing synchronous helpers internally (no IO, safe in async context).
+
+2. **Create `src/api/routers/recommendations.py`:**
+   - Define `recommendations_router = APIRouter(prefix="/api/v1/sales", tags=["sales"])`.
+   - Register `@recommendations_router.get("/opportunities/{opp_id}/recommendations")` with `opp_id: int`, `ctx: AuthContext = Depends(require_auth)`, `session: AsyncSession = Depends(get_db)`.
+   - In the handler: instantiate `SalesRecommendationService(session)`, `await service.get_recommendations(opp_id, ctx.tenant_id)`, return `{"success": True, "data": data}`.
```
yihua-zhuo, line 31: **WARNING** (suggestion)

**Issue:** Plan says the router will instantiate `SalesRecommendationService(session)`, but the actual router calls `SalesRecommendationService()` with no arguments — the session dependency is never used (Rule 140).

**Suggestion:** Correct the plan to match the actual call: `SalesRecommendationService()` (no session), or change the implementation to pass the session.

.plans/issue-599.md
```diff
@@ -0,0 +1,61 @@
...
+   - No try/catch — `AppException` is handled by the global handler in `main.py:71-77`.
```
yihua-zhuo, line 32: **INFO** (suggestion)

**Issue:** Plan says 'No try/catch — AppException is handled by the global handler', but the router has a manual `if ctx.tenant_id is None: raise HTTPException(status_code=400)` inline error check that bypasses the AppException pattern (Rule 140).

**Suggestion:** Add a note in the plan that the router performs a manual tenant_id validation via HTTPException.

src/api/routers/recommendations.py
```diff
@@ -0,0 +1,40 @@
...
+        raise HTTPException(status_code=400, detail="Tenant ID is required")
+    service = SalesRecommendationService()
+    result = await service.get_recommendations(opportunity_id, ctx.tenant_id)
+    next_action = result.next_best_action
```
yihua-zhuo: **INFO** (suggestion)

**Issue:** Router manually unpacks `RecommendationResult` and its nested `SalesActionRecommendation` into a response dict, duplicating field names inline. This serialization logic belongs on a `.to_dict()` method on the dataclass (Rule 127).

**Suggestion:** Add a `to_dict()` method to `RecommendationResult` and `SalesActionRecommendation`, and have the router call `result.to_dict()`.

src/api/routers/recommendations.py
```diff
@@ -0,0 +1,40 @@
+"""Recommendations router — GET /api/v1/sales/opportunities/{opportunity_id}/recommendations."""
+
+from fastapi import APIRouter, Depends, HTTPException
```
yihua-zhuo: **INFO** (suggestion)

**Issue:** `HTTPException` is imported but only used for the non-`AppException` tenant check; if the check is converted to `ValidationException`, this import becomes dead (Rule 17).

**Suggestion:** Remove the `HTTPException` import after switching to `ValidationException`.

src/services/sales_recommendation.py
```diff
@@ -268,3 +279,28 @@ def predict_conversion_probability(self, opportunity_id: int) -> float:
         conversion_prob = base_prob * market_sentiment * competition_factor
 
         return round(min(max(conversion_prob, 0.0), 1.0), 2)
+
+    async def get_recommendations(self, opportunity_id: int, tenant_id: int) -> RecommendationResult:
```
yihua-zhuo, line 321: **WARNING** (suggestion)

**Issue:** `async def get_recommendations` contains zero `await` statements — it is a wrapper around purely synchronous methods. The `async` signature is misleading and forces the router to `await` a non-async call (Rule 29).

**Suggestion:** Make the method synchronous until real async I/O is introduced, or add a real `await` (e.g., for a DB lookup) to justify the async signature.

src/services/sales_recommendation.py
```diff
@@ -59,8 +69,9 @@ class SalesRecommendationService:
         "enterprise": ["premium"],
     }
 
-    def __init__(self):
-        """初始化服务"""
+    def __init__(self, seed: int | None = None):
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** `__init__` stores `self._seed` but never uses it to seed the random module — the `seed` parameter is effectively dead (Rule 17).

**Suggestion:** Remove the unused `seed` parameter, or actually use it: `self._rng = random.Random(seed)` and replace all `random.*` calls with `self._rng.*`.

tests/unit/test_recommendations_router.py
```diff
@@ -0,0 +1,132 @@
...
+
+        app = _build_app(mock_service)
+        # Override auth to return a context with tenant_id = None
+        app.dependency_overrides[require_auth] = lambda: AuthContext(user_id=99, tenant_id=None, roles=[])
+        client = TestClient(app, raise_server_exceptions=False)
+        resp = client.get("/api/v1/sales/opportunities/1/recommendations")
+        assert resp.status_code == 400
+        mock_service.get_recommendations.assert_not_called()
+
+    def test_missing_auth_returns_401(self, no_auth_client):
```
yihua-zhuo, line 110: **WARNING** (suggestion)

**Issue:** `test_missing_auth_returns_401` uses a standalone `FastAPI()` without overriding `require_auth` and without setting up a JWT secret — the real `require_auth` will execute against missing/invalid token state, producing an unpredictable result depending on the environment (Rule 63).

**Suggestion:** Override `require_auth` with a function that raises `UnauthorizedException`, or use the `no_auth_client` fixture with a known-no-token request that exercises the real handler deterministically.

tests/unit/test_recommendations_router.py
```diff
@@ -0,0 +1,132 @@
...
+def _build_app(mock_service: MagicMock | None) -> FastAPI:
+    """Build a FastAPI app with the router, the global exception handlers from
+    main.py, and dependency overrides. `mock_service=None` means do not patch
+    the service (used for tests that exercise auth only)."""
+    app = FastAPI()
+    app.include_router(recommendations_router)
+    app.dependency_overrides[require_auth] = lambda: _make_auth_ctx()
+    app.dependency_overrides[get_db] = lambda: MagicMock()
+
+    @app.exception_handler(AppException)
```
yihua-zhuo: **INFO** (suggestion)

**Issue:** Exception handlers are manually duplicated from `main.py` in `_build_app`. If the global handlers' response format changes, these tests will silently pass against the old shape and miss the regression (Rule 70).

**Suggestion:** Import the handler functions from `main.py` and register them on the test app, or spin up the full application via `TestClient(main.app)`.

tests/unit/test_recommendations_router.py
```diff
@@ -0,0 +1,132 @@
...
+            content={"success": False, "message": "Internal server error", "code": "INTERNAL_ERROR"},
+        )
+
+    return app
+
+
+@pytest.fixture
+def mocked_service_client(monkeypatch):
+    """Return a TestClient with SalesRecommendationService fully mocked."""
+    mock_service = MagicMock()
```
yihua-zhuo, line 40: **INFO** (suggestion)

**Issue:** Service is mocked with bare `MagicMock`/`AsyncMock` rather than the composable `MockState`/`MockRow`/`MockResult` pattern — acceptable here since no SQL is involved, but worth noting for consistency (Rule 135).

**Suggestion:** No change needed; this is a service-level mock, not a SQL mock, so the composable handler pattern does not apply.

tests/unit/test_recommendations_router.py
```diff
@@ -0,0 +1,132 @@
...
+    def test_internal_error_returns_500(self, mocked_service_client):
+        client, svc = mocked_service_client
+        svc.get_recommendations = AsyncMock(side_effect=RuntimeError("unexpected"))
+        resp = client.get("/api/v1/sales/opportunities/1/recommendations")
+        assert resp.status_code == 500
+        body = resp.json()
+        assert body["success"] is False
+        assert body["code"] == "INTERNAL_ERROR"
+
+    def test_missing_tenant_returns_400(self, monkeypatch):
```
yihua-zhuo: **INFO** (suggestion)

**Issue:** `test_missing_tenant_returns_400` builds its own app and monkeypatches the module differently from the `mocked_service_client` fixture — the patching mechanism is inconsistent and harder to maintain (Rule 66).

**Suggestion:** Extend the `mocked_service_client` fixture to accept an optional `auth_ctx` parameter so the missing-tenant test reuses the same setup path.

.plans/issue-599.md
```diff
@@ -0,0 +1,61 @@
...
+- `tests/unit/test_recommendations_router.py` — new file: 4 unit tests (200 success, 404, 500, 401 auth-required)
+
+## Implementation Steps
+
+1. **Modify `src/services/sales_recommendation.py`:**
+   - Add `from sqlalchemy.ext.asyncio import AsyncSession` to imports.
+   - Change `def __init__(self):` → `def __init__(self, session: AsyncSession | None = None):` and store `self.session = session` (CLAUDE.md requires `AsyncSession` type with no default, but this service is synchronous/mock-based and does not issue SQL — `AsyncSession | None = None` is the dev-plan's explicit choice for backward compatibility).
+   - Append `async def get_recommendations(self, opportunity_id: int, tenant_id: int) -> dict:` after line 270 (end of `predict_conversion_probability`). The method returns a dict with keys: `opportunity_id`, `conversion_probability`, `similar_opportunities` (list of `customer_id`/`current_tier`/`monthly_revenue`), `next_best_action` (action/target/reason/confidence). Calls existing synchronous helpers internally (no IO, safe in async context).
```
yihua-zhuo, line 26: **WARNING** (suggestion)

**Issue:** Plan says the method 'calls existing synchronous helpers internally (no IO, safe in async context)' but the implementation adds an async DB query `_resolve_customer_id` using raw SQL — plan description is misleading (rule 140).

**Suggestion:** Revise step 1 to mention the new `_resolve_customer_id` async DB lookup and the raw SQL `text()` usage.

.plans/issue-599.md
yihua-zhuo, line 44: **WARNING** (suggestion)

**Issue:** Plan claims 'no integration tests — the service uses mock/hashed data, no real DB queries' but the implementation now executes a real SQL query via `_resolve_customer_id`, making the 'no real DB' claim false (rule 140).

**Suggestion:** Revise the Test Plan to note that integration tests may need coverage for the new DB-backed `_resolve_customer_id` path.

src/api/routers/recommendations.py
```diff
@@ -0,0 +1,28 @@
...
+
+@recommendations_router.get("/opportunities/{opportunity_id}/recommendations")
+async def get_opportunity_recommendations(
+    opportunity_id: int,
+    ctx: AuthContext = Depends(require_auth),
+    session: AsyncSession = Depends(get_db),
+):
+    if ctx.tenant_id is None:
+        raise ValidationException("Tenant ID is required")
+    service = SalesRecommendationService(session)
```
yihua-zhuo, line 22: **INFO** (suggestion)

**Issue:** The injected `session` is passed to `SalesRecommendationService(session)` but is also unused in the handler body — minor: could be clearer to only inject what is used (rule 12: function signatures should be precise).

**Suggestion:** No change needed; the session is correctly forwarded to the service per CLAUDE.md's session-injection pattern.

src/services/recommendation_service.py
```diff
@@ -28,7 +28,7 @@ class RecommendationService:
 
     def __init__(self, session: AsyncSession):
```
yihua-zhuo, line 46: **INFO** (suggestion)

**Issue:** `__slots__` is declared but the class holds only two attributes — a premature optimization that adds maintenance friction (rule 120: simplicity preferred).

**Suggestion:** Remove `__slots__` unless profiling shows measurable memory pressure.

src/services/recommendation_service.py
```diff
@@ -28,7 +28,7 @@ class RecommendationService:
...
         self.session = session
-        self._sales_svc = SalesRecommendationService()
+        self._sales_svc = SalesRecommendationService(session)
 
     async def get_recommendations(self, opportunity_id: int, tenant_id: int) -> dict:
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** `get_recommendations` returns a raw `dict` instead of a domain object/dataclass, inconsistent with the sibling `SalesRecommendationService.get_recommendations` which returns `RecommendationResult` (rule 127: services return domain objects).

**Suggestion:** Return the same `RecommendationResult` type or a new `CachedRecommendationResult` dataclass.

src/services/sales_recommendation.py
```diff
@@ -59,14 +90,13 @@ class SalesRecommendationService:
         "enterprise": ["premium"],
     }
 
-    def __init__(self):
-        """初始化服务"""
-        self._customer_cache: dict[int, dict] = {}
+    def __init__(self, session: AsyncSession, seed: int | None = None):
```
yihua-zhuo, line 103: **INFO** (suggestion)

**Issue:** `__init__` now correctly takes `session: AsyncSession` with no default, fixing the CLAUDE.md violation noted in the plan — this is a good correction.

src/services/sales_recommendation.py
```diff
@@ -78,23 +108,32 @@ def _get_mock_customer_data(self, tenant_id: int, customer_id: int) -> dict:
...
             i
             for i in range(1, 1000)
             if int(hashlib.sha256(f"{tenant_id}:{i}".encode()).hexdigest()[:8], 16) % 4 == tier_index
         ][:limit]
-        random.seed()  # 重置随机种子
         return similar
 
+    async def _resolve_customer_id(self, opportunity_id: int, tenant_id: int) -> int:
+        """通过 DB 查询解析商机关联的 customer_id。如果商机不存在则抛出 NotFoundException。"""
+        result = await self.session.execute(
```
yihua-zhuo, line 138: **WARNING** (suggestion)

**Issue:** Raw SQL via `text('SELECT customer_id FROM opportunities ...')` bypasses the ORM model — if `OpportunityModel` column names change, this query silently drifts until runtime (rule 132: SQL construction must stay structured; rule 90: ORM usage does not hide expensive behavior).

**Suggestion:** Use the ORM: `select(OpportunityModel.customer_id).where(OpportunityModel.id == opportunity_id, OpportunityModel.tenant_id == tenant_id)`.

src/services/sales_recommendation.py
```diff
@@ -268,3 +305,27 @@ def predict_conversion_probability(self, opportunity_id: int) -> float:
...
+        然后组装转换概率、相似客户和下一步最佳行动。
+        """
+        customer_id = await self._resolve_customer_id(opportunity_id, tenant_id)
+        conversion_prob = self.predict_conversion_probability(opportunity_id)
+        similar = self.get_similar_customers(tenant_id, customer_id)
+        next_action = self.get_next_best_action(tenant_id, customer_id)
+        return RecommendationResult(
+            opportunity_id=opportunity_id,
+            conversion_probability=conversion_prob,
+            similar_opportunities=[
```
yihua-zhuo, line 334: **INFO** (suggestion)

**Issue:** The `to_dict()` method manually serializes nested objects — fine for a dataclass, but if `SimilarCustomer` also gets a `to_dict()`, this loop should reuse it for consistency (rule 110: no copy-paste duplication).

**Suggestion:** Use `[s.to_dict() for s in similar]` after giving `SimilarCustomer` a `to_dict()` method.

tests/unit/test_recommendations_router.py
```diff
@@ -0,0 +1,125 @@
...
+    """Build a FastAPI app with the router and dependency overrides.
+
+    `auth_ctx=None` means require_auth is NOT overridden (to test 401).
+    Exception handlers come from main.create_app and are registered here as
+    copies — the source of truth is main.create_app's create_app(), so any
+    format change there must be mirrored here.
+    """
+    from main import create_app
+
+    app = create_app()
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** create_app() is called and its result immediately discarded by `app = FastAPI()` on the next line — wasteful and confusing (Rule 17, 36).

**Suggestion:** Remove the first `app = create_app()` call entirely; keep only the exception-handler copy from the second call.

tests/unit/test_recommendations_router.py
```diff
@@ -0,0 +1,125 @@
...
+    # Only include the recommendations router for isolation
+    app = FastAPI()
+    app.include_router(recommendations_router)
+    # Copy exception handlers from main's create_app
```
yihua-zhuo: **WARNING** (suggestion)

**Issue:** Exception handlers are copied from main.create_app by hand, creating hidden coupling — any format change there silently breaks these tests (Rule 112).

**Suggestion:** Extract exception-handler registration into a reusable function in main.py (e.g. `register_exception_handlers(app)`) and import it instead of copying handlers.

tests/unit/test_recommendations_router.py
```diff
@@ -0,0 +1,125 @@
...
+    _main_app = create_app()
+
+    for exc_type, handler in _main_app.exception_handlers.items():
+        app.add_exception_handler(exc_type, handler)
+
+    if auth_ctx is not None:
+        app.dependency_overrides[require_auth] = lambda ctx=auth_ctx: ctx
+    app.dependency_overrides[get_db] = lambda: MagicMock()
```
yihua-zhuo: **INFO** (suggestion)

**Issue:** get_db is overridden with `lambda: MagicMock()`, but the service is already fully monkeypatched — the DB mock is dead wiring (Rule 36).

**Suggestion:** Remove `app.dependency_overrides[get_db] = lambda: MagicMock()` if no router code path actually uses the session.

tests/unit/test_recommendations_router.py
```diff
@@ -0,0 +1,125 @@
...
+
+class TestGetRecommendations:
+    def test_success_returns_200(self, mocked_service_client):
+        client, svc = mocked_service_client
+        svc.get_recommendations = AsyncMock(return_value=_make_recommendation_result())
+        resp = client.get("/api/v1/sales/opportunities/5/recommendations")
+        assert resp.status_code == 200
+        body = resp.json()
+        assert body["success"] is True
+        assert body["message"] == "Recommendations fetched"
```
yihua-zhuo: **INFO** (suggestion)

**Issue:** Asserting the exact message string `"Recommendations fetched"` is brittle — any copy change breaks the test for non-functional reasons (Rule 64, 67).

**Suggestion:** Assert only on the response shape (`success`, `data`) and leave message text to a snapshot or omit the assertion.

tests/unit/test_recommendations_router.py
```diff
@@ -0,0 +1,125 @@
...
+def _build_app(
+    mock_service: MagicMock | None = None,
+    auth_ctx: AuthContext | None = None,
+) -> FastAPI:
+    """Build a FastAPI app with the router and dependency overrides.
+
+    `auth_ctx=None` means require_auth is NOT overridden (to test 401).
+    Exception handlers come from main.create_app and are registered here as
+    copies — the source of truth is main.create_app's create_app(), so any
+    format change there must be mirrored here.
```
yihua-zhuo: **INFO** (suggestion)

**Issue:** Comment explicitly notes 'any format change there must be mirrored here' — this is an acknowledged tech-debt/maintenance burden (Rule 106).

**Suggestion:** Resolve the coupling by extracting shared exception-handler registration as suggested above, then remove the comment.

src/services/recommendation_service.py
```diff
@@ -5,38 +5,55 @@
...
 from db.models.opportunity import OpportunityModel
 from pkg.errors.app_exceptions import NotFoundException
-from services.sales_recommendation import SalesRecommendationService
+from services.sales_recommendation import SalesActionRecommendation, SalesRecommendationService
 
 _CACHE_TTL = 3600.0  # seconds
 
 # Module-level singleton cache — survives across service instances within a process.
-_cache: dict[str, tuple[float, dict]] = {}
+_cache: dict[str, tuple[float, "CachedRecommendationResult"]] = {}
```
yihua-zhuo, line 20: **CRITICAL** (critical)

**Issue:** Module-level mutable `_cache: dict` is global state that persists across requests and never expires entries (TTL is checked on read but stale entries are never evicted) — a memory leak in long-running processes (Rules 28, 36).

**Suggestion:** Use a bounded LRU cache (e.g. `cachetools.LRUCache(maxsize=1024)`) or add periodic cleanup of expired entries on insert.

.plans/issue-599.md
```diff
@@ -0,0 +1,61 @@
+I have all the context needed. Here is the plan:
+
+# Implementation Plan — Issue #599
+
+## Goal
+Wire up the `GET /api/v1/sales/opportunities/{id}/recommendations` HTTP endpoint by creating a new router that calls the existing `SalesRecommendationService`, adding an async `get_recommendations` method to that service, and registering the router (which `api/__init__.py`'s `iter_routers()` discovers automatically).
+
+## Source Contract
+Dev-plan target: `/home/runner/work/agent-job/agent-job/docs/dev-plan/20-sales/0599-wire-up-get-sales-opportunities-id-recommendations-endpoint.md`
```
yihua-zhuo, line 9: **WARNING** (suggestion)

**Issue:** Plan references dev-plan path `docs/dev-plan/20-sales/0599-wire-up-get-sales-opportunities-id-recommendations-endpoint.md` but there's no indication this file was verified to exist in the current repo state.

**Suggestion:** Run `ls docs/dev-plan/20-sales/0599-*` to confirm the dev-plan source still exists before relying on its requirements.

.plans/issue-599.md
```diff
@@ -0,0 +1,61 @@
...
+Template depth: `deep`
+Reading order followed:
+1. `/home/runner/work/agent-job/agent-job/docs/dev-plan/README.md`
+2. `/home/runner/work/agent-job/agent-job/docs/dev-plan/_template-deep.md`
+3. `/home/runner/work/agent-job/agent-job/docs/dev-plan/20-sales/0599-wire-up-get-sales-opportunities-id-recommendations-endpoint.md`
+
+## Affected Files
```
yihua-zhuo, line 16: **WARNING** (suggestion)

**Issue:** Plan targets `src/services/sales_recommendation.py` for modification, but the actual service that owns caching and DB access is `src/services/recommendation_service.py` — the plan's file list is stale and misaligned with the implementation.

**Suggestion:** Update Affected Files to reference `src/services/recommendation_service.py` (the `RecommendationService` class) and remove `src/services/sales_recommendation.py` from the target list.

.plans/issue-599.md
```diff
@@ -0,0 +1,61 @@
...
+   - Test 1 (`test_success_returns_200`): set `svc.get_recommendations = AsyncMock(return_value={...})`; assert 200 + `body["data"]["opportunity_id"] == 5`.
+   - Test 2 (`test_not_found_returns_404`): set side effect to `NotFoundException("Opportunity")`; assert 404.
+   - Test 3 (`test_internal_error_returns_500`): set side effect to `RuntimeError`; assert 500 (caught by generic handler at `main.py:87-93`).
+   - Test 4 (`test_missing_auth_returns_401`): do not override `require_auth`; assert 401.
+
+## Test Plan
+- Unit tests in `tests/unit/`: new `test_recommendations_router.py` covering 200/404/500/401 per dev-plan §5 Step 3
+- Integration tests in `tests/integration/`: none — the service uses mock/hashed data, no real DB queries; dev-plan §3.1 lists no integration test file
+- Dev-plan verification:
+  - `ruff check src/api/routers/recommendations.py src/services/sales_recommendation.py` → 0 errors (§6)
```
yihua-zhuo, line 46: **INFO** (suggestion)

**Issue:** Plan verification lists `ruff check src/api/routers/recommendations.py src/services/sales_recommendation.py` but the dev-plan §6 acceptance should also include `src/services/recommendation_service.py` per Rule 140 (acceptance criteria must name commands that actually apply to involved files).

**Suggestion:** Add `src/services/recommendation_service.py` to the ruff lint/format check commands in the Test Plan.

src/api/routers/recommendations.py
```diff
@@ -0,0 +1,28 @@
+"""Recommendations router — GET /api/v1/sales/opportunities/{opportunity_id}/recommendations."""
+
+from fastapi import APIRouter, Depends
+from sqlalchemy.ext.asyncio import AsyncSession
+
+from db.connection import get_db
+from internal.middleware.fastapi_auth import AuthContext, require_auth
+from pkg.errors.app_exceptions import ValidationException
```
yihua-zhuo, line 8: **INFO** (suggestion)

**Issue:** `ValidationException` is imported but only used for the defensive tenant check above; if the check is removed, the import becomes dead code.

**Suggestion:** Remove the import if the tenant guard is eliminated.

src/api/routers/recommendations.py
```diff
@@ -0,0 +1,28 @@
...
+from services.sales_recommendation import SalesRecommendationService
```
yihua-zhuo, line 9: **WARNING** (suggestion)

**Issue:** Router instantiates `SalesRecommendationService` directly, bypassing the cached `RecommendationService` in `src/services/recommendation_service.py` which has the TTL cache and tenant-scoped DB lookup. The cache layer is never hit, defeating its purpose (Rule 35).

**Suggestion:** Use `RecommendationService(session)` from `src/services/recommendation_service.py` to benefit from the 3600s TTL cache and proper tenant-scoped opportunity lookup.

src/api/routers/recommendations.py
```diff
@@ -0,0 +1,28 @@
+"""Recommendations router — GET /api/v1/sales/opportunities/{opportunity_id}/recommendations."""
```
yihua-zhuo, line 1: **INFO** (suggestion)

**Issue:** Router module docstring is fine, but the file is auto-discovered by `iter_routers()` — per Rule 146, it must be import-safe (no side effects on import). Currently it is, which is correct.

**Suggestion:** No change needed; the module is import-safe.

src/main.py
```diff
@@ -34,6 +34,38 @@ async def lifespan(app: FastAPI):
...
+    Extracted from create_app so tests can apply the same handlers without
+    duplicating the JSON envelopes here.
+    """
+
+    @app.exception_handler(AppException)
+    async def app_exception_handler(request: Request, exc: AppException) -> JSONResponse:
+        logger.error("app_exception", code=exc.code, detail=exc.detail, path=request.url.path)
+        return JSONResponse(
+            status_code=exc.status_code,
+            content={"success": False, "message": exc.detail, "code": exc.code},
```
yihua-zhuo, line 49: **INFO** (suggestion)

**Issue:** Generic 500 handler does not include a request ID or correlation ID, making it hard to correlate logs with user-reported errors (Rule 41).

**Suggestion:** Include `request_id` (from middleware or a new dependency) in the error response and log payload.

src/services/recommendation_service.py
```diff
@@ -5,38 +5,55 @@
...
+class RecommendationService:
     def __init__(self, session: AsyncSession):
         self.session = session
-        self._sales_svc = SalesRecommendationService()
+        self._sales_svc = SalesRecommendationService(session)
 
-    async def get_recommendations(self, opportunity_id: int, tenant_id: int) -> dict:
+    async def get_recommendations(self, opportunity_id: int, tenant_id: int) -> CachedRecommendationResult:
         key = _cache_key(opportunity_id, tenant_id)
         now = time.time()
```
yihua-zhuo, line 52: **WARNING** (suggestion)

**Issue:** `time.time()` is used instead of `time.monotonic()` — system clock changes (NTP sync, DST) can cause cache entries to live longer or shorter than the intended TTL (Rule 143 adjacent: clock safety).

**Suggestion:** Use `time.monotonic()` for TTL comparisons.

src/services/recommendation_service.py
```diff
@@ -5,38 +5,55 @@
...
         if key in _cache:
-            ts, data = _cache[key]
+            ts, cached = _cache[key]
             if now - ts < _CACHE_TTL:
-                return data
+                return cached
         result = await self.session.execute(
```
yihua-zhuo, line 57: **WARNING** (suggestion)

**Issue:** Cache is populated but `invalidate_cache()` is the only eviction path; if the underlying data changes (e.g. opportunity is updated), stale results are served for up to 3600s. No cache-invalidation hooks are wired in (Rule 35).

**Suggestion:** Wire `invalidate_cache()` into the opportunity update/delete paths, or reduce TTL to a value matched to data freshness requirements.

src/services/recommendation_service.py
```diff
@@ -46,12 +63,12 @@ async def get_recommendations(self, opportunity_id: int, tenant_id: int) -> dict
...
-            "opportunity_id": opportunity_id,
-            "conversion_probability": self._sales_svc.predict_conversion_probability(opportunity_id),
-            "next_action": self._sales_svc.get_next_best_action(tenant_id, opp.customer_id),
-            "similar_deals": [],
-        }
+        data = CachedRecommendationResult(
+            opportunity_id=opportunity_id,
+            conversion_probability=self._sales_svc.predict_conversion_probability(opportunity_id),
+            next_action=self._sales_svc.get_next_best_action(tenant_id, opp.customer_id),
+            similar_deals=[],
```
yihua-zhuo, line 70: **WARNING** (suggestion)

**Issue:** `similar_deals` is always set to an empty list — the service never calls `self._sales_svc.get_similar_customers()`, so the similar-opportunities feature is dead/missing despite the cache and response envelope supporting it (Rule 17).

**Suggestion:** Populate `similar_deals` via `self._sales_svc.get_similar_customers(tenant_id, opp.customer_id)` or document explicitly that this field is reserved for future use.

src/services/recommendation_service.py
```diff
@@ -5,38 +5,55 @@
...
+@dataclass
+class CachedRecommendationResult:
+    """Cached recommendation result returned by RecommendationService.get_recommendations."""
+
+    opportunity_id: int
+    conversion_probability: float
+    next_action: SalesActionRecommendation
+    similar_deals: list[dict] = field(default_factory=list)
+
+    def to_dict(self) -> dict:
```
yihua-zhuo, line 36: **INFO** (suggestion)

**Issue:** `to_dict()` output keys (`next_action`, `similar_deals`) differ from the `RecommendationResult.to_dict()` keys in `sales_recommendation.py:76` (`next_best_action`, `similar_opportunities`) — two different response shapes for the same conceptual endpoint will confuse API consumers (Rule 91).

**Suggestion:** Align the key names across both DTOs to a single contract, e.g. always `next_best_action` and `similar_opportunities`.

src/services/sales_recommendation.py
```diff
@@ -78,23 +118,34 @@ def _get_mock_customer_data(self, tenant_id: int, customer_id: int) -> dict:
...
             "satisfaction_score": (seed % 100) / 100.0,
         }
 
     def _get_similar_customers_by_tier(self, tenant_id: int, tier: str, limit: int = 5) -> list[int]:
         """基于层级获取相似客户ID"""
-        random.seed(42)  # 固定随机种子保证一致性
         tier_index = list(self.PRODUCTS.keys()).index(tier)
         similar = [
             i
             for i in range(1, 1000)
```
yihua-zhuo, line 131: **WARNING** (suggestion)

**Issue:** Magic number 1000 in `range(1, 1000)` and 10 as a scan limit — no constant or comment explains the choice (Rule 19).

**Suggestion:** Extract `_MAX_SCAN_RANGE = 1000` and `_SIMILAR_SCAN_LIMIT = 10` as module-level constants.

src/services/sales_recommendation.py
```diff
@@ -130,7 +181,7 @@ def get_next_best_action(self, tenant_id: int, customer_id: int) -> SalesActionR
             action=action,
             target=target,
             reason=reason,
-            confidence=round(random.uniform(0.6, 0.95), 2),  # noqa: S311 - non-security recommendation scoring
+            confidence=round(self._rng.uniform(0.6, 0.95), 2),  # noqa: S311 - non-security recommendation scoring
```
yihua-zhuo, line 184: **INFO** (suggestion)

**Issue:** Confidence range 0.6–0.95 is a magic number pair (Rule 19).

**Suggestion:** Extract `_CONFIDENCE_MIN = 0.6`, `_CONFIDENCE_MAX = 0.95` as module constants.

tests/unit/test_recommendations_router.py
```diff
@@ -0,0 +1,113 @@
...
+
+    if auth_ctx is not None:
+        app.dependency_overrides[require_auth] = lambda ctx=auth_ctx: ctx
+    return app
+
+
+@pytest.fixture
+def mocked_service_client(monkeypatch):
+    """Return a TestClient with SalesRecommendationService fully mocked."""
+    mock_service = MagicMock()
```
yihua-zhuo, line 40: **INFO** (suggestion)

**Issue:** Uses a heavy `MagicMock()` for the entire service instead of a narrow stub with only the methods exercised — acceptable for router tests but could be tightened (Rule 66).

**Suggestion:** Consider using a small `SimpleNamespace` with only `get_recommendations` set to an `AsyncMock`, making the contract being tested explicit.

tests/unit/test_recommendations_router.py
```diff
@@ -0,0 +1,113 @@
...
+    monkeypatch.setattr(
+        "api.routers.recommendations.SalesRecommendationService",
```
yihua-zhuo, line 42: **INFO** (suggestion)

**Issue:** Lambda `lambda session=None: mock_service` has a mutable-style default argument pattern (session=None) — works for a mock but doesn't match the project's rule that service constructors must not allow session=None (Rule 121).

**Suggestion:** Use `lambda session: mock_service` since the router always passes a session, and this avoids implying None is acceptable.

tests/unit/test_recommendations_router.py
```diff
@@ -0,0 +1,113 @@
...
+    def test_internal_error_returns_500(self, mocked_service_client):
+        client, svc = mocked_service_client
+        svc.get_recommendations = AsyncMock(side_effect=RuntimeError("unexpected"))
+        resp = client.get("/api/v1/sales/opportunities/1/recommendations")
+        assert resp.status_code == 500
+        body = resp.json()
+        assert body["success"] is False
+        assert body["code"] == "INTERNAL_ERROR"
+
+    def test_missing_tenant_returns_422(self, mocked_service_client):
```
yihua-zhuo, line 101: **INFO** (suggestion)

**Issue:** No edge-case tests for invalid opportunity_id values (e.g. negative, zero, very large) — router-level validation paths are untested (Rule 49).

**Suggestion:** Add tests for opportunity_id=0, -1, and a very large number to confirm FastAPI's int path parameter handles them.

tests/unit/test_recommendations_router.py
```diff
@@ -0,0 +1,113 @@
...
+            reason="高使用率",
+            confidence=0.85,
+        ),
+    )
+
+
+class TestGetRecommendations:
+    def test_success_returns_200(self, mocked_service_client):
+        client, svc = mocked_service_client
+        svc.get_recommendations = AsyncMock(return_value=_make_recommendation_result())
```
yihua-zhuo, line 77: **INFO** (suggestion)

**Issue:** Good coverage of success, 404, 500, 422, and 401 paths — test names are clear and follow the `test_<scenario>_<expected_outcome>` convention (Rule 65, Rule 67).

**Suggestion:** No change needed — this is a well-structured router test suite.

PR #823, opened 2026-06-07
develop → master

yihua-zhuo:
Auto-created by CI from branch `develop`.

### DIFF
--- a/frontend/next.config.ts
+++ b/frontend/next.config.ts
@@ -1,6 +1,10 @@
 import type { NextConfig } from "next";
+import path from "path";
 
 const nextConfig: NextConfig = {
+  turbopack: {
+    root: path.join(__dirname),
+  },
   async rewrites() {
     if (!process.env.NEXT_PUBLIC_BACKEND_ORIGIN) {
       throw new Error(
--- a/frontend/src/app/(app)/customers/page.tsx
+++ b/frontend/src/app/(app)/customers/page.tsx
@@ -1,5 +1,5 @@
 "use client";
-import { useState, useRef, useCallback, useEffect, Suspense } from "react";
+import { useState, useRef, useCallback, useEffect, useMemo, Suspense } from "react";
 import { useSearchParams } from "next/navigation";
 import { useTableState } from "@/lib/hooks/useTableState";
 import { ColumnDef } from "@tanstack/react-table";
@@ -345,6 +345,8 @@ interface CreateForm {
 
 const blankCreateForm: CreateForm = { name: "", email: "", phone: "", company: "", status: "lead" };
 
+const CUSTOMER_SEARCHABLE_KEYS = ["name", "email", "phone"];
+
 function CustomersPageInner() {
   const searchParams = useSearchParams();
 
@@ -368,7 +370,10 @@ function CustomersPageInner() {
   const [showDelete, setShowDelete] = useState(false);
   const deleteCustomer = useDeleteCustomer();
 
-  const [savedViews, setSavedViews] = useState<SavedView[]>(() => loadViews());
+  const [savedViews, setSavedViews] = useState<SavedView[]>([]);
+  useEffect(() => {
+    setSavedViews(loadViews());
+  }, []);
   const [showSaveDialog, setShowSaveDialog] = useState(false);
   const [newViewName, setNewViewName] = useState("");
 
@@ -390,33 +395,38 @@ function CustomersPageInner() {
 
   const { data, isLoading, isError, refetch, isFetching } = useCustomers(page, pageSize);
 
+  const refetchRef = useRef(refetch);
+  refetchRef.current = refetch;
   useEffect(() => {
-    if (!refetch) return;
-    const interval = setInterval(() => { refetch(); }, 30_000);
+    const interval = setInterval(() => { refetchRef.current?.(); }, 30_000);
     return () => clearInterval(interval);
-  }, [refetch]);
+  }, []);
 
   const { widths, onMouseDown } = useColumnResize({
     name: 160, email: 200, phone: 140, status: 120, company: 160, created_at: 120,
   });
 
-  const rawItems = data?.data?.items ?? [];
+  const rawItems = data?.data?.items;
   const info = data?.data;
 
-  const items: CustomerRowData[] = rawItems.map((c) => ({
-    id: Number(c.id),
-    name: String(c.name ?? ""),
-    email: String(c.email ?? ""),
-    phone: String(c.phone ?? ""),
-    status: String(c.status ?? ""),
-    company: String(c.company ?? ""),
-    created_at: String(c.created_at ?? ""),
-  }));
+  const items: CustomerRowData[] = useMemo(
+    () =>
+      (rawItems ?? []).map((c) => ({
+        id: Number(c.id),
+        name: String(c.name ?? ""),
+        email: String(c.email ?? ""),
+        phone: String(c.phone ?? ""),
+        status: String(c.status ?? ""),
+        company: String(c.company ?? ""),
+        created_at: String(c.created_at ?? ""),
+      })),
+    [rawItems]
+  );
 
   const { table, globalFilter, setGlobalFilter } = useTableState({
     data: items,
     columns: customerColumns,
-    searchableKeys: ["name", "email", "phone"],
+    searchableKeys: CUSTOMER_SEARCHABLE_KEYS,
   });
 
   const clearSearch = useCallback(() => {
--- a/frontend/src/app/(app)/layout.tsx
+++ b/frontend/src/app/(app)/layout.tsx
@@ -17,9 +17,12 @@ const STORAGE_KEY = "crm_sidebar_collapsed";
 
 export default function AppLayout({ children }: { children: React.ReactNode }) {
   const [sidebarOpen, setSidebarOpen] = useState(false);
-  const [collapsed, setCollapsed] = useState(() => {
-    try { return localStorage.getItem(STORAGE_KEY) === "true"; } catch { return false; }
-  });
+  const [collapsed, setCollapsed] = useState(false);
+  const [collapsedHydrated, setCollapsedHydrated] = useState(false);
+  useEffect(() => {
+    try { setCollapsed(localStorage.getItem(STORAGE_KEY) === "true"); } catch {}
+    setCollapsedHydrated(true);
+  }, []);
   const pathname = usePathname();
   const quickAdd = useQuickAddTask();
   const create = useCreateTask();
@@ -34,8 +37,9 @@ export default function AppLayout({ children }: { children: React.ReactNode }) {
   }
 
   useEffect(() => {
+    if (!collapsedHydrated) return;
     try { localStorage.setItem(STORAGE_KEY, String(collapsed)); } catch {}
-  }, [collapsed]);
+  }, [collapsed, collapsedHydrated]);
 
   return (
     <AuthGuard>
--- a/frontend/src/app/(auth)/login/page.tsx
+++ b/frontend/src/app/(auth)/login/page.tsx
@@ -4,16 +4,15 @@ import { zodResolver } from "@hookform/resolvers/zod";
 import { z } from "zod";
 import { useRouter } from "next/navigation";
 import { useAuthStore } from "@/lib/store/auth-store";
-import { Providers } from "@/lib/components/providers";
 import { Input } from "@/components/ui/input";
 
 const loginSchema = z.object({
-  username: z.string().min(1, "Username is required").email("Please enter a valid email address"),
+  username: z.string().min(1, "Username is required"),
   password: z.string().min(1, "Password is required"),
 });
 type LoginForm = z.infer<typeof loginSchema>;
 
-function LoginForm() {
+export default function LoginPage() {
   const router = useRouter();
   const form = useForm<LoginForm>({
     resolver: zodResolver(loginSchema),
@@ -27,7 +26,7 @@ function LoginForm() {
   const onSubmit = form.handleSubmit(async (values) => {
     try {
       await login({ username: values.username, password: values.password });
-      router.push("/dashboard");
+      router.push("/");
     } catch {
       // error is exposed via the store; nothing more to do here
     }
@@ -78,10 +77,3 @@ function LoginForm() {
   );
 }
 
-export default function LoginPage() {
-  return (
-    <Providers>
-      <LoginForm />
-    </Providers>
-  );
-}
--- a/frontend/src/lib/components/auth-guard.tsx
+++ b/frontend/src/lib/components/auth-guard.tsx
@@ -1,17 +1,21 @@
 "use client";
 import { useRouter } from "next/navigation";
-import { useEffect } from "react";
+import { useEffect, useState } from "react";
 import { useAuthStore } from "@/lib/store/auth-store";
 
 export function AuthGuard({ children }: { children: React.ReactNode }) {
   const router = useRouter();
-  const isHydrated = useAuthStore((s) => s.isHydrated);
-  const isAuthenticated = useAuthStore((s) => s.isAuthenticated());
+  const [mounted, setMounted] = useState(false);
+  const isAuthenticated = useAuthStore((s) => s.isAuthenticated);
 
   useEffect(() => {
-    if (isHydrated && !isAuthenticated) router.push("/login");
-  }, [isHydrated, isAuthenticated, router]);
+    setMounted(true);
+  }, []);
 
-  if (!isHydrated || !isAuthenticated) return null;
+  useEffect(() => {
+    if (mounted && !isAuthenticated()) router.push("/login");
+  }, [mounted, isAuthenticated, router]);
+
+  if (!mounted || !isAuthenticated()) return null;
   return <>{children}</>;
 }
--- a/frontend/src/lib/components/offline-banner.tsx
+++ b/frontend/src/lib/components/offline-banner.tsx
@@ -4,9 +4,11 @@ import { WifiOff } from "lucide-react";
 import { cn } from "@/lib/utils";
 
 export function OfflineBanner() {
-  const [isOffline, setIsOffline] = useState(() => !navigator.onLine);
+  const [isOffline, setIsOffline] = useState(false);
 
   useEffect(() => {
+    setIsOffline(!navigator.onLine);
+
     function onOnline() { setIsOffline(false); }
     function onOffline() { setIsOffline(true); }
 
--- a/frontend/src/lib/components/session-timeout-guard.tsx
+++ b/frontend/src/lib/components/session-timeout-guard.tsx
@@ -14,9 +14,11 @@ import { Button } from "@/components/ui/button";
 
 const IDLE_TIMEOUT_MS = 25 * 60 * 1000;   // 25 min — warning appears
 const LOGOUT_TIMEOUT_MS = 30 * 60 * 1000; // 30 min — forced logout
+const ACTIVITY_THROTTLE_MS = 1000;        // re-arm timers at most once per second
 
 export function SessionTimeoutGuard() {
-  const { isAuthenticated, clearAuth } = useAuthStore();
+  const isAuthenticated = useAuthStore((s) => s.isAuthenticated);
+  const clearAuth = useAuthStore((s) => s.clearAuth);
   const router = useRouter();
   const [showWarning, setShowWarning] = useState(false);
   const idleTimerRef = useRef<ReturnType<typeof setTimeout> | null>(null);
@@ -38,7 +40,13 @@ export function SessionTimeoutGuard() {
     if (!isAuthenticated()) return;
     scheduleWarning();
 
-    function onActivity() { scheduleWarning(); }
+    let lastActivityAt = 0;
+    function onActivity() {
+      const now = performance.now();
+      if (now - lastActivityAt < ACTIVITY_THROTTLE_MS) return;
+      lastActivityAt = now;
+      scheduleWarning();
+    }
     window.addEventListener("mousemove", onActivity);
     window.addEventListener("mousedown", onActivity);
     window.addEventListener("keydown", onActivity);
--- a/scripts/dev/seed_admin_user.py
+++ b/scripts/dev/seed_admin_user.py
@@ -7,12 +7,13 @@ the bcrypt password hash matches what login expects) and flips status to
 ``active`` so the OAuth2 password flow succeeds immediately.
 
 Env vars (all optional, defaults are for LOCAL DEV ONLY):
-    DATABASE_URL    Required. e.g. ``postgresql+asyncpg://test_user:test_pass@localhost:5432/test_db``
-    SEED_USERNAME   default ``admin``
-    SEED_EMAIL      default ``admin@example.com``
-    SEED_PASSWORD   default ``Admin12345``  (must meet UserService validator)
-    SEED_ROLE       default ``admin``
-    SEED_TENANT_ID  default ``1``
+    DATABASE_URL      Required. e.g. ``postgresql+asyncpg://test_user:test_pass@localhost:5432/test_db``
+    SEED_USERNAME     default ``admin``
+    SEED_EMAIL        default ``admin@example.com``
+    SEED_PASSWORD     default ``Admin12345``  (must meet UserService validator)
+    SEED_ROLE         default ``admin``
+    SEED_TENANT_ID    default ``1``
+    SEED_TENANT_NAME  default ``Default Tenant``
 
 Designed to be invoked from the Makefile after ``db-up`` + ``migrate``.
 """
@@ -24,9 +25,10 @@ import os
 import sys
 
 # scripts/dev/ runs with PYTHONPATH=src set by the Makefile target.
-from sqlalchemy import and_, select
+from sqlalchemy import and_, select, text
 from sqlalchemy.ext.asyncio import async_sessionmaker, create_async_engine
 
+from db.models.tenant import TenantModel
 from db.models.user import UserModel
 from services.auth_service import AuthService
 
@@ -47,6 +49,7 @@ async def main() -> int:
     password = env("SEED_PASSWORD", "Admin12345")
     role = env("SEED_ROLE", "admin")
     tenant_id = int(env("SEED_TENANT_ID", "1"))
+    tenant_name = env("SEED_TENANT_NAME", "Default Tenant")
 
     # AuthService requires a non-empty secret. The seed script doesn't issue
     # any tokens — it only needs the constructor to succeed — so pass a
@@ -58,6 +61,8 @@ async def main() -> int:
 
     try:
         async with session_factory() as session:
+            await _ensure_tenant(session, tenant_id, tenant_name)
+
             existing = await session.execute(
                 select(UserModel).where(
                     and_(UserModel.username == username, UserModel.tenant_id == tenant_id)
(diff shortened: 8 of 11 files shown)


frontend/src/app/(app)/customers/page.tsx
```diff
@@ -345,6 +345,8 @@ interface CreateForm {
 
 const blankCreateForm: CreateForm = { name: "", email: "", phone: "", company: "", status: "lead" };
 
+const CUSTOMER_SEARCHABLE_KEYS = ["name", "email", "phone"];
+
 function CustomersPageInner() {
```
yihua-zhuo, line 350: **CRITICAL** (critical)

**Issue:** `CustomersPageInner` is ~670 lines and mixes column-resize logic, CSV export, saved-view serialization, polling, table rendering, and four modal dialogs in a single component (Rule 5, 18).

**Suggestion:** Extract into focused modules: `useColumnResize` hook (already extracted), `SavedViews` component/hook, `CustomerTable` component, `BulkActionsToolbar` component, and `exportCsv(rows)` utility.

frontend/src/lib/components/auth-guard.tsx
```diff
@@ -1,17 +1,21 @@
...
-import { useEffect } from "react";
+import { useEffect, useState } from "react";
 import { useAuthStore } from "@/lib/store/auth-store";
 
 export function AuthGuard({ children }: { children: React.ReactNode }) {
   const router = useRouter();
-  const isHydrated = useAuthStore((s) => s.isHydrated);
-  const isAuthenticated = useAuthStore((s) => s.isAuthenticated());
+  const [mounted, setMounted] = useState(false);
+  const isAuthenticated = useAuthStore((s) => s.isAuthenticated);
```
yihua-zhuo, line 9: **CRITICAL** (critical)

**Issue:** `useAuthStore((s) => s.isAuthenticated)` selects the function reference from the store, not the boolean value it returns — since function references are stable, the selector never reports a change and the component will not re-render when auth state flips (e.g. token expiry, external logout). The effect on line 15 likewise only runs on mount.

**Suggestion:** Either store `isAuthenticated` as a boolean property on the Zustand state and select it directly, or call the function in the selector: `useAuthStore((s) => s.isAuthenticated())`. If `isAuthenticated` must remain a method, subscribe to the underlying state (e.g. `s.token`) instead.

frontend/src/lib/components/session-timeout-guard.tsx
```diff
@@ -38,7 +40,13 @@ export function SessionTimeoutGuard() {
     if (!isAuthenticated()) return;
```
yihua-zhuo, line 40: **CRITICAL** (critical)

**Issue:** `isAuthenticated()` is called as a function, but the Zustand selector on line 20 reads `s.isAuthenticated` — if it is a boolean property (the common pattern), this throws TypeError at runtime and the guard never activates.

**Suggestion:** Verify the auth store's type. If `isAuthenticated` is a boolean, use `if (!isAuthenticated) return;` (no parens). If it is a method, add a type annotation or rename to `checkAuth()` to disambiguate.

frontend/next.config.ts
```diff
@@ -1,6 +1,10 @@
 import type { NextConfig } from "next";
+import path from "path";
 
 const nextConfig: NextConfig = {
+  turbopack: {
+    root: path.join(__dirname),
```
yihua-zhuo, line 6: **WARNING** (suggestion)

**Issue:** Using `__dirname` in an ESM context can fail at startup — Next.js 15+ compiles `next.config.ts` in a mixed/ESM module context where `__dirname` may not be defined.

**Suggestion:** Use `import.meta.url` with `fileURLToPath` instead: `import { fileURLToPath } from "url"; const root = path.dirname(fileURLToPath(import.meta.url));`

frontend/next.config.ts
```diff
@@ -1,6 +1,10 @@
...
+  },
   async rewrites() {
     if (!process.env.NEXT_PUBLIC_BACKEND_ORIGIN) {
```
yihua-zhuo, line 9: **INFO** (suggestion)

**Issue:** Failing fast on a missing env var is good, but the error message could include the expected variable name more prominently for ops visibility.

**Suggestion:** No code change needed — the current message is clear. Just ensure this is documented in onboarding/deploy runbooks (Rule 77).

frontend/src/app/(app)/customers/page.tsx
```diff
@@ -390,33 +395,38 @@ function CustomersPageInner() {
 
   const { data, isLoading, isError, refetch, isFetching } = useCustomers(page, pageSize);
 
+  const refetchRef = useRef(refetch);
+  refetchRef.current = refetch;
   useEffect(() => {
-    if (!refetch) return;
-    const interval = setInterval(() => { refetch(); }, 30_000);
+    const interval = setInterval(() => { refetchRef.current?.(); }, 30_000);
```
yihua-zhuo, line 401: **INFO** (suggestion)

**Issue:** Magic number `30_000` (30s polling interval) is hardcoded with no configuration (Rule 19).

**Suggestion:** Extract `const POLL_INTERVAL_MS = 30_000;` at module top, or move to a config constant.

frontend/src/app/(app)/layout.tsx
```diff
@@ -17,9 +17,12 @@ const STORAGE_KEY = "crm_sidebar_collapsed";
 
 export default function AppLayout({ children }: { children: React.ReactNode }) {
   const [sidebarOpen, setSidebarOpen] = useState(false);
-  const [collapsed, setCollapsed] = useState(() => {
-    try { return localStorage.getItem(STORAGE_KEY) === "true"; } catch { return false; }
-  });
+  const [collapsed, setCollapsed] = useState(false);
+  const [collapsedHydrated, setCollapsedHydrated] = useState(false);
+  useEffect(() => {
+    try { setCollapsed(localStorage.getItem(STORAGE_KEY) === "true"); } catch {}
```
yihua-zhuo, line 23: **INFO** (suggestion)

**Issue:** Silent `try/catch {}` around `localStorage` access — errors are swallowed without logging even for debugging (Rule 46, 41).

**Suggestion:** At minimum log: `catch (e) { console.warn("localStorage unavailable", e); }`

frontend/src/app/(app)/layout.tsx
```diff
@@ -17,9 +17,12 @@ const STORAGE_KEY = "crm_sidebar_collapsed";
...
+    setCollapsedHydrated(true);
+  }, []);
   const pathname = usePathname();
   const quickAdd = useQuickAddTask();
   const create = useCreateTask();
```
yihua-zhuo, line 28: **INFO** (suggestion)

**Issue:** Variable names `create` and `update` are too vague for domain-specific task mutations (Rule 11).

**Suggestion:** Rename to `createTask` and `updateTask` to match the hook names (`useCreateTask`, `useUpdateTask`).

frontend/src/app/(auth)/login/page.tsx
```diff
@@ -27,7 +26,7 @@ function LoginForm() {
   const onSubmit = form.handleSubmit(async (values) => {
     try {
       await login({ username: values.username, password: values.password });
-      router.push("/dashboard");
+      router.push("/");
     } catch {
       // error is exposed via the store; nothing more to do here
```
yihua-zhuo, line 31: **WARNING** (suggestion)

**Issue:** The `catch` block is empty — the comment says the error is exposed via the store, but this silently swallows the rejection if the store does not set `error` before re-throwing, violating the no-silent-swallowing guideline.

**Suggestion:** Guard the call with an explicit check: `if (useAuthStore.getState().error) { /* show from store */ } else { console.error(e); }`, or simply remove the try/catch and let the error propagate.

frontend/src/lib/components/auth-guard.tsx
```diff
@@ -1,17 +1,21 @@
...
-  }, [isHydrated, isAuthenticated, router]);
+    setMounted(true);
+  }, []);
 
-  if (!isHydrated || !isAuthenticated) return null;
+  useEffect(() => {
+    if (mounted && !isAuthenticated()) router.push("/login");
+  }, [mounted, isAuthenticated, router]);
+
+  if (!mounted || !isAuthenticated()) return null;
```
yihua-zhuo, line 19: **WARNING** (suggestion)

**Issue:** Returning `null` while unauthenticated causes a brief flash of empty content and possible layout shift on protected pages, and — because of the selector bug above — the guard may never redirect at all if the user was authenticated on mount and logs out later.

**Suggestion:** Return a loading skeleton (`<div className="flex h-screen items-center justify-center"><Spinner /></div>`) while the initial auth check resolves, so users see a stable layout.

frontend/src/lib/components/offline-banner.tsx
```diff
@@ -4,9 +4,11 @@ import { WifiOff } from "lucide-react";
 import { cn } from "@/lib/utils";
 
 export function OfflineBanner() {
-  const [isOffline, setIsOffline] = useState(() => !navigator.onLine);
+  const [isOffline, setIsOffline] = useState(false);
```
yihua-zhuo, line 7: **INFO** (suggestion)

**Issue:** `useState(false)` always renders the banner as collapsed on the server and first client paint; if the user is offline on page load, the banner flashes from hidden to visible after hydration.

**Suggestion:** Initialize from a safe sync check: `const [isOffline, setIsOffline] = useState(() => typeof navigator !== 'undefined' && !navigator.onLine);` so the first client render already reflects the correct state.

frontend/src/lib/components/session-timeout-guard.tsx
```diff
@@ -14,9 +14,11 @@ import { Button } from "@/components/ui/button";
 
 const IDLE_TIMEOUT_MS = 25 * 60 * 1000;   // 25 min — warning appears
 const LOGOUT_TIMEOUT_MS = 30 * 60 * 1000; // 30 min — forced logout
+const ACTIVITY_THROTTLE_MS = 1000;        // re-arm timers at most once per second
 
 export function SessionTimeoutGuard() {
```
yihua-zhuo, line 19: **INFO** (suggestion)

**Issue:** The component is named 'Guard' but also renders a visible Dialog — consider naming it `SessionTimeoutDialog` or extracting the timer logic into a hook consumed by a presentational component.

**Suggestion:** Split into `useIdleTimer` hook and `SessionTimeoutDialog` for single responsibility (rule 4).

scripts/dev/seed_admin_user.py
```diff
@@ -96,6 +101,28 @@ async def main() -> int:
...
+    await session.flush()
+    # Explicit id insert bypasses the sequence; align it so future
+    # auto-generated tenants don't collide on the primary key.
+    await session.execute(
+        text(
+            "SELECT setval(pg_get_serial_sequence('tenants', 'id'), "
+            "GREATEST((SELECT MAX(id) FROM tenants), 1))"
+        )
+    )
+    await session.commit()
```
yihua-zhuo, line 122: **WARNING** (suggestion)

**Issue:** `_ensure_tenant` calls `session.commit()` independently of the main user-creation transaction — if `create_user` fails afterwards, the tenant row persists in a half-seeded state, and the two-commit pattern diverges from the project rule that services never commit (rule 121).

**Suggestion:** Remove the inner `session.commit()` and `session.flush()` in `_ensure_tenant`; let the outer scope commit once after both tenant and user are staged, or restructure so `_ensure_tenant` only does `session.add()` + `flush()`.

yihua-zhuo:

**Status:** `fail`
**Inline comments posted:** 13

**Summary:** Reviewed 11 file(s) between origin/master and origin/develop.

`frontend/src/app/(app)/customers/page.tsx` L706: 
**CRITICAL** (critical)

**Issue:** Six near-identical `<th>` blocks (lines 706-813) copy-paste the sort button, sort indicator, and resize handle with only the column id changing — ~110 lines of duplication (Rule 110).

**Suggestion:** Extract a `<SortableHeader column={col} label={col} widths={widths} onResize={onMouseDown} />` component and map over `["name","email","phone","status","company","created_at"]`. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`frontend/src/app/(app)/customers/page.tsx` L521: 
**WARNING** (suggestion)

**Issue:** `mutateAsync(createForm as unknown as Record<string, unknown>)` bypasses the typed mutation contract — if the backend expects different field names, this won't be caught at compile time (Rule 27).

**Suggestion:** Define a `CreateCustomerPayload` type and pass `createForm` directly. If the hook expects a different shape, add a typed adapter function instead of a double cast. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`frontend/src/app/(app)/customers/page.tsx` L662: 
**WARNING** (suggestion)

**Issue:** CSV export wraps every value in double-quotes but does not escape embedded `"` characters, producing malformed CSV when a customer name, company, or status contains a quote (Rule 49).

**Suggestion:** Escape values: `.map((v) => `"${String(v).replace(/"/g, '""')}"`)`. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`frontend/src/app/(app)/customers/page.tsx` L513: 
**WARNING** (suggestion)

**Issue:** `handleDelete` and `handleCreate` catch blocks discard the error object entirely — toast says "Failed to delete customer" with no error context for debugging (Rule 41, 46).

**Suggestion:** Pass the error: `catch (err) { toast.error(`Failed: ${err instanceof Error ? err.message : "unknown"}`); console.error(err); }` — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`frontend/src/app/(app)/customers/page.tsx` L818: 
**INFO** (suggestion)

**Issue:** Skeleton row count `6` is a magic number (Rule 19).

**Suggestion:** Extract `const SKELETON_ROW_COUNT = pageSize;` or a constant. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`frontend/src/app/(app)/customers/page.tsx` L1022: 
**INFO** (suggestion)

**Issue:** `CustomersPage` wrapper and `CustomersPageInner` split is a Suspense boundary, but the naming is uninformative (Rule 12).

**Suggestion:** Rename `CustomersPageInner` to `CustomersTableView` or similar to describe the domain, not the nesting. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`frontend/src/app/(app)/customers/page.tsx` L976: 
**WARNING** (suggestion)

**Issue:** Bulk delete fires `Promise.allSettled` over individual delete mutations — N round-trips for N selected customers with no batch endpoint (Rule 85, 39).

**Suggestion:** Add a `useBulkDeleteCustomers()` hook that calls a single `DELETE /api/customers?ids=1,2,3` endpoint. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`frontend/src/app/(app)/customers/page.tsx` L65: 
**INFO** (suggestion)

**Issue:** `loadViews`/`saveViews` use `localStorage` directly, which throws in SSR or private-browsing contexts — the `try/catch` masks this (Rule 146, 50).

**Suggestion:** Guard with `typeof window !== "undefined"` before access, or extract a `safeStorage` wrapper used consistently. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`frontend/src/app/(app)/layout.tsx` L97: 
**WARNING** (suggestion)

**Issue:** Empty `catch {}` in the task `onSubmit` silently swallows all errors — the comment claims the mutation hook handles it, but if it doesn't, failures vanish with no log (Rule 46, 41).

**Suggestion:** Add a `console.error(err)` inside the catch, or pass the error to the mutation hook's `onError` handler explicitly. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`frontend/src/app/(app)/layout.tsx` L95: 
**INFO** (suggestion)

**Issue:** `data.status as string | undefined` cast suggests a type mismatch between `TaskModal`'s `onSubmit` payload and the mutation's expected input shape (Rule 27).

**Suggestion:** Define a shared `TaskSubmitPayload` type in the modal and mutation hook so the cast is unnecessary. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`frontend/src/app/(app)/layout.tsx` L88: 
**INFO** (suggestion)

**Issue:** The layout embeds a feature-specific `TaskModal` wired to a global store — the layout should be domain-agnostic; this is a separation-of-concerns concern (Rule 2).

**Suggestion:** Move the `<TaskModal>` into a `<TaskQuickAdd />` provider mounted within the layout, so the layout stays a shell. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`frontend/src/app/(auth)/login/page.test.tsx` L8: 
**INFO** (suggestion)

**Issue:** The crypto-js mock (lines 8-30) reimplements encrypt/decrypt with base64+JSON — this is 22 lines of test-only logic in a component test file (Rule 5).

**Suggestion:** Move to `tests/mocks/crypto-js.ts` or a shared `setupCryptoMock()` helper if other test files need it. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`frontend/src/app/(auth)/login/page.test.tsx` L59: 
**INFO** (suggestion)

**Issue:** The `mockAuthStore` pattern with a manual `selector` function is reinventing a mini-zustand store inside the test — fragile if the real store's selector API changes (Rule 66, 135).

**Suggestion:** Use `vi.mocked(useAuthStore)` with `vi.fn().mockImplementation((sel) => sel(mockAuthStore))`, or extract a `createMockAuthStore(overrides)` factory. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`frontend/src/app/(auth)/login/page.test.tsx` L120: 
**INFO** (suggestion)

**Issue:** Hardcoded test passwords (`"correctpassword"`, `"wrongpassword"`) are fine for fixtures, but ensure CI does not echo full test logs containing these to public dashboards (Rule 57).

**Suggestion:** No code change — just verify CI log retention/redaction policies are configured. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`frontend/src/app/(auth)/login/page.test.tsx` L73: 
**INFO** (suggestion)

**Issue:** `beforeEach` resets mocks manually after `vi.clearAllMocks()` — `vi.restoreAllMocks()` would also restore spy implementations automatically (Rule 63).

**Suggestion:** Consider `vi.resetAllMocks()` which clears call history and resets implementations in one call. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`frontend/src/app/(auth)/login/page.tsx` L44: 
**INFO** (suggestion)

**Issue:** The `<Input>` component is passed a long inline className that duplicates base layout/typography styles — either the shared `Input` is not applying its own styles (leaking layout concerns into consumers) or these classes are dead duplication.

**Suggestion:** Check `components/ui/input.tsx` — if it already applies padding/border/ring classes, drop the className prop; if it does not, move those styles into the shared component so all callers benefit. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`frontend/src/app/(auth)/login/page.tsx` L43: 
**INFO** (suggestion)

**Issue:** Neither input has an associated `<label>` or `aria-label`, so screen readers will announce only the placeholder text (which disappears once the user types).

**Suggestion:** Import the `Label` component from `@/components/ui/label` and wrap each input: `<Label htmlFor="username">Username</Label><Input id="username" ... />`. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`frontend/src/components/layout/app-sidebar.tsx` L94: 
**WARNING** (suggestion)

**Issue:** Casting through `Record<string, unknown>` to reach `unread_count` bypasses the type system — a backend rename of the field will silently fall back to 0 with no compile-time signal.

**Suggestion:** Define a proper response type: `interface NotificationsResponse extends PaginatedResponse<Notification> { unread_count: number }` and update the `useNotifications` hook's return type, then read `notifData?.unread_count ?? 0`. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`frontend/src/components/layout/app-sidebar.tsx` L46: 
**INFO** (suggestion)

**Issue:** `items: typeof crmItems` couples `NavGroup` to one specific array's inferred type, making it fragile to the next menu addition that has a slightly different shape.

**Suggestion:** Extract a shared interface: `interface NavItem { href: string; label: string; icon: LucideIcon; badge?: boolean }` and type the prop as `items: NavItem[]`. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`frontend/src/components/layout/app-sidebar.tsx` L136: 
**INFO** (suggestion)

**Issue:** The 'Add task' quick-action button sits inside the sidebar header with no visual separation from the collapse toggle — both are `h-7 w-7` ghost icons and can be mis-tapped, especially when the sidebar is collapsed.

**Suggestion:** Add `aria-label="Add task"` (already present) but also add a visible separator or move the `+` button into the nav body to distinguish it from the collapse control. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`frontend/src/lib/components/offline-banner.tsx` L26: 
**INFO** (suggestion)

**Issue:** The collapsed-state classes (`h-0 overflow-hidden py-0 px-0 border-0`) work but are tightly coupled to specific Tailwind utilities — if the parent layout uses flexbox or grid, `h-0` may still leave a gap.

**Suggestion:** Add `aria-hidden={!isOffline}` and conditionally render `{isOffline && <div ...>...</div>}` to fully remove it from the DOM when online, avoiding any layout interference. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`frontend/vitest.config.ts` L9: 
**WARNING** (suggestion)

**Issue:** Hardcoded `NEXT_PUBLIC_AUTH_KEY` value committed in test config — even as a placeholder, hardcoding auth material in version-controlled config violates rule 51 and sets a bad precedent.

**Suggestion:** Move the test value to a `.env.test` file loaded by vitest setup, or use an obviously fake sentinel like `'test-key-not-secret'`. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`frontend/vitest.config.ts` L11: 
**INFO** (suggestion)

**Issue:** No `setupFiles` configured — shared test setup (e.g. `@testing-library/jest-dom` matchers, polyfills) likely lives in individual test files, causing duplication.

**Suggestion:** Add `setupFiles: ['./src/test-setup.ts']` and import shared matchers there. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`scripts/dev/seed_admin_user.py` L57: 
**INFO** (suggestion)

**Issue:** Hardcoded JWT secret fallback (`'dev-seed-only'`) is acceptable for a dev-only script but the value is not obviously a placeholder — a future reader might copy it to production config (rule 51).

**Suggestion:** Use an obviously fake sentinel like `'_dev_seed_placeholder_'` or raise immediately if `JWT_SECRET_KEY` is unset. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`scripts/dev/seed_admin_user.py` L131: 
**INFO** (suggestion)

**Issue:** The seed script prints the chosen password to stdout in `_print_login_hint`, which will appear in CI logs and shared terminal scrollback (rule 57).

**Suggestion:** Print the username and a hint to look up the password elsewhere (e.g. `.env` file), or only print the password when `--verbose` is passed. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_
`scripts/dev/seed_admin_user.py` L36: 
**INFO** (suggestion)

**Issue:** The `env()` helper is over-engineered: `os.environ.get(key, default) or default` already handles both missing keys and empty-string values — the separate `val if val else` check is redundant.

**Suggestion:** Replace with `os.environ.get(key, default) or default` and drop the helper. — _HTTP 422: {"message":"Validation Failed","errors":[{"resource":"PullRequestReviewComment","code":"custom","field":"pull_request_review_thread.line","message":"could not be resolved"}],"documentation_url":"https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request","status":"422"}_

